PR #32, opened 2015-01-14
worker warmup callback

divaykin:
feature allowing to run certain code in the beginning of the worker thread.
one can use it to set thread priority, affinity mask, etc.

### DIFF
--- a/include/spdlog/spdlog.h
+++ b/include/spdlog/spdlog.h
@@ -63,7 +63,7 @@ void set_level(level::level_enum log_level);
 
 // Turn on async mode and set the queue size for each async_logger
 
-void set_async_mode(size_t queue_size, const std::function<void()>& worker_warmup_cb);
+void set_async_mode(size_t queue_size, const std::function<void()>& worker_warmup_cb = nullptr);
 // Turn off async mode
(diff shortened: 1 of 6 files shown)


include/spdlog/spdlog.h
```diff
@@ -63,7 +63,7 @@ void set_level(level::level_enum log_level);
 
 // Turn on async mode and set the queue size for each async_logger
 
-void set_async_mode(size_t queue_size);
+void set_async_mode(size_t queue_size, const std::function<void()>& worker_warmup_cb);
```
xaqq: Shouldn't this use a default value of `nullptr` ?

xaqq:
This looks like a nice addition.
See my question in diff comment.
Also there is a few indentation problem.

gabime:
Yes. I agree. This should have default null

divaykin:
No problem. For indentation - do you use always spaces? I typically use tabs

gabime:
You can run the top level astyle.sh

divaykin:
Should be better now. astyle.sh does some extra mods which I guess you don't really want - don't know why exactly, no expert

PR #33, opened 2015-01-14
discard policies on queue overflow

divaykin:
for me it's much better to lose few lines of log vs stopping the processing thread due to buffer overflow.
maybe set_async_mode is somewhat over complicated now, feel free to suggest better design.

thx a lot

### DIFF
--- a/include/spdlog/async_logger.h
+++ b/include/spdlog/async_logger.h
@@ -38,6 +38,7 @@
 #include <functional>
 #include "common.h"
 #include "logger.h"
+#include "spdlog.h"
 
 
 namespace spdlog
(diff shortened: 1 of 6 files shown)


xaqq:
> for me it's much better to lose few lines of log vs stopping the processing thread due to buffer overflow.

And some people will want to make sure they do not lose any entry: I like your customizable solution !
I think `set_async_mode` is still fine and we need not worry.

PS: Your commits are not linked back to your github account, probably because you didn't set the email address or something.

divaykin:
You right, added another email to profile thx.

gabime:
Thanks for the pull. Nice addition indeed.
The 'enum class async_queue_overflow_policy' is better to be placed  in common.h instead of spdlog.h though..

PR #36, opened 2015-01-18
respect user-set compiler name

wilhelmtell:
What if, for example, my clang compiler executable name is not clang but clang-3.5?


wilhelmtell:
... This to allow saying

    CXX=clang++-3.5 make -f Makefile.clang

or

    make -f Makefile.clang CXX=clang++-3.5

PR #37, opened 2015-01-22
Fix broken examples in run from arbitrary dirs

wilhelmtell:
Examples work fine when run from example/, as in

```bash
$ cd example/
$ ./example
```

but otherwise break:

```bash
$ cd
$ ~/path/to/spdlog/example/bench
```

The above would fail on trying to open the logs/somefile for writing if the logs directory doesn't exist in the current directory.

Let's not assume there is a logs/ directory and just write our logs in the current directory. We can suffix the logs with .log so they're easy to find in .gitignore for cleanup.

And with that, we don't need the logs direcory, or the .gitignore inside it that keeps it hanging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,4 +33,4 @@ example/*
 !example/makefile
 !example/makefile.clang
 
-
+*.log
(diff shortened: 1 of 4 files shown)


gabime:
The logs dir is very convient to put all the junk log files without messing up the source dir. A better solution would be a command line flag for the logs dir, but i dont think it worth the effort. 
Lets leave it as is.

wilhelmtell:
Alternatively, have the build create a logs directory.

PR #38, opened 2015-01-27
coarse clock, short log level

divaykin:
There are two changes here - short (one letter) level name which is to me important (less space / more efficient / better readability due to constant prefix length).

And another is performance optimization - if SPDLOG_CLOCK_COARSE is defined than COARSE clock is used. Which makes HUGE impact.

W/o COARSE (normal logging)
`[denis@centosvm example]$ ./bench 
*******************************************************************************
Single thread, 1,048,576 iterations, auto flush=0
*******************************************************************************
rotating_st...		895,865/sec
daily_st...		790,050/sec
null_st...		2,152,852/sec

*******************************************************************************
10 threads sharing same logger, 1,048,576 iterations, auto_flush=0
*******************************************************************************
rotating_mt...		1,303,881/sec
daily_mt...		1,112,445/sec
null_mt...		2,111,470/sec

*******************************************************************************
async logging.. 10 threads sharing same logger, 1,048,576 iterations, auto_flush=0
*******************************************************************************
as...		2,254,303/sec
as...		2,735,628/sec
as...		2,716,915/sec`

With COARSE
`[denis@centosvm example]$ ./bench 
*******************************************************************************
Single thread, 1,048,576 iterations, auto flush=0
*******************************************************************************
rotating_st...		966,354/sec
daily_st...		862,434/sec
null_st...		2,835,814/sec

*******************************************************************************
10 threads sharing same logger, 1,048,576 iterations, auto_flush=0
*******************************************************************************
rotating_mt...		1,492,719/sec
daily_mt...		1,248,143/sec
null_mt...		2,209,334/sec

*******************************************************************************
async logging.. 10 threads sharing same logger, 1,048,576 iterations, auto_flush=0
*******************************************************************************
as...		2,841,635/sec
as...		4,245,019/sec
as...		4,100,125/sec`

Huge bust in async case because most of the time is spent on getting the clock value

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -71,9 +71,16 @@ typedef enum
 
 static const char* level_names[] { "trace", "debug", "info", "notice", "warning", "error", "critical", "alert", "emerg", "off"};
 
+static const char* short_level_names[] { "T", "D", "I", "N", "W", "E", "C", "A", "M", "O"};
+
 inline const char* to_str(spdlog::level::level_enum l)
 {
     return level_names[l];
+}
+
+inline const char* to_short_str(spdlog::level::level_enum l)
+{
+    return short_level_names[l];
 }
 } //level
 
--- a/include/spdlog/details/line_logger.h
+++ b/include/spdlog/details/line_logger.h
@@ -26,6 +26,9 @@
 
 #include "../common.h"
 #include "../logger.h"
+#ifdef SPDLOG_CLOCK_COARSE
+#include <time.h>
+#endif
 
 
 // Line logger class - aggregates operator<< calls to fast ostream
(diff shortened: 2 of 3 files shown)


gabime:
This is great. Why not always use CLOCK_REALTIME_COARSE in linux? What's the down side?

divaykin:
Less precision - sometimes maybe interesting to see how many micros passed by

divaykin:
But you can leave it as a default - doubt that there are many people who even print micros in their logs

gabime:
Accroding to [StackOverflow](http://stackoverflow.com/questions/6498972/faster-equivalent-of-gettimeofday)  -it could be tens of millis -  "your timing will only be accurate to the last tick, meaning it could be off by tens of milliseconds"

divaykin:
Yeah, I think most of the systems are on 1000Hz ticks nowadays. 

Would be even cooler to use RDTSC but it's somewhat difficult to translate to wall clock time. We can get another 15-20% improvement if we manage to do so.

gabime:
caching would be another option. but has it's own complications (locking between threads)

divaykin:
Indeed, if I'll have some time I'd try to do something. But currently more time spent on format() - maybe we can explore how to offload format() to logger working thread.

Issue #48, opened 2015-02-17
create console logger for android


fcoulombe:

fcoulombe:
I think I have an implementation if you're interested in checking it out

gabime:
Sure. Even better would be pull

gabime:
@fcoulombe Any news on this ?

fcoulombe:
I haven't forgotten, still busy with other things sorry.

gabime:
Ok, I will leave this open in the meanwhile

ruslo:
See https://github.com/gabime/spdlog/pull/130

gabime:
Implemented in #130

PR #51, opened 2015-02-17
creating an android sinker.

fcoulombe:
https://github.com/gabime/spdlog/issues/48


gabime:
Thanks..
Why the sink ctor requires unused args like ostream and force flush?
Also, I think that the sink should be just independent sink and not inherited by outstream sink.. Its up to the user to choose the right sink under android

fcoulombe:
I was trying to follow your design.
ostream looked like an interface to output to stdout but i guess that could also mean writing to a file?
gimme a bit and i'll refactor to its own thing

joewan:
Is spdlog support android platform?

gabime:
@joewan no, not oficially, but you can try the android sink in the sinks folder

Issue #65, opened 2015-03-24
replace SPDLOG_DEBUG/TRACE macro from  force_log to log


theambient:
because it is impossible to handle log level runtime otherwise.

gabime:
Log levels can be set at runtime..
The idea is that the DEBUG/TRACE macros will always appear if SPDLOG_DEBUG_ON  is defined, no matter the actual level.

theambient:
I understand you idea but it seems to be inconvenient, cause you specify at runtime, say, just error level, but accidentally see debug or trace log.

Also it would be nice to have such macros for all log levels forwarding log place (file, line) to actual log function.

I can prepare patch, but i need your decision on this subject.

gabime:
Ok. Lets try those suggestions. Pull would be nice.

gabime:
Fixed in 9603c2a20cc1f7356e9bcd250ba466a9c0842a6d

PR #70, opened 2015-03-26
issue #65: added macroses like SPDLOG_TRACE but with no force

theambient:
see #65


gabime:
Thanks. I also commited some comments to make it clear

theambient:
not at all)

PR #73, opened 2015-04-05
Locking standard I/O when logging from multiple threads.

laszlo-kiss:
This change set allows one to use standard I/O from multiple threads as long as those threads have a reference (pointer) to a shared mutex. This is optional and the old behavior is transparently available.

I use the library from multiple threads and create loggers in each thread that have different names, but use standard I/O as the backend. In that case a mutex is required if the messages are to be delivered w/o scrambling.

Note that I did not address the posibility that multiple threads log to the same file.

### DIFF
--- a/include/spdlog/details/logger_impl.h
+++ b/include/spdlog/details/logger_impl.h
@@ -36,7 +36,8 @@ template<class It>
 inline spdlog::logger::logger(const std::string& logger_name, const It& begin, const It& end) :
     _name(logger_name),
     _sinks(begin, end),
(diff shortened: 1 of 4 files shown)


gabime:
Thanks for the suggestion. It complicates the API though with the user need to deal with mutexes and stuff..
The basic model is that thread safe sinks (e.g stdout_sink_mt) can be shared between loggers  (to write same i/o stream for instance). That's why sinks in spdlog are defined as std::shared_ptr.

PR #89, opened 2015-05-08
customizable type of logger name via define

divaykin:
allows to define arbitrary type of a logger name


gabime:
Thanks for the pull!
The problem is that it changes the API in many places and makes it confusing and rigid (#defining in tweakme.h a major type of the API)

I prefer to have something simpler, which will not change the basic API so profoundly.

Maybe start with allowing the user to define vector<string> instead of unordered_map<string> which seems to be x2 faster when small number of loggers are used (do you know approx how many?)

in twaekme.h, something like:
```c++
//uncomment to use std::vector as the registry underlying container.
//Would result in faster get(..) calls when number of registered loggers is low (upto 10)
#define VECTOR_BASED_REGISTRY
```

divaykin:
okay, I can also do something like this
`if (a[0]==b[0] && strcmp(a, b)==0)`

gabime:
I think strcmp do it anyway internally..

Anyway let's just use std::string's operator==

gabime:
Another related optimization would be to avoid the mutex in the ```spdlog::get()```
If the client code is single threaded or never modifies the registry after the initialization phase, then there is no need to lock the registry at all

in the tweakme:
```c++
//Uncomment to avoid locking in the registry operations (spdlog::get() spdlog::drop() spdlog::register())
//This will work only if your code is single threaded or if you create all loggers at some initialization //phase and never modify the registry in other threads afterwards
#define NO_REGISTRY_MUTEX
```

PR #90, opened 2015-05-08
explicit flush()

divaykin:


gabime:
If we go this way (support flush on every sink) then the sink interface (in sink.h) is the one to contain the ```virtual void flush()=0;``` declaration.

I assume the user calls ```logger->flush()``` so the logger need to implement this too..

divaykin:
right, here you go

PR #92, opened 2015-05-09
vector based registry & no mutex at get

divaykin:


gabime:
Thanks. I will also templatize the registry class to minimize the ```ifdef```s

```c++
template<Mutex> class registry
{
  private Mutex _mutex;
}
```
 and then in common.h something like
```c++
#ifndef SPDLOG_NO_REGISTRY_MUTEX
using registry_t = registry<std::mutex>
#else
using registry_t = registry<details::null_mutex>
#endif
```

include/spdlog/details/registry.h
```diff
@@ -152,15 +184,25 @@ class registry
     void register_logger_impl(std::shared_ptr<logger> logger)
     {
         auto logger_name = logger->name();
-        if (_loggers.find(logger_name) != std::end(_loggers))
+        if (get(logger_name) != nullptr)
```
gabime: There is a critical bug here. A recursive lock 
I will fix

PR #93, opened 2015-05-09
async auto flush

divaykin:
how about this? please share your thoughts

### DIFF
--- a/include/spdlog/async_logger.h
+++ b/include/spdlog/async_logger.h
@@ -59,21 +59,21 @@ public:
                  size_t queue_size,
                  const async_overflow_policy overflow_policy =  async_overflow_policy::block_retry,
                  const std::function<void()>& worker_warmup_cb = nullptr,
-                 const std::chrono::milliseconds auto_flush_millis = std::chrono::milliseconds::zero());
+                 const std::chrono::milliseconds& flush_interval_ms = std::chrono::milliseconds::zero());
 
     async_logger(const std::string& logger_name,
                  sinks_init_list sinks,
                  size_t queue_size,
                  const async_overflow_policy overflow_policy = async_overflow_policy::block_retry,
                  const std::function<void()>& worker_warmup_cb = nullptr,
-                 const std::chrono::milliseconds auto_flush_millis = std::chrono::milliseconds::zero());
+                 const std::chrono::milliseconds& flush_interval_ms = std::chrono::milliseconds::zero());
 
     async_logger(const std::string& logger_name,
                  sink_ptr single_sink,
                  size_t queue_size,
                  const async_overflow_policy overflow_policy =  async_overflow_policy::block_retry,
                  const std::function<void()>& worker_warmup_cb = nullptr,
-                 const std::chrono::milliseconds auto_flush_millis = std::chrono::milliseconds::zero());
+                 const std::chrono::milliseconds& flush_interval_ms = std::chrono::milliseconds::zero());
 
 
 protected:
--- a/include/spdlog/async_logger.h
+++ b/include/spdlog/async_logger.h
@@ -58,19 +58,22 @@ public:
                  const It& end,
                  size_t queue_size,
                  const async_overflow_policy overflow_policy =  async_overflow_policy::block_retry,
-                 const std::function<void()>& worker_warmup_cb = nullptr);
+                 const std::function<void()>& worker_warmup_cb = nullptr,
+                 const std::chrono::milliseconds auto_flush_millis = std::chrono::milliseconds::zero());
 
     async_logger(const std::string& logger_name,
                  sinks_init_list sinks,
                  size_t queue_size,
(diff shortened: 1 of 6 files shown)


divaykin:
should I also keep counter of number of log writes and flush only in case if there is something? or libc would handle it for me?

include/spdlog/details/async_log_helper.h
```diff
@@ -261,7 +271,7 @@ inline void spdlog::details::async_log_helper::set_formatter(formatter_ptr msg_f
 
 
 // sleep,yield or return immediatly using the time passed since last message as a hint
-inline void spdlog::details::async_log_helper::sleep_or_yield(const clock::time_point& last_op_time)
+inline std::chrono::nanoseconds spdlog::details::async_log_helper::sleep_or_yield(const clock::time_point& last_op_time)
```
gabime: I don't think this function should be changed. The time_since_op can be calculated when needed. Not here

include/spdlog/details/async_log_helper.h
```diff
@@ -249,7 +254,12 @@ inline bool spdlog::details::async_log_helper::process_next_msg(clock::time_poin
     }
     else //empty queue
     {
-        sleep_or_yield(last_pop);
+        auto time_since_op = sleep_or_yield(last_pop);
+        if (_auto_flush_millis > std::chrono::milliseconds::zero() && time_since_op > _auto_flush_millis)
+        {
+            for (auto &s : _sinks)
+                s->flush();
```
gabime: There is a problem here. flush() will be called continuously as long as the queue is empty, not just every _auto_flush_millis.

include/spdlog/details/async_log_helper.h
divaykin: Then we need to call ::now() again (twice) which is relatively expensive. I can make sleep_or_yield() accepting time_since_op not last_op_time though.

include/spdlog/async_logger.h
```diff
@@ -58,19 +58,22 @@ class async_logger :public logger
                  const It& end,
                  size_t queue_size,
                  const async_overflow_policy overflow_policy =  async_overflow_policy::block_retry,
-                 const std::function<void()>& worker_warmup_cb = nullptr);
+                 const std::function<void()>& worker_warmup_cb = nullptr,
+                 const std::chrono::milliseconds auto_flush_millis = std::chrono::milliseconds::zero());
```
gabime: Please rename  ```auto_flush_millis``` to ```flush_interval_ms``

include/spdlog/details/async_log_helper.h
divaykin: That's right, then I need to save last flush time

gabime: Also - this might not work as expected when user logs at certain slow rate - the flush will never be called in the defined interval.

gabime: Also, please use spdlog::log_clock::time_point now() - it will be faster (and can be tweaked to use coarse clock)

divaykin:
thanks please review the latest @gabime

gabime:
Thanks. Seems fine. 
The only issue I have is whether it's ok to flush only when queue is empty..Might be confusing.
I think it is better to do the check in each iteration regardless the state of the queue (right after line 245)
something like 
```c++
auto now = details::os::now();
handle_flush(now);
 if (_q.dequeue(incoming_async_msg))
    {
        last_pop = now;    
         ...
     }
     else //empty queue
    { 
        ...
    }
}

//Do flush if needed and update "now"
void handle_flush(log_clock::time_point& now)
{
   if (_flush_interval_ms <= std::chrono::milliseconds::zero() || (now - last_flush) < _flush_interval_ms)
        return;
    last_flush = now;
    for (auto &s : _sinks)
    s->flush();
       now = details::os::now(); //because the flush above takes time

}```

divaykin:
@gabime i think if the queue is not empty there is no point of flushing yet. the problem what i'm trying to solve is messages what are not appearing in the file during low activity phase. i'd rather keep worker thread going through the queue as fast as it could

gabime:
Right. And can also might degrade performance with redundant flushes..

On the other hand, flushing interval becomes less predictable..
For example on busy times, when queue is never empty then flush will not occur in the requested interval..
So as a user it would never be clear to me when the flush would be called..As a user I would always need to speculate which is not good

divaykin:
if your queue is always non empty then your output buffer would fill very fast & it'd be flushed anyway by the glibc. unless you managed to set the buffer size to very high numner

gabime:
Yes. This was my first thought, but then it felt a little unpredictable.
Lets try to summerize the pros/cons of flushing only in empty queue

**pros:**
Somewhat better performance on busy times.

**cons:**
Less predictable and harder to understand behaviour.

Lets go for your option but we need to document properly in the docs what is going on..

PR #95, opened 2015-05-11
adds missing flush override to simple_file_sink

### LINKED ISSUE #94: Did something changed in `simple_file_sink` ?
Hi the latest commits break my builds with the following error:

```bash
error: field type 'spdlog::sinks::simple_file_sink<spdlog::details::null_mutex>' 
is an abstract class

...

spdlog/include/spdlog/sinks/../details/../sinks/sink.h:38:18: 
note: unimplemented pure virtual method 'flush' in 'simple_file_sink'
    virtual void flush() = 0;
                 ^
```

when doing:

`std::make_shared<spdlog::sinks::simple_file_sink_st>(name + ".log", false);`

If `spdlog::sinks::simple_file_sink_st` is deprecated maybe it could be better to mark it with an attribute containing how to fix it and maybe the reason: `[[deprecated("reason + how to fix it")]]`.

gnzlbg:
closes #94

### DIFF
--- a/include/spdlog/sinks/file_sinks.h
+++ b/include/spdlog/sinks/file_sinks.h
@@ -47,6 +47,10 @@ public:
     {
         _file_helper.open(filename);
     }
+    void flush() override
+    {
+        _file_helper.flush();
+    }
 
 protected:
     void _sink_it(const details::log_msg& msg) override


gabime:
Thanks. Very helpful pull!

PR #96, opened 2015-05-11
travis-ci automated testing

gnzlbg:
- add travis-ci automated testing
- only for the examples
- compiles with clang-3.5, 3.6, 3.7 (using libc++), and gcc 4.9 (using libstdc++)
- with and without address sanitizer
- in c++11 and c++14 `-std=` modes
- runs example and bench with and without valgrind

Note: the only thing one can get right now from this is that the code compiles successfully. 
To check that is doing the right thing one would need to add unit/regression tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,31 +27,28 @@ matrix:
             - ubuntu-toolchain-r-test
             - llvm-toolchain-precise-3.5
 
-    # - env: CLANG_VERSION=3.5 BUILD_TYPE=Release CPP=11 ASAN=On LIBCXX=On
-    #   os: linux
-    #   addons: *clang35
+    - env: CLANG_VERSION=3.5 BUILD_TYPE=Release CPP=11 ASAN=On LIBCXX=On
+      os: linux
+      addons: *clang35
+    - env: CLANG_VERSION=3.5 BUILD_TYPE=Debug CPP=14 ASAN=On LIBCXX=On
+      os: linux
+      addons: *clang35
+    - env: CLANG_VERSION=3.5 BUILD_TYPE=Release CPP=14 ASAN=On LIBCXX=On
+      os: linux
+      addons: *clang35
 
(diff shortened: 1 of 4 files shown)


gnzlbg:
This will take a while to run because I've included all compiler configurations I could think of. 
Let's see first if all compilers work fine, and then it is maybe worth it to discuss how we can reduce the number of tested configurations a bit.

- It might be worth it to extend this to test with the ThreadSanitizer, and to add test that check the output.

gabime:
OK. Do whatever you see fit

gnzlbg:
A brief travis tutorial:

The whole point of travis is to make sure that you don't break your build or your users build.

Next to each commit in each pull-request it shows a circle (yellow = test running, green = test passed, red = test failed). Below each pull-request it shows a message (again in yellow, red, or green) telling you if merging the pull-request breaks your build. You can click on the circle and on the link to go to the travis build bot and see which tests are running, which tests passed, which failed, and look at the console logs.

Right now the circle is on yellow (test running). If you click on it, you will see that some tests actually fail. When it finishes running, it is going to turn red and tell you "Do not merge this pull-request yet!". In your Readme.md it is going to say that the travis build for your project is actually failing.

Long story short, you should not merge code that breaks your build. It has costed me time this morning to figure out why my code was breaking. It was breaking because _you_ merged the `flush` feature without testing it first. 

Travis prevents you from doing that accidentally (if you do it, you do it knowingly). 

If you develop code in your own branches and commit there, travis checks there that your code doesn't break anything, and then you can merge it safely.

gnzlbg:
Finally I should have probably rebased this into a single commit. 
You should ask me to do this before merging it or otherwise you might pull in some broken history.
I've sent you another pull-request, right now I cannot rebase any changes without going through a lot of git hacks because rebasing in my branch would break yours.

gabime:
Ok. Thanks for the detailed explanation.

PR #103, opened 2015-05-26
Add shared registry capability

AlexandrePTJ:
This is a proposal to enable shared registry.
It can be improved, but the idea is here.


include/spdlog/details/registry.h
```diff
@@ -144,12 +144,15 @@ template <class Mutex> class registry_t
         _async_mode = false;
     }
 
-    static registry_t<Mutex>& instance()
+    static registry_t<Mutex>* instance()
```
gabime: Why the change from ref to pointer?

include/spdlog/details/registry_shared.h
```diff
@@ -0,0 +1,66 @@
...
+#   else
+        typedef registry_shared_t<std::mutex> registry;
+#   endif //SPDLOG_NO_REGISTRY_MUTEX
+#endif //SPDLOG_USE_SHARED_REGISTRY
+
+template<class Mutex>
+std::shared_ptr<registry_shared_t<Mutex> > registry_shared_t<Mutex>::_instance = nullptr;
+
+}
+}
```
gabime: If the only thing that is added is the  set_registry function, Why not just to add this function to the orig registry class . This way instead of all this new code and ifdefs, only 3 lines of code would be added.

include/spdlog/details/registry.h
AlexandrePTJ: Because we need to get its intance from the application to be able to set the instance inside the library. But this should be a `std::shared_ptr` instead.

From the application:
```cpp
auto shared_registry = spdlog::details::registry::instance();
MyLib::setLoggerRegistry(shared_registry);
```

In the library:
```cpp
static void MyLib::setLoggerRegistry(std::shared_ptr<spdlog::details::registry> registry)
{
    spdlog::details::registry::set_registry(registry};
}
```

include/spdlog/details/registry_shared.h
AlexandrePTJ: You're right. I correct this.

AlexandrePTJ:
Unfortunately I am not very familiar with thread safety. I though that shared_ptr was.
I will look forward about this. Do you have a test case I could work with ?

gabime:
shared_ptr is thread safe but it is not enough here. This is classic singleton thread problem, which is most easily solved with static instance. see http://stackoverflow.com/questions/2576022/efficient-thread-safe-singleton-in-c for discussion about it

xaqq:
Note that with C++11 this problem is auto-magically fixed.

This is from the standard:
"If control enters the declaration concurrently while the variable is being
initialized, the concurrent execution shall wait for completion of the
initialization."


On Tue, May 26, 2015 at 3:50 PM, Gabi Melman <notifications@github.com>
wrote:

> shared_ptr is thread safe but it is not enough here. This is classic
> singleton thread problem, which is most easily solved with static instance.
> see
> http://stackoverflow.com/questions/2576022/efficient-thread-safe-singleton-in-c
> for discussion about it
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gabime/spdlog/pull/103#issuecomment-105530903>.
>



-- 
Kapp Arnaud - Xaqq

AlexandrePTJ:
The problem magically fixed is the regular singleton pattern ?

xaqq:
Yes that's what I meant.

On Wed, May 27, 2015 at 8:45 AM, Alexandre Petitjean <
notifications@github.com> wrote:

> The problem magically fixed is the regular singleton pattern ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gabime/spdlog/pull/103#issuecomment-105778186>.
>



-- 
Kapp Arnaud - Xaqq

PR #104, opened 2015-05-27
Add default null sink based logger

AlexandrePTJ:
Proposal to add a default logger.
Allow `spdlog::get` to return an usable logger when trying to get an unregistered logger.

### DIFF
--- a/include/spdlog/details/registry.h
+++ b/include/spdlog/details/registry.h
@@ -37,6 +37,7 @@
 #include "../logger.h"
 #include "../async_logger.h"
 #include "../common.h"
+#include "../sinks/null_sink.h"
 
 namespace spdlog
 {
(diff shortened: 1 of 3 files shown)


gabime:
What is the use case?
Also, It fails the unit tests (see the travis log)

AlexandrePTJ:
Currently when a trying to get a logger not registred, it return a `nullptr`.
Then, to follow the design, we should check each time if returned logger is valid before using it. I think we shouldn't to benefit spdlog performance. That is why I made this proposal.

Tests fails because it check for a `nullptr` after `registry::drop` and `registry::dropall` calls.

gabime:
You should not check each time. If the program creates the logger at some init stage, then there is no need.
Anyway getting a default null logger is probably even worse. It better to crash on such cases and fix the bug than logging to null sink without knowing about it.

AlexandrePTJ:
That is a good point. Then I remove this PR.

PR #106, opened 2015-06-02
Expose line_logger enabled state for custom types in stream operators

nickolasrossi:
Custom stream operators for line_logger need to be able to check line_logger.enabled and skip execution if it's false.

### DIFF
--- a/include/spdlog/details/line_logger.h
+++ b/include/spdlog/details/line_logger.h
@@ -206,6 +206,11 @@ public:
         _enabled = false;
     }
 
+    bool is_enabled()
+    {
+        return _enabled;
+    }
+
 
 private:
     logger* _callback_logger;
--- a/include/spdlog/details/line_logger.h
+++ b/include/spdlog/details/line_logger.h
@@ -206,7 +206,7 @@ public:
         _enabled = false;
     }
 
-    bool is_enabled()
+    bool is_enabled() const
     {
         return _enabled;
     }


xaqq:
You might as well mark it `const`.

gabime:
@nickolasrossi What is the intended usage? Do you have an example

nickolasrossi:
Sure...is there a better way to accomplish this sort of thing?

```c++
spdlog::details::line_logger& operator<<(spdlog::details::line_logger& log, boost::asio::const_buffer& buffer)
{
    if (!log.is_enabled()) return log;

    const unsigned char* data = boost::asio::buffer_cast<const unsigned char*>(buffer);
    size_t size = boost::asio::buffer_size(buffer);

    log << '[';

    char hex[3];
    for (int i = 0; i < size; i++) {
        sprintf(hex, "%02x", data[i]);
        if (i > 0) log << ' ';
        log << hex;
    }

    log << ']';

    return log;
}
```

gabime:
Why not do same as the example of custom class logging ?

nickolasrossi:
In this case I'm using a class from a library (not const_buffer as above) that implements ostream operator<< in a way I can't use or redefine. Implementing a line_logger operator<< directly was my workaround.

nickolasrossi:
line_logger is a super nice design btw. Made this very easy to achieve.

gabime:
Good to hear. Thanks :)

PR #111, opened 2015-07-13
Add support for wchar_t on Windows

Kentzo:


gabime:
Thanks for the pull. Two problems: 
1.the build fails in travis 
2. I dont want to impose wchar on windows. It should be configurable

Kentzo:
1. We will fix that, @a-martynovich Please take a look.
2. Do you have some concept in mind to make it configurable?

gabime:
Not sure. Maybe simple ifdef will do. Most of them are defined in  tweakme.h

Kentzo:
@gabime Could you put some explanation behind making it configurable? The majority of windows desktop applications will use wchar_t (either right away or after they realize their app crash for a lot of users).

gabime:
@Kentzo  add '''#define use_whcar''' to tweakme.h

gabime:
Seems the tests fails because it gets filesize of 2048 but expets 1024.. Makes sense

a-martynovich:
@gabime Fixed, please take a look.

gabime:
Thanks. Looks good.
Please remove the use of```SPDLOG_NO_WCHAR``` from the code -It is redundant
```#define SPDLOG_USE_WCHAR``` is enough (and should  commented out by default in tweakme.h)

**Comment by Kentzo** [comment]
@gabime Done but tests failed. Not sure if it's due to our changes.

**Comment by gabime** [comment]
I reverted this - it causes problems - global defines that conflicts with use code. We need to find a cleaner solution. Maybe a dedicated  file sink

**Comment by Kentzo** [comment]
Current version is completely unusable without that change on Windows (except developer's environment I guess)...

Do you have a set of changes in mind to be made?

**Comment by Kentzo** [comment]
@gabime I'm willing to contribute so we can use spdlog's upstream in our product. Please think of steps this change requires to make it happen.

**Comment by gabime** [comment]
Main issue was that it broke user code because it introdeced global defines..

**Comment by Kentzo** [comment]
@gabime We could use templates

**Comment by gabime** [comment]
Another option would be defining a new  sink class which supports wchar. This would be the less intrusive option i think.

**Comment by VikingExplorer** [comment]
What's the status of this?  I'm getting
```
		auto log = spdlog::get( "debug" );
		log->debug( fmt.value().c_str(), value1 );
		auto msg = fmt::format( fmt.value(), value1 );

1>Log.cpp(75): error C2664: 'spdlog::details::line_logger spdlog::logger::debug<wchar_t>(const char *,const wchar_t &)': cannot convert argument 1 from 'const wchar_t *' to 'const char *'
```
I have already integrated with CppFormat, so I know it supports wchar_t. 

What am I doing wrong?  I searched the code for SPDLOG_USE_WCHAR and couldn't find it.

Kentzo:
@VikingExplorer Change was reverted in the mainstream. We're using our fork for know. Feel free to improve this PR.

Kentzo:
@gabime Is it all right to have the SPDLOG_USE_WCHAR macro though?

Kentzo:
I don't think supporting 2 sink is a good solution: you would end up with 2 identical version of the code… Maybe I put the S macro under the SPDLOG_ namespace?

gabime:
I don't know.. The code got quite ugly with all those "tstrings" everywhere.. Not sure it worth it

Kentzo:
@gabime I tried to do it with templates, it becomes even uglier: you cannot easily get around with char and wchar_t string literals.

Current implementation is terribly unusable on Windows though: you cannot have a path with Chinese symbols for example.

Kentzo:
@gabime Made another PR that attempts to solve this problem.

gabime:
@Kentzo it fails the tests. included file not found..

Kentzo:
@gabime Fixed.

Issue #86, opened 2015-05-05
Library versioning?


mattklein123:
Hi,

Library looks pretty interesting and am considering using it in a project. Are you thinking about starting to tag versions at some point soon? I would prefer to be able to import at specific versions vs. master if possible.

(This isn't really an issue per say but I figured I would ask here)

Thanks,
Matt

gabime:
Sure, At some point versioning will be applied - when the API stabilizes a bit..not sure when though..

dkopecek:
I'm also a bit interested in this topic. I've packaged your project for Fedora ([pkgdb link](https://admin.fedoraproject.org/pkgdb/package/spdlog/), [package review link](https://bugzilla.redhat.com/1210753)) because I use it in my project which I'd like to have there and Fedora has strict library bundling rules -- i.e. you can't bundle anything unless it's really necessary.

If are going to do tagged releases at some point in the future, then I'm looking forward to it, because that will simplify the decision when to update the package in Fedora.

mattklein123:
Out of curiosity, do you foresee breaking changes coming soon? Is there any reason not to put a version on what is there right now? It seems fairly complete (for what it does currently).

gabime:
There probably will be breaking changes.. There is always room for improvment..

gabime:
@dkopecek thanks for the packaging! 

Btw, I would be happy to get some feedback on the lib (strong/weak points etc.)

gregoire-astruc:
Hi,

I just had a glance at this project and I'm very likely to use it in a couple projects.

Regarding versioning, I'd strongly recommend following the [SemVer](http://semver.org) approach: it would enable you to start versioning right now, even if you feel you're still in alpha/beta stage.

API might break now or even in the future and this scheme gives you the chance to inform clients about it.

Hope that helps.

gabime:
Thanks for the feedback. I will take a look at SemVer approach..

memsharded:
I know this is an old issue, but I also think that tagging at some points would be beneficial. I have created a spdlog package for conan: https://www.conan.io/source/spdlog/0.1/memsharded/testing, and used the latest commit, named it 0.1. I can do a "master" version getting the latest commit, but for package management and reproducibility, I think is better to fix the package to one commit. 

So it would be great if you could tag your project at some points, it doesn't really require to follow semver, or be really special releases. It is just an easy name for consumers to handle reproducibility without having to resort to ugly sha1 commit hashes. Thank you, and thanks for such a great project.

[...]

PR #127, opened 2015-09-10
Add CMake build

ruslo:
With this CMake code library can be installed and found by `find_package` like this:

```cmake
find_package(spdlog CONFIG REQUIRED)

add_executable(example example.cpp)
target_link_libraries(example spdlog::spdlog)
```

also tests can be run using CTest.

I see no releases so version set to `1.0.0`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,80 @@
+# *************************************************************************/
+# * Copyright (c) 2015 Ruslan Baratov.                                    */
+# *                                                                       */
+# * Permission is hereby granted, free of charge, to any person obtaining */
+# * a copy of this software and associated documentation files (the       */
+# * "Software"), to deal in the Software without restriction, including   */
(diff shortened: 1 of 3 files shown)


ruslo:
> I see no releases so version set to 1.0.0.

Related: https://github.com/gabime/spdlog/issues/86

gabime:
Thanks. Could you please add a section in the wiki that briefly explains the cmake usage?

ruslo:
> Could you please add a section in the wiki that briefly explains the cmake usage?

@gabime please take a look here: https://github.com/gabime/spdlog/wiki/CMake-build

gabime:
Thanks. What do you mean that it "can be built"?  It is a header only lib..

ruslo:
> What do you mean that it "can be built"?

What about: `Library can be installed and examples can be built` ?

gabime:
Yes. Much better. Thanks for contributing this stuff

PR #130, opened 2015-09-10
Add Android sink

ruslo:


include/spdlog/sinks/android_sink.h
```diff
@@ -0,0 +1,101 @@
...
+            case spdlog::level::trace: return ANDROID_LOG_VERBOSE;
+            case spdlog::level::debug: return ANDROID_LOG_DEBUG;
+            case spdlog::level::info: return ANDROID_LOG_INFO;
+            case spdlog::level::notice: return ANDROID_LOG_INFO;
+            case spdlog::level::warn: return ANDROID_LOG_WARN;
+            case spdlog::level::err: return ANDROID_LOG_ERROR;
+            case spdlog::level::critical: return ANDROID_LOG_FATAL;
+            case spdlog::level::alert: return ANDROID_LOG_FATAL;
+            case spdlog::level::emerg: return ANDROID_LOG_FATAL;
+            case spdlog::level::off: throw spdlog_ex("Unexpected off");
```
ruslo: Not sure about what is this level about, so just throw exception. Let me know if it's not correct.

include/spdlog/details/spdlog_impl.h
```diff
@@ -31,6 +31,7 @@
 #include "../sinks/file_sinks.h"
 #include "../sinks/stdout_sinks.h"
 #include "../sinks/syslog_sink.h"
+#include "../sinks/android_sink.h"
```
gabime: No need to include this file. It is not used from splog_imple.h as far as i can tell

include/spdlog/sinks/android_sink.h
gabime: Off means no logging. It should not throw, just ignore the log

include/spdlog/sinks/android_sink.h
```diff
@@ -0,0 +1,101 @@
...
+        else
+        {
+            throw spdlog_ex("Send to Android logcat failed");
+        }
+    }
+
+private:
+    static android_LogPriority convert_to_android(spdlog::level::level_enum level)
+    {
+        switch(level)
```
gabime: Could be lot faster using a static lookup array. See the syslog sink for an example

ruslo: @gabime I've wrote a [test](https://github.com/ruslo/spdlog/blob/android.test/example/android.cpp) with [two variants of converting](https://github.com/ruslo/spdlog/blob/android.test/include/spdlog/sinks/android_sink.h). Here is result:
```
Test:switch time:11258605569
Test:array time:10944504215
Test:switch time:11012342079
Test:array time:11089379475
Test:switch time:10916645205
Test:array time:10934150672
```
I see no significant performance difference even in this syntetic test. Actually in two last tests "switch" is faster.

include/spdlog/sinks/android_sink.h
ruslo: > Off means no logging. It should not throw, just ignore the log

Should we return from `_sink_it` (i.e. `if (level == off) return`) or check `assert(level != off)`?

include/spdlog/details/spdlog_impl.h
ruslo: I was thinking that the only file user need to include to work with library is `#include "spdlog/spdlog.h"` so I guess we need to include `sinks/android_sink.h` somewhere internally. I've just found other sinks and add android one there.

gabime: No. They are included because they are used in this file. If android is not used in spdlog_impl.h, it should not included

include/spdlog/sinks/android_sink.h
gabime: Ok fair enough - even though the array approach "feels" better(simple lookup instead of recalculating over and over again),
Also maybe if the test used emerg() instead of info() youd see a bigger difference.

include/spdlog/sinks/android_sink.h
gabime: Just return i think

include/spdlog/details/spdlog_impl.h
ruslo: @gabime Okay, so to clarify: I'm removing this line and it means that if user want to use android sink he should include header manually, like `#include <spdlog/sinks/android_sink.h`. Guess it would be nice to document it somewhere (?)

gabime: Right.

include/spdlog/sinks/android_sink.h
ruslo: > simple lookup instead of recalculating over and over again

`simple lookup` mean loading value from memory and `recalculating over and over again` mean quite trivial arithmetic on registers. Just imagine if `spdlog::level::FOO` equals to `ANDROID_LOG_FOO`, i.e. this function just do nothing and return `level` argument. Anyway I think both operations take very small time so I doubt we can take any adequate benchmarks here.

> Also maybe if the test used emerg() instead of info() youd see a bigger difference.

nope, still the same

Also you may find interesting this test: https://gist.github.com/ruslo/f0bb18c73f3d08ea544b
As you can see in case of `boo` (with switch) compiler is smart enough to figure out that result is 0x76 = 0x34 + 0x42, but not in case of `foo` (with loading from memory).

ruslo:
@gabime updated

PR #133, opened 2015-09-17
Add distributional sink

daersc:
Setting up dist_sink to a logger will inject a layer between the logger and
(some of) it's sinks. Sinks can be added and removed to dist_sink dynamically.

### DIFF
--- a/include/spdlog/sinks/dist_sink.h
+++ b/include/spdlog/sinks/dist_sink.h
@@ -60,6 +60,7 @@ protected:
 public:
     void flush() override
     {
+        std::lock_guard<Mutex> lock(base_sink<Mutex>::_mutex);
         for (auto iter = _sinks.begin(); iter != _sinks.end(); iter++)
             (*iter)->flush();
     }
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. It is not fully thread safe though.. The sink_it()or flush() could be called while another thread adds or deletes a sink..

daersc:
Yes you're right. The flush() method was not thread safe. But the _sink_it() method will be invoked by the log() method from base class, which locks the sink, isn't it?

gabime:
yes, but not with the same mutex as the one used to lock the add/remove functions

daersc:
The derived class uses the protected _mutex field from it's base class, so this is the same mutex.

gabime:
Right. My mistake

gabime:
I am not sure what is the use case here..
In what real world scenario would such sink be needed?

daersc:
The use case is to have multiple loggers working on the same sinks (let's say one for stderr and one for a logfile). Now the stderr-sink should be disabled for all loggers.
I could not find an easy way to do this for multiple loggers. Is there any?

gabime:
And how would it be disabled in the proposed solution? Removed  from the dist sink?

daersc:
Yes, this was my intention.

gabime:
ok. will merge. please add some documentarion explaining this (in the wiki please)

PR #141, opened 2015-10-27
Fix filename creation for daily_file_sink

AlexandrePTJ:


gabime:
what's the purpose of this fix?

PR #152, opened 2015-12-19
Add new feature directory

x86kernel:
if filename include realative path and it does not exist, create diretories by referring to filename

### DIFF
--- a/include/spdlog/details/file_helper.h
+++ b/include/spdlog/details/file_helper.h
@@ -47,16 +47,18 @@ public:
     {
 
         close();
-        std::list<std::string> dirs;
         const char* mode = truncate ? "wb" : "ab";
         _filename = fname;
         for (int tries = 0; tries < open_tries; ++tries)
         {
-            if (dir_check(fname)) 
+            if(!_dir_created) 
             {
-                get_dirlist(fname, dirs);
-                if(create_dirs(dirs))
-                    break;
+                if (dir_check(fname)) 
+                {
+                    get_dirlist(fname);
+                    if(create_dirs())
+                        break;
+                }
             }
 
             if(!os::fopen_s(&_fd, fname, mode))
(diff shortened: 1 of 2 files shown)


include/spdlog/details/os.h
```diff
@@ -135,8 +195,17 @@ inline int fopen_s(FILE** fp, const std::string& filename, const char* mode)
     *fp = _fsopen((filename.c_str()), mode, _SH_DENYWR);
     return *fp == nullptr;
 #else
+    std::list<std::string> dirs;
```
gabime: i am not sure its good idea to check dirs on every file open. should be done only once at some initial stage.

include/spdlog/details/os.h
```diff
@@ -128,6 +129,65 @@ constexpr inline unsigned short eol_size()
...
+
+#endif
+
+}
+
+//Create directories by referring to list 
+//Return 0 if success 
+inline bool create_dirs(std::list<std::string>& dirs)
+{
+#ifdef __linux__
```
gabime: Should be on windows too as spdlog supports windows and should behave the same where possible on all supported plaforms

include/spdlog/details/os.h
```diff
@@ -128,6 +129,65 @@ constexpr inline unsigned short eol_size()
...
 
+//Return true and list if filename contains relative
+inline bool dir_check(const std::string& filename, std::list<std::string>& dirs)
+{
+#ifdef __linux__
+    std::string directory;
+    std::string::size_type index, previndex = 0, size;
+
+    index = filename.find("/", previndex);
+    if(!index)
```
gabime: i am not sure i understand this line. why would anybody want to create "/"?

x86kernel: Ok, Refactoring of my code is partically done.

include/spdlog/details/os.h
x86kernel: Ok, Refactoring of my code is partically done.

include/spdlog/details/os.h
x86kernel: if the filename contains the absolute path, there is a "/" in front of the filename and a "/" will be appended into the list (the "/" will be passed as the argument of the 'mkdir' function).

if the directory which is the argument of the 'mkdir' function already exists, the 'errno' will be set as the 'EEXIST', So i put some code to ignore it.

As a result, the "/" will be passed as the argument of the 'chdir' function which is next to the 'mkdir' function

include/spdlog/details/file_helper.h
```diff
@@ -129,6 +137,77 @@ class file_helper
...
+                    return r;
+#endif
+            }
+
+            r = os::_chdir(dirname);
+            if(r) return r;
+        }
+
+        r = os::_chdir(origin_path);
+        if(r) return r;
```
gabime: chdir is an absolute no no. This will break code in other  threads that depend on the current directory

include/spdlog/details/os.h
gabime: too complicated in my opinion. i am sure there are simpler ways to create set of folders without trying to create "/"

x86kernel:
I changed the code so not to use the chdir function, and though I was unable to test it, it shall work fine in Windows.

PR #160, opened 2016-01-06
Fixed issues with time zone UTC offset and daylight saving offset on Windows

glebov-andrey:
### Issue 1:
This code:
```c++
if (!rv)
        return -1;
```
compares `rv` to `TIME_ZONE_ID_UNKNOWN == 0`
`TIME_ZONE_ID_UNKNOWN` means

> "Daylight saving time is not used in the current time zone, because there are no transition dates or automatic adjustment for daylight saving time is disabled." ([see Microsoft documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724421(v=vs.85).aspx))

so if the time zone doesn't have DST (like Europe/Moscow UTC+03:00) the function returns `-1`.
`Get[Dynamic]TimeZoneInformation` returns `TIME_ZONE_ID_INVALID` if it fails so comparing `rv` to it is the correct way to detect an error.

Also (not in the pull request as would violate the function's interface) it would be logical to throw an exception instead of returning `-1`:
```c++
throw std::system_error(GetLastError(), std::system_category());
```
### Issue 2:
```c++
return -1 * (tzinfo.Bias + tzinfo.DaylightBias);
```
In time zones with DST this function accounts for the DST offset even if currently daylight saving time is not in effect, so we must check the `tm::tm_isdst` flag (you can also check the return value of `Get[Dynamic]TimeZoneInformation` but using the `tm` field seems more correct from a logical standpoint).

### Issue 3:
[Microsoft documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/ms725481(v=vs.85).aspx) says that `[DYNAMIC_]TIME_ZONE_INFORMATION::StandardBias` contains

> "The bias value to be used during local time translations that occur during standard time... This value is added to the value of the Bias member to form the bias used during standard time. In most time zones, the value of this member is zero"

so we must account for this offset if `tm::tm_isdst` is `false`

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -170,7 +170,6 @@ inline int utc_minutes_offset(const std::tm& tm = details::os::localtime())
 {
 
 #ifdef _WIN32
-    (void)tm; // avoid unused param warning
 #if _WIN32_WINNT < _WIN32_WINNT_WS08
     TIME_ZONE_INFORMATION tzinfo;
     auto rv = GetTimeZoneInformation(&tzinfo);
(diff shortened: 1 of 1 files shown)


gabime:
Very useful pull. Thank !

PR #172, opened 2016-01-29
Add support for clearing file_sink current contents

nickaein:
This pull request adds `clear()` virtual function to the logger class and implements it for `file_sink` by calling `reopen(true)` on the `file_helper`. 
One should implement similar function for the other sinks. Some sink will not have a `clear` function (e.g. null_sink).


gabime:
this is too specific to be added to the logger class. for example what does clear mean for other sinks like stdout_sink?

Issue #177, opened 2016-02-13
syslog logger on Mac OS X


smessmer:
Is there a reason the syslog logger is only supported on linux?
As far as I know, Mac OS X also has a syslog.

gabime:
No, not good reason. 
Pull requests are welcome.

PR #178, opened 2016-02-13
Allow syslog logger on Mac OS X

smessmer:
Syslog is supported on Mac OS X, so there is no reason disallowing it. https://github.com/gabime/spdlog/issues/177


gabime:
Thanks for the pull

PR #185, opened 2016-03-12
add msvc logging sink

AlexanderDalshov:


gabime:
Thanks. what is the reason you called it "base"

AlexanderDalshov:
It was copy-pasted from android version, so if you consider that name should be other - please, change it.

PR #188, opened 2016-03-20
Added SQLite3 sink to the solution

PedroRod:
You can find more info on how to implement it here [https://github.com/PedroRod/sqlite3_database_sink](url)


include/spdlog/sinks/database_logger_sink.h
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	std::string ColumnName;
+	TableColumn ColumnMap;
+	std::string Value;
+};
+
+struct database_schema
+{
+	std::string TableName;
```
gabime: To be consistent with spdlog's naming convention, please use snake_case for all variable names ("table_name", "column_map", "column_name", "value", "thread_id" ,etc..)

include/spdlog/sinks/database_logger_sink.h
```diff
@@ -0,0 +1,156 @@
...
+					case Column::ThreadId:
+					{
+						column.Value = std::to_string(msg.thread_id);
+						break;
+					}
+					}
+				}
+
+				auto query = fmt::format("INSERT INTO {0} ({1},{3},{5},{7},{9}) VALUES ('{2}','{4}','{6}','{8}',{10})",
+					_schema.TableName,
```
gabime: This is sql injection waiting to happen.. Please use sql params instead

gabime:
Thanks. I added my comments in places that need to be fixed

include/spdlog/sinks/database_logger_sink.h
```diff
@@ -0,0 +1,156 @@
...
+
+				if (sqlite3_exec(_database, query.c_str(), nullptr, nullptr, &errorMessage) != SQLITE_OK)
+				{
+					throw spdlog_ex(errorMessage);
+				}
+
+
+			}
+
+			explicit database_logger_sink(const std::string& databaseName)
```
gabime: Please place ctors first, before all other public metthods

PedroRod:
I made some adjustments based on your comments.

You are totally right about the sql injection, I did not take that into consideration since for my application I do not need to worry about it, but you are totally right.
That got changed, changed code style to use same one you use for spdlog and removed the column mapping since there is no clean way of mapping columns in a prepared statement.

gabime:
Thanks. prepare_statement  should be called only once (e.g. upon ctor and the resultant sqlite3_stmt* stored as private member to be binded upon each call). Otherwise the performance will suffer. 
Also, please rename the sink and .h to sqlite_sink please

PR #194, opened 2016-04-13
Ansicolor sink now works correctly on windows

RavenX8:
The previous implementation of ansicolor_sink would print the ansi codes instead of change the colors on a windows console. The changes I did here allow this sink to work as intended on windows as well as linux.

If needed I can create another sink that will do the same, just for windows.


include/spdlog/sinks/ansicolor_sink.h
```diff
@@ -82,25 +132,38 @@ inline ansicolor_sink::ansicolor_sink(sink_ptr wrapped_sink) : sink_(wrapped_sin
...
+    sink_->log( m );
+#endif
 }
 
 inline void ansicolor_sink::flush()
 {
     sink_->flush();
 }
-
+#ifdef _WIN32
```
gabime: Why the different API ? The public API should be the same as spdlog is cross platform as much as  possible

RavenX8: The only way that this would be not cross platform is if you do not use the values given or store the color values.
There's three other ways I could implement it if you do not like this.

1. I could create a color_t type and have that change depending on the platform.
2. I could use a unordered_map to map the strings to the short values (Would only be used in windows)
3. I could write a praser that would interpreter the ansi codes (which would also support clearing the screen, moving the cursor and other stuff like that)

gabime: I prefer option 2, but I think we should separate the sink to a dedicated "wincolor_sink" (so there will be 2 sinks : "ansicolor_sink" and "wincolor_sink")

RavenX8: Alright, I'll put it in it's own sink. :+1:

godbyk: My initial thinking was that we'd have an `ansicolor_sink`, a `wincolor_sink`, and a `color_sink`. The `color_sink` would just `#ifdef` to use the appropriate platform-specific color sink. Top-level code could just use `color_sink` and expect it to work everywhere while the implementation details are hidden away in platform-specific classes.

RavenX8:
Updated. The wincolor_sink only changes colors for a console. If you try to wrap any other sink, it doesn't effect anything.

include/spdlog/sinks/wincolor_sink.h
```diff
@@ -0,0 +1,136 @@
...
+    colors_[level::critical] = bold | red;
+    colors_[level::alert]    = bold | white | on_red;
+    colors_[level::emerg]    = bold | yellow | on_red;
+}
+
+inline void wincolor_sink::log(const details::log_msg& msg)
+{
+    // Wrap the originally formatted message in color codes
+#ifdef _WIN32
+    SetConsoleTextAttribute(GetStdHandle( STD_OUTPUT_HANDLE ), colors_[msg.level]);
```
godbyk: Should this also handle `STD_ERROR_HANDLE` for when the sink logs to `STDERR`?

RavenX8: Good call.

include/spdlog/sinks/wincolor_sink.h
```diff
@@ -0,0 +1,139 @@
...
+
+#include <string>
+#include <map>
+
+namespace spdlog
+{
+namespace sinks
+{
+
+#ifndef _WIN32
```
gabime: I think all the ```#ifdef _WIN32``` in this file can be omitted. It is called ```wincolor_sink``` after all. It should not be included by the wrapper in non windows

include/spdlog/sinks/wincolor_sink.h
```diff
@@ -0,0 +1,139 @@
...
+    colors_[level::warn]     = bold | yellow;
+    colors_[level::critical] = bold | red;
+    colors_[level::alert]    = bold | white | on_red;
+    colors_[level::emerg]    = bold | yellow | on_red;
+}
+
+inline void wincolor_sink::log(const details::log_msg& msg)
+{
+    // Wrap the originally formatted message in color codes
+#ifdef _WIN32
```
gabime: what would happen if 2 threads call this ? protection is needed..

RavenX8: Added.

RavenX8:
Don't merge this yet. I'm currently testing.

gabime:
the lines
```c++
 SetConsoleTextAttribute(GetStdHandle( STD_OUTPUT_HANDLE ), colors_[msg.level]);
 SetConsoleTextAttribute(GetStdHandle( STD_ERROR_HANDLE ), colors_[msg.level]);
```

are problematic with threads..
For example one thread writes to stderr color sink while a different thread writes to stdout color sink.
The mutex won't help here since there are two seperate mutexes here ( one of for each sink)

In fact maybe the wrapper stuff should be avoided. 
A copy/paste of the stdout_sink (with the color additions) would be more straight forward in both ansicolor and wincolor sinks

zmaktouf:
Hi,
When this will be merged into main?
Thanks

RavenX8:
I have been testing the code changes and I need to make one more change and it should be ready to merge.

petke:
Is this working and will it be merged soon? If not maybe the documentation should be updated to make it clear that colors do not working on windows.

RavenX8:
This is working, colors change and everything. There is a race condition that I need to fix that can cause colors to change mid print. I need to make sure it's fixed.

gabime:
@RavenX8  I will be happy to help if you need it..

include/spdlog/sinks/ansicolor_sink.h
```diff
@@ -33,6 +33,40 @@ class ansicolor_sink : public sink
     virtual void log(const details::log_msg& msg) override;
     virtual void flush() override;
 
+#ifdef _WIN32
+    void set_color(level::level_enum level, const short& color);
+
+	/// Formatting codes
+    const short reset      = 0;
+    const short bold       = FOREGROUND_INTENSITY;
+    const short dark       = reset; // Not implemented in windows
```
KindDragon: Can you use ANSI sequences on Windows automatically if `ANSICON` environment variable is set, because windows console not support all features. Programs like ConEmu or AnsiCon set this variable when they support ANSI escape codes (http://conemu.github.io/en/AnsiEscapeCodes.html).

include/spdlog/details/spdlog_impl.h
```diff
@@ -61,7 +62,11 @@ inline std::shared_ptr<spdlog::logger> spdlog::daily_logger_st(const std::string
 inline std::shared_ptr<spdlog::logger> create_console_logger(const std::string& logger_name, spdlog::sink_ptr sink, bool color)
 {
     if (color) //use color wrapper sink
+#ifdef _WIN32
+        sink = std::make_shared<spdlog::sinks::wincolor_sink>(sink);
```
KindDragon: Can you use ANSI sequences on Windows automatically if `ANSICON` environment variable is set, because windows console not support all features? Programs like ConEmu or AnsiCon set this variable when they support ANSI escape codes (http://conemu.github.io/en/AnsiEscapeCodes.html).

gabime:
Implemented in 0cfdad4d0b6dde761debe408fc7c9501e697b914

PR #196, opened 2016-04-14
Performance improvement of stdout sink 

gorunovanton:
The current implementation of stdout sink uses std::cout.write, which is really slow.
Writing to stdout via fwrite can increase performance.
I wrote a simple benchmark to compare performance of fwrite(..., stdout) and std::cout.write. fwrite works  more than 28x faster (90 ms VS 2926 ms) in my configuration (Intel Core i7 2630QM, Windows 10, Visual Studio 2015).
Code of benchmark:
```C++
#include <iostream>
#include <chrono>

int main() {
    constexpr int howmany = 1000;
    const std::string message("spdlog message #: This is some text for your pleasure\n");
    
    // warm up
    for (int i = 0; i < howmany; ++i) {
        fwrite(message.c_str(), sizeof(char), message.size(), stdout);
        fflush(stdout);

        std::cout.write(message.c_str(), message.size());
        std::cout.flush();
    }

    const auto fwrite_start = std::chrono::steady_clock::now();
    for (int i = 0; i < howmany; ++i) {
        fwrite(message.c_str(), sizeof(char), message.size(), stdout);
        fflush(stdout);
    }
    const auto fwrite_end = std::chrono::steady_clock::now();

    const auto cout_start = std::chrono::steady_clock::now();
    for (int i = 0; i < howmany; ++i) {
        std::cout.write(message.c_str(), message.size());
        std::cout.flush();
    }
    const auto cout_end = std::chrono::steady_clock::now();

    std::cout << "fwrite: "<< std::chrono::duration_cast<std::chrono::milliseconds>(fwrite_end - fwrite_start).count() << " ms" << std::endl;
    std::cout << "std::cout: " << std::chrono::duration_cast<std::chrono::milliseconds>(cout_end - cout_start).count() << " ms" << std::endl;

    return 0;
}
```

### DIFF
--- a/include/spdlog/sinks/stdout_sinks.h
+++ b/include/spdlog/sinks/stdout_sinks.h
@@ -5,10 +5,9 @@
 
 #pragma once
 
-#include <spdlog/sinks/ostream_sink.h>
 #include <spdlog/details/null_mutex.h>
 
-#include <iostream>
+#include <cstdio>
 #include <memory>
 #include <mutex>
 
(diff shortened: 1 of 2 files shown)


gabime:
This is great. I was a avoiding iostreams like the plague througout the most of spdlog..
It might be interesting to bench in linux too

godbyk:
I set `howmany` to 10,000 and ran the benchmark on Linux and OS X. On OS X `fwrite` was consistently faster. On Linux, it varied a lot. Sometimes they were the same, sometimes `fwrite` was faster, and sometimes it was slower.

gabime:
Interesting.. 
In any case, I am pretty sure the underlying implementation of ```fwrite()``` is way leaner than ```std::cout.write(..)``` in most (if not all) platforms.

godbyk:
No arguments from me.

PR #203, opened 2016-05-03
daily_file_sink with custom file name calculator

eao197:
Standard sinks::daily_file_sink uses hardcoded way of filename calculation. If an user don't like names in form base_YYYY-MM-DD_hh_mm.ext it has to write its own daily_file_sink implementation. But this implementation will have only one difference -- body of calc_filename() method. All other code will be a copy-paste from original daily_file_sink class. I think it is not good.

I have added a possibility to customize filename calculation via additional template parameter for sinks::daily_file_sink class. This parameter has default value. It means that there can be a good compatibility with old code.

With this approach an user can define its own daily_file_name_calculator class with just one static method and use it as a parameter to daily_file_sink class.

I have also extented tests/file_log.cpp file and added some new tests for this functionality. I have tested this change under VC++ 14.0 and MinGW GCC 4.9.3 and 5.3.0 under Windows, GCC 5.3.0 and clang 3.7.1 under ArchLinux, clang 3.4.1 under FreeBSD.

### DIFF
--- a/include/spdlog/sinks/file_sinks.h
+++ b/include/spdlog/sinks/file_sinks.h
@@ -140,9 +140,39 @@ typedef rotating_file_sink<std::mutex> rotating_file_sink_mt;
 typedef rotating_file_sink<details::null_mutex>rotating_file_sink_st;
 
 /*
+* Default generator of daily log file names.
+*/
+struct default_daily_file_name_calculator
+{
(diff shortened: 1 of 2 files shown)


gabime:
Thanks. might be useful. Could you add few lines decribing this in the wiki?

PR #208, opened 2016-05-11
In async mode, teardown callback on worker thread exit

SBAM:
Hello Gabime,

I need to write a customized sink that invokes a JNI callback. As you may know, such calls require that running thread was previously attached to JVM.
This is fine, and I can use warmup callback to attach it.
However, it's also required to detach thread from JVM upon its exit. That's why I'm suggesting adding a new teardown callback that would be invoked before worker thread exit.

Please let me know what you think about this new feature.

Have a good day !


gabime:
Seems fine. Please add some words in the wiki about it, so it get documented..

PR #210, opened 2016-05-13
Fix throwing exception if DYNAMIC_TIME_ZONE_INFORMATION fails

BOOtak:
under windows.

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -209,7 +209,7 @@ inline int utc_minutes_offset(const std::tm& tm = details::os::localtime())
     auto rv = GetDynamicTimeZoneInformation(&tzinfo);
 #endif
     if (rv == TIME_ZONE_ID_INVALID)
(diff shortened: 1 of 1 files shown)


gabime:
thanks. please use the shorter form and remove sstream.h:
```c++
throw spdlog::spdlog_ex(std::string("Failed gettin..")+GetLastError());
```

BOOtak:
@gabime GetLastError() returns DWORD which is typedef for unsigned int. C++ can't concatenate std::string and int. Instead we can use std::to_string(), so, it will look like
```c++
throw spdlog::spdlog_ex("Failed gettin.." + std::to_string(GetLastError()));
```
Should I edit my code to look line example above?

gabime:
@BOOtak you are right. sorry for the confusion. please use your suggested 

```c++
throw spdlog::spdlog_ex("Failed gettin.." + std::to_string(GetLastError()));
```

BOOtak:
@gabime I've updated my commit, you can merge now

PR #211, opened 2016-05-13
RFC: Support custom eol style

hvellyr:
By default use the platform convention (i.e. "\r\n" on windows), but make it possible to set a custom end-of-line string without implementing a custom formatter.  This is not so much for having an end-of-line convention like `<EOL>`, but more to force a specific line convention (e.g. always use `"\n"`, independent of platform).

What do you think?


gabime:
The changes in the code are much too big for such small feature.  I would prefer just to add a option to ```tweakme.h``` file to change the default eol.

something like 
```c++
//uncomment the below to override spdlog's default eol
//#define SPDLOG_EOL "\r\n
```

hvellyr:
Ok since it's really about "\n" vs. "\r\n" and not anything more complicated (where you could, theoretically, imagine having special end-of-line conventions per logger), that should work.  I've change the commit accordingly.

gabime:
Thanks. I merged, but I am planning to move thie ```#ifdef``` to ```details/os.h``` as I am trying to keep all the ugly #ifdefs in in one place as much as possible.

PR #212, opened 2016-05-13
Add custom format strings to pattern_formatter

cedral:
I need custom logging attributes and this method seemed easiest. Thought you might want it back.


gabime:
Thanks. I will might use it in the future if there is a demand..

lilinjn:
I would find this useful and was thinking of writing something similar.

PR #222, opened 2016-05-30
add hourly log

ewanshin:
I've huge daily logs and I can't delete it (at least 6 month) so I suggest save a log file every each hour.

It use 
hourly_file_sink_mt
hourly_file_sink_st

### DIFF
--- a/include/spdlog/details/spdlog_impl.h
+++ b/include/spdlog/details/spdlog_impl.h
@@ -57,6 +57,17 @@ inline std::shared_ptr<spdlog::logger> spdlog::daily_logger_st(const std::string
     return create<spdlog::sinks::daily_file_sink_st>(logger_name, filename, SPDLOG_FILENAME_T("txt"), hour, minute, force_flush);
 }
 
+// Create file logger which creates new file every hour):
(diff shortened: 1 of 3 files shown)


gabime:
Thanks, i will add it if there is demand.

gjvc:
This is great -- please merge to master, gabime :+1:

ldionne:
Actually, I think what we really need is a way to rotate logs every `<arbitrary duration>`. For example, rotate every `std::chrono::minutes{10}`. I work at a place where there's so much log that a hourly log would be useless; we currently rotate every few minutes or so.

gabime:
@ldionne Sounds that the size based rotating logger is more suitable in this case

ldionne:
I guess you're right (that's what I'm using), but I think the point remains that it would be very easy to extend this PR to support arbitrary time durations.

jzakrzewski:
Is there any chance of this being merged?

mmanoj:
> 

I implemented base sink for minutes, however getting some errors like this while calling the sink.

mmanoj:
> I guess you're right (that's what I'm using), but I think the point remains that it would be very easy to extend this PR to support arbitrary time durations.

https://github.com/gabime/spdlog/issues/2727

PR #236, opened 2016-07-08
No streams

gabime:


godbyk:
Hi, @gabime.

Can you expound a bit on the problems with the stream operators and why they're being removed? 

I'm asking because I rather prefer using them. I don't mind implementing them outside of spdlog, but if there are known issues with them, I'd love to know what they are so I can determine if  I can avoid them or not.

Thanks!

gabime:
There are several reasons:

1. The code need the line_logger destructor to do the important stuff (because this is the only place the we can know that the stream statement ends) . if somehing goes wrong and an exception is thrown, bad things happen (as we all know throwing from dtor is bad)

2. The code is much more comlicated than it should be. In particular it is very hard to deal with errors correctly since they might occur in many placed (like in the middle of a stream operation : for example : loggere<<x<<y<<z can fail in logging x or y or z and it is very hard to add in spdlog correct error handling.

3. spdlog sohuld provide single way of doing things. providing multiple ways (fmt+streaming) to achieve the same thing is not good Imo.

 4.  I always prefer keep thing as simple as possibe and removing them made the code so much simpler.
 Now for example it will be much easiar to perform the formatting in the async thread and to provide the much needed error handling thru callbakcs instead of exceptions.

godbyk:
Thanks for the informative response!

For (1), could we wrap the body of the dtor in `try...catch` and call the (`noexcept`) error handler in the `catch` block?

Argument (2) is a really good point. I'll have to examine the code more closely to see if I can find a good option there.

I also understand (3). I really do like the stream feature, though, so I may spend a little time to see if I cn find a way of making it work. (Whether that work lives in spdlog or in my own code base doesn't matter too much. Would you be open to having an optional header that provided stream support if it were possible?)

gabime:
> Would you be open to having an optional header that provided stream support if it were possible?

Nope. I prefer keeping things as small and as simple as possible.

Of course, you could always add/wrap with external code, but I doubt it is a good idea.

jgornowich:
@gabime I use this logger for a separate project I am working on and just pulled updates.  I noticed that the stream API was deprecated which I understand your reasoning behind that, but in these commits you have also removed some of the logging levels.   I would imagine this was done for simplification and would just like to verify before making changes since I could not find any issue or suggestion about removing these.

gabime:
@jgornowich yes, i removed some of the levels for further simplification.
 I should have done it in a separate commit though.. sorry for the confusion..

dontsovcmc:
Terrible :(. We need to refactor thousand calls..

gabime:
@dontsovcmc sorry man.. it was a tough decision for me too

daravi:
@gabime I know this issue is old. But I think having a python script in this repo to help refactor GLog macros to spdlog function calls would be useful for people wanting to make the switch. I am making one that seems to work for our code-bases (applying a bunch of regex find/replace). But I am not a regex guru so I am sure others can do a better job.

SimpleMango:
> @gabime I know this issue is old. But I think having a python script in this repo to help refactor GLog macros to spdlog function calls would be useful for people wanting to make the switch. I am making one that seems to work for our code-bases (applying a bunch of regex find/replace). But I am not a regex guru so I am sure others can do a better job.

I use copilot to  replace the log code

PR #260, opened 2016-08-16
Use wchar filenames in spdlog

jasonjei:


jasonjei:
Oop, wrong repository! Sorry.

PR #261, opened 2016-08-16
level color

snaiffer:


gabime:
full formatter affects all sinks not just terminal.. this change will  add color codes in the log files

snaiffer:
I suppose, it'll be useful sometimes, having color output in log files too. 
For example, during debugging, we run some apps and open log with '$ tail -f log.txt' command (it outputs the log in real-time). It's very hard to detect any problem with help of one-color text. But it's very easy if we will see something like "[<red-color highlight>error</red-color highlight>] ..."
Of course, this feature should be switchable. (I have to improve my pull-request if you'll like this idea)

What do you think about it?

gabime:
You could always use [grep with color ](http://linuxcommando.blogspot.co.il/2007/10/grep-with-color-output.html) or something for this.  For example grep --color=always -R "[error]" * | less

Issue #134, opened 2015-09-27
spdlog scatters logs when caller do chdir


objectx:
When user pass the relative path to `spdlog::rotate_logger_??` and rotation was occured, `spdlog` simply create a log relative to current directory.

Is it intended ?

# I've created the [example](https://github.com/objectx/spdlog-rotate-test).

gabime:
It is not intended.

(I think) it can be considered a bug. Thanks for reporting

agauniyal:
Still a bug or should be closed?

gabime:
It still exists but I dont think it is really an issue. Since relative path is defined as relative to current working dir spdlog behaves correctly. A full path should be used if this is not a desirable behavior

PR #262, opened 2016-08-16
Fix issue #134

jks-liu:
- Add function `get_cwd` to get current directory
- Record the absolute path when create `spdlog::rotate_logger_??`

And I executed the code on Windows (Both ANSI and UNICODE version) only.

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -187,7 +187,25 @@ inline bool file_exists(const filename_t& filename)
 }
 
 
-
+inline filename_t get_cwd()
+{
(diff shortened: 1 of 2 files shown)


gabime:
Thanks for the pull. There are several issues with it though:

-FILENAME_MAX is not of enough size. it should be maxpath or similiar(directory path can be much larger than FILENAME_MAX)

- spdlog_ex expect errno and not the GetLastError() retval

- This will break if user already gives full path to the logger

- This should use the right seperator according to the os(\ and not / under windows)

PR #263, opened 2016-08-22
Add support for Solaris operating system

jan-dx:
In current form, library can't be compiled on Solaris operating system because of 'undefined symbol' error. This pull request adapt library to be usable on Solaris.


gabime:
Thanks. With which version of Solaris and GCC have you tested ?

I like the SPDLOG_ENABLE_SYSLOG macro. I think I will just disable syslog by default, since most users don't use syslog and won't have to compile this feature again and again.

I will merge and then update my changes too.

jan-dx:
Solaris version:
`bash-3.2$ uname -a
SunOS unknown 5.10 Generic_147441-01 i86pc i386 i86pc`

GCC version (from OpenCSW repository):
`bash-3.2$ g++ -dumpversion
5.2.0`

gabime:
Updated in 20cb73e9da293076d76a581308deef852700354a (disable syslog by default)

Issue #275, opened 2016-09-11
can async_logger use a thread pool instead of a single thread per logger?


facontidavide:
Hi,

apparently each async_logger has its own thread.
Since the queue used is already a lock free multi producer / multi consumer queue, it should be relatively easy to have a SINGLE queue shared by all of the async_loggers.
One or multiple dedicated worker threads can consume the elements of the queue.

Is there any technical reason that makes this hard to implement?

gabime:
The only problem i can think of is that flush() would be way more complicated. In current version it waits until the queue is empty fefore it reurns to make sure thall the messages in it got processed.

facontidavide:
Good point.
In that case it might sense to have a a single queue for each logger and the threads in the pool consuming those queues using something similar to a simple round robin.
I will see if I can implement a simple solution.

gabime:
@facontidavide Please take a look at v1.x branch. There is a shared thread pool for all loggers

vantheo:
I have a related question, but post to a less related issue.

https://github.com/gabime/spdlog/issues/2202#issuecomment-2340059291

PR #276, opened 2016-09-11
Async task queue

facontidavide:
Hi,

accidentally and for reasons I completely ignore, I might have improved the speed to the asyn_logger.

I modified the lock free queue. Instead of an `async_msg`, it now contains a `std::function<void()>` that works as a "packaged task" that executes the method equivalent to `process_next_msg`.

These changes are needed to implement a thread pool (https://github.com/gabime/spdlog/issues/275), but when I executed the unit tests and benchmark, I realized that the outpout is (apparently) the same but the speed of async is considerably higher... on my computer!

### DIFF
--- a/include/spdlog/details/async_log_helper.h
+++ b/include/spdlog/details/async_log_helper.h
@@ -45,6 +45,7 @@ class async_log_helper
         flush,
         terminate
     };
+
     struct async_msg
     {
         std::string logger_name;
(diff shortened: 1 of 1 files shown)


facontidavide:
sorry. It is actually slower, not faster.

gabime:
Could it be that the queue is faster when holding small objects like function pointers?
Please post the the benchmark results and the platform details.

PR #277, opened 2016-09-12
Fix compilation on Android.

hghwng:
2 errors will occur during `ndk-build`, in the original version.

```
In file included from spdlog/include/spdlog/details/log_msg.h:9:0,
                 from spdlog/include/spdlog/sinks/sink.h:9,
                 from spdlog/include/spdlog/sinks/base_sink.h:13,
                 from spdlog/include/spdlog/logger.h:15,
                 from spdlog/include/spdlog/spdlog.h:13,
spdlog/include/spdlog/details/os.h: In function 'std::string spdlog::details::os::errno_str(int)':
spdlog/include/spdlog/details/os.h:352:57: error: invalid conversion from 'int' to 'const char*' [-fpermissive]
     return std::string(strerror_r(err_num, buf, buf_size));
```

```
spdlog/include/spdlog/fmt/bundled/format.h:3263: error: undefined reference to 'localeconv'
```

The pull request fixes the errors above.

### DIFF
--- a/include/spdlog/details/spdlog_impl.h
+++ b/include/spdlog/details/spdlog_impl.h
@@ -14,6 +14,7 @@
 #include <spdlog/sinks/stdout_sinks.h>
 #include <spdlog/sinks/syslog_sink.h>
 #include <spdlog/sinks/ansicolor_sink.h>
+#include <spdlog/sinks/android_sink.h>
 
 #include <chrono>
 #include <functional>
@@ -101,6 +102,13 @@ inline std::shared_ptr<spdlog::logger> spdlog::stderr_logger_st(const std::strin
 inline std::shared_ptr<spdlog::logger> spdlog::syslog_logger(const std::string& logger_name, const std::string& syslog_ident, int syslog_option)
 {
     return create<spdlog::sinks::syslog_sink>(logger_name, syslog_ident, syslog_option);
+}
+#endif
+
+#if defined(__ANDROID__)
+inline std::shared_ptr<spdlog::logger> spdlog::android_logger(const std::string& logger_name, const std::string& tag)
+{
+  return create<spdlog::sinks::android_sink>(logger_name, tag);
 }
 #endif
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-# Auto generated files 
+# Auto generated files
 *.slo
 *.lo
 *.o
@@ -46,6 +46,7 @@ example/*
 !example/example.vcxproj
 !example/CMakeLists.txt
 !example/multisink.cpp
+!example/jni
 
 # generated files
 generated
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -13,6 +13,7 @@
 
 void async_example();
 void syslog_example();
+void android_example();
 void user_defined_example();
 void err_handler_example();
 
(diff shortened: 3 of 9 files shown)


gabime:
thanks for the pull. but please send it to fmtlib directly, since spdlg does not maintain format.h, it just budles it

hghwng:
Thanks for the notice, now only `os.h` is included. I'll submit the patch of  `format.h` to upstream.

hghwng:
BTW, IMHO managing third-party libs by `git submodule` will be easier to keep up with upstream. It would also clearer for me to figure out where to submit the pull request related to fmtlib.

godbyk:
You'll also need to update `android_sink.cpp` to (1) not throw exceptions, and (2) remove the deprecated log levels. (See [this diff](https://github.com/gabime/spdlog/compare/master...OSVR:android-sink) for my untested implementation.)

hghwng:
@godbyk I forgot to write `#ifdef` to switch between the android sink and the stdout sink in different platform!

After switching to the android sink and removing the deprecated log levels, it compiles.

I don't think it's necessary to remove exceptions in `android_sink.h`:

- Other parts of spdlog uses exception. It's not enough just removing exceptions in `android_sink.h`.
- Execption handling can be enabled by passing `-fexceptions` to the compiler and using a execption-supported C++ runtime.

godbyk:
spdlog now guarantees that it won't throw exceptions except during construction, though. So the `android_sink` needs to maintain that guarantee.

gabime:
its ok for sinks to throw since the logger code will catch any exceptions from them and call the error handler.

gabime:
@hghwng Could you please add to this pull the fixes to the android sink?

hghwng:
@gabime OK, I'll fix it and test to make sure it works on Android. I guess I'll finish it by this weekend.

hghwng:
@gabime Please review the pull request. BTW, are there any plan of converting fmtlib to a submodule? It's kinda awkward to write "Sync with fmtlib to fix Android compilation."

include/spdlog/details/spdlog_impl.h
```diff
@@ -14,6 +14,7 @@
 #include <spdlog/sinks/stdout_sinks.h>
 #include <spdlog/sinks/syslog_sink.h>
 #include <spdlog/sinks/ansicolor_sink.h>
+#include <spdlog/sinks/android_sink.h>
```
gabime: Please wrap with ```#if defined(__ANDROID__)``` here

include/spdlog/fmt/bundled/format.h
```diff
@@ -3260,7 +3260,10 @@ void BasicWriter<Char>::write_int(T value, Spec spec)
     case 'n':
     {
         unsigned num_digits = internal::count_digits(abs_value);
-        fmt::StringRef sep = internal::thousands_sep(std::localeconv());
+        fmt::StringRef sep = "";
+#ifndef ANDROID
```
gabime: Is this fix is already in fmt lib ? If yes I will take it from there directly, so please remove it from this file

gabime:
@hghwng Thanks for the fix. I added few comments. About the fmt fix. I will take directly from fmt.

include/spdlog/details/spdlog_impl.h
hghwng: `android_sink.h` has already been guarded with `#if defined(__ANDROID__)`. I think it's redundant to test twice.

include/spdlog/fmt/bundled/format.h
hghwng: My fix has been merged to fmtlib. I'll remove it.

include/spdlog/details/spdlog_impl.h
gabime: Yes I know but it looks weird that the file even gets included.

gabime:
Updated to latest fmt version (https://github.com/gabime/spdlog/commit/6312748cc73b8f6429e3b5ebf5fd508aa3e42345)

PR #278, opened 2016-09-12
New approach to async: use thread pool to consume the queues.

facontidavide:
Hi,

this is a PR related to https://github.com/gabime/spdlog/issues/275

The new implementation use a **thread pool** with 1 or more threads to serve ALL the `async_loggers` in the process.

There is still a single queue for each logger; error handling, flush and push should work as usual.
All the unit tests pass and the benchmark has the same throughput as before.

The way the thread_pool consumes the messages in the queue(s) is a bit counter-intuitive, but is necessary if we don't want to modify the semantic of `flush()` and queue_size.

1) Each `async_logger` passes function pointer (called further "handle") pointing to **`spdlog::details::async_log_helper::worker_loop()`** to the pool. By the way, that method is not a loop anymore. 
2) using **std::weak_ptr**, the pool is able to disconnect automatically the handle of any destroyed instance of `async_log_helper`.
3) The workers cycle through the available handles (one per async_logger) to find one that needs to be served.

It is quite a large change and I am not 100% sure that you want to accept this PR without any further modification. The user facing APi and the expected behaviour should be the same, though.

### DIFF
--- a/include/spdlog/async_logger.h
+++ b/include/spdlog/async_logger.h
@@ -15,6 +15,7 @@
 //    3. will throw spdlog_ex upon log exceptions
 // Upon destruction, logs all remaining messages in the queue before destructing..
 
+#include <spdlog/details/thread_pool.h>
 #include <spdlog/common.h>
 #include <spdlog/logger.h>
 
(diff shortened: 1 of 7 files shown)


gabime:
Thanks. I will need to find the time to review and test such big change, but will be done

include/spdlog/async_logger.h
```diff
@@ -39,26 +40,23 @@ class async_logger :public logger
                  const It& begin,
                  const It& end,
                  size_t queue_size,
```
facontidavide: Note: worker_warmup_cb and worker_teardown_cb were moved to the thread pool.

include/spdlog/details/async_log_helper.h
```diff
@@ -218,7 +216,11 @@ inline spdlog::details::async_log_helper::~async_log_helper()
     try
     {
         push_msg(async_msg(async_msg_type::terminate));
```
facontidavide: New approach: flush and wait for the thread pool to consume the queue.
Tell me what you think about this.

include/spdlog/details/async_log_helper.h
```diff
@@ -257,15 +256,11 @@ inline void spdlog::details::async_log_helper::flush(bool wait_for_q)
         wait_empty_q(); //return only make after the above flush message was processed
 }
 
-inline void spdlog::details::async_log_helper::worker_loop()
+inline bool spdlog::details::async_log_helper::worker_loop()
 {
     try
     {
```
facontidavide: This is not a loop anymore. it is just a method wrapped into a function pointer that will be executed by the threads in the thread pool.

include/spdlog/details/registry.h
```diff
@@ -131,15 +134,23 @@ template <class Mutex> class registry_t
         _err_handler = handler;
     }
```
facontidavide: worker_warmup_cb and worker_teardown_cb are now shared by ALL the threads in the pool.
I wonder if this could be a problem for some users or not.

Issue #79, opened 2015-04-16
log level by sink


nmarwen:
I think It can be interesting to have log levels and patterns by sink in addition to the definition by logger.

p0fi:
+1 for this!

It would make it possible to create a logger with multiple sinks (e.g. console output and a rotating file) and write everything in the console, but only the errors in the file!

I'd love to see this feature!

gabime:
I think that in order to keep things consistent and simple, a good approach would be for spdlog to provide a combined logger instead - a logger that holds a list of loggers to forward message to. 
Something like
```c++
auto console = spdlog::console_logger_mt("console");
console->set_level(level::info);

auto file_logger = spdlog::simple_file_logger_mt("filelogger", "filename");
file_logger->set_level(level::debug);
file_logger->set_pattern("some pattern %t %v");

auto combined = spdlog::combined_logger("combined", console, file_logger);
combined->debug("Should not appear on the screen, only in the log file");

```

p0fi:
I can't find a way to combine loggers into one using the current version. Is that something you already implemented or you are going to do this?

As long as this loggers are still thread safe and globally useable like the ones I create right now with this:
```
auto Debug = make_shared<spdlog::logger>("someName", begin(sinks), end(sinks));
spdlog::register_logger(EODebug);
```

it seems to me a rather good approach.

You could set a own pattern and level for this logger, which would basically serve every need I mentioned.

gabime:
I created a [branch](https://github.com/gabime/spdlog/tree/sink_level_support) with support for setting levels on sinks.

Please try and give me feedback on this.

example:
```c++

//first sink with info level
auto sink1 = std::make_shared<spdlog::sinks::stdout_sink_st>();
sink1->set_level(spdlog::level::info);
//second sink with debug level
auto sink2 = std::make_shared<spdlog::sinks::daily_file_sink_st>("logfile", "txt", 23, 59);
sink2->set_level(spdlog::level::debug);

//combined logger with debug level
auto combined = std::make_shared<spdlog::logger>("combined", spdlog::sinks_init_list{ sink1, sink2 });
combined->set_level(spdlog::level::debug);
spdlog::register_logger(combined); //Needed if you want to use spdlog::get("combined") elsewhere

```

The problem here is that a bit confusing - both logger and sinks can now have log levels. 

[...]

PR #279, opened 2016-09-12
Loglevel per sink

facontidavide:
Added a log_level per sink. See discussion here https://github.com/gabime/spdlog/issues/79
To be logged, a message should pass the check of the logger AND the sync.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ example/*
 !example/example.sln
 !example/example.vcxproj
 !example/CMakeLists.txt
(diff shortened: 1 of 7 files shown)


gabime:
Thanks looking good. I would also make the sinks vector const public while we are at it (so sink levels can be changed more easily) . 
At first  glance I thought that the should_log check should be done inside the sink itself(in the log function) but then considering that log(..) is virtual.. better call should_log from outside :+1:

PR #288, opened 2016-09-23
SpdLog as DLL and logs in wchar_t ...

vwbij:
Changes I have done:

• Added a bitmask to filter the logs. Now it is possible to use the bitmask and/or the loglevel to filter messages. You can disable this feature with the switch “SPDLOG_BITMASK_LOG_FILTER” in “tweakme.h”.
• Support for widechar format strings and logfiles, activated by  setting “SPDLOG_WCHAR_LOGGING” in “tweakme.h”. This selects the format of the format strings and of the logfiles. For Unicode logfiles a correct BOM is written as first character of the file. 
• Created a DLL with a C-Syle call interface “spdloglb.dll”. To do so, I’ve done the following:
• Created the folder “include\spdlog\cstyle” with “logger_handle.h”, “sink_handle.h”
• Created c-style interface under “spdloglb”, “spdloglb.cpp/h”
• Interface to format strings using an object and stream operators: “spdloglb_vararg.cpp/h”, “vararg_list.h”, “vararg_listitem.h” under “include\spdlog\cstyle”
• Taking out essential types from “common.h” to “common_types.h”: minimize informations required for the DLL interface.

• “Spdloglb” contains the project and wrapper codefiles for the DLL, the core itself is taken from “include\spdlog”.
• “SpdLogLbTest” contains the solution file for testing “spdloglb.dll” and the project file for “tests” that is an modified version of your tests project.
• “SpeedTest” contains my profiling project, integrated in the “spdlogLbTest”-solution, building into “SpdLogLb<config>”.

Have Fun,

Volker Bijewitz


gabime:
Thanks for this big pull. I might use parts of it.
Please note that the ci tests fail in both travis and appveyor

vwbij:
Ok, you are right, tests/tests.sln is failing to translate, it misses some adaptions to the new datatypes. "spdloglbtest/tests.sln" is translatable and works like a charm, the subprojects "SpeedTest" and "tests" from this project was my test base. I do the neccessary changes to the "tests/tests.sln" project ASAP that you may test the code without the dll.

vwbij:
Ok, have fixed the issues with test. It was caused mostly by a uncommon configuration in tweakme.h (wchar_t filenames, logging with chars) which I have not tested with this project. Ok, now it works in the checked in configuration, the tests run without errors. 

I've pushed the cahnges to my branch on GitHub. Should I issue another pull request? Oh, GIT is not very common to me, sorry!

pangpang7:
replace some NULL with nullptr can avoid some compile error(C2893) in VS2015.

vwbij:
Hi pangpang7, where do you have found C2893 errors? I have build all 4 projects with VS2015 and have not seen any C2893 error.

pangpang7:
I found 2 errors as below when building spdloglb project. Maybe some differnce with ur VS2015 like SDK or sth...

line 109, spdloglb.cpp
spdlog::details::registry::instance().set_async_mode(QueueSize, opc, NULL, fims, NULL);

line 48, spdloglb.cpp
spdlog::set_error_handler(NULL);

vwbij:
Ok, thank you, I have corrected it. I even do not get a warning compiling the code. No idea, I use VS 2015 Build 14.0.25425.01, update 3. I started using it just for spdlog, mostly I use VS 2010.

pangpang7:
Thank you for your correction. Another question, I have built the project spdloglb and spdloglbtest, but when I run the test, found that the sink file cannot be created at line 12,13 in CStyle.cpp.
Code:
Sinks[0] = spdlog::SpdLog_CreateFileSink(_SFT("logs/simple_log.txt"), false, true);
Sinks[1] = spdlog::SpdLog_CreateFileSink(_SFT("logs/simple_log2.txt"), false, true);
prompt:
...\utils.h<25>:
warning:
     Fail opening file logs/simple_log.txt for writing: No such file or directory.

But if I change the code from "logs/simple_log.txt" to "simple_log.txt". It seems ok. Could you have a look at these code? Thanks!

vwbij:
Hi pangpang7,
I have seen this behaviour before. You need to create the folder "logs" before you are able to run the test. You need to create this folder in the current directory where the exe runs. That means also that you need another folder if you run the exe under the debugger, because under the debugger the project folder is set as current folder, not the output folder.
For me, I like it more to have it in a subfolder. Maybe add some empty folders into the project to avoid this pitfall?

pangpang7:
I see. But maybe add auto mkdir function before invoke fopen will be useful for user. I add some code in the function
**inline int fopen_s(FILE** fp, const filename_t& filename, const filename_t& mode)** in os.h.

#ifdef _WIN32
#ifdef SPDLOG_WCHAR_FILENAMES

	wchar_t drive[_MAX_DRIVE];

	wchar_t dir[_MAX_DIR];

	wchar_t fname[_MAX_FNAME];

	wchar_t ext[_MAX_EXT];

	_wsplitpath((filename.c_str()), drive, dir, fname, ext);

	std::wstring path = drive;
	path += dir;

	_wmkdir((path.c_str()));

    *fp = _wfsopen((filename.c_str()), mode.c_str(), _SH_DENYWR);

it can make dir automatically before create file. by the way, I just add the code in WIN32 &&　SPDLOG_WCHAR_FILENAMES．

pangpang7:
Another question, I can log sth like "12345", but I want to log it with some pre-decoration like "yy/mm/dd hh:mm:ss [warn/info/error...] 12345". Could you please advise me that how can I achieve this by invoke the api? Thanks!

vwbij:
Hi Pangpang7, you find the formatstring in the struct "SPDLOGLB_CREATELOGGER_PARAM" that is issued in the functions that creates a logger. You can also use SpdLog_SetLoggerFormatPattern. You find the pattern on https://github.com/gabime/spdlog/wiki/3.-Custom-formatting.

PR #297, opened 2016-10-09
Windows-specific console coloring.

SlavMFM:
Previous generous (with console codes) technique not working on Windows (having **□m03h** symbols or something at the beginning of every message).
_WIN32 macro handles x64 architectures as well (_WIN64 definition not needed).
For other platforms functionality remained untouched.
Please refer to 2 TODOs:
    - get rid of **static** declaration
    - introduce mutex to ensure sequential calls
I'm lacking knowledge about project structure to implement those TODOs on my own.


gabime:
thanks, but there is already work for in progress to solve this issue(will be hopefully commited soon)

SlavMFM:
Good to know.
Thanks for your work.

PR #301, opened 2016-10-27
made the opened file handle not inherited by child process on windows

memetea:
hi,
I faced a problem when using CreateProcess to create a child process in windows.  the child process was left untouched after was launched and will not exit.  when i restart my program i maybe not able to open app log because the file handle is held by the child process i create.    so maybe it's better to let the opened log file handle not be inherited by child process.

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -25,7 +25,7 @@
 #define WIN32_LEAN_AND_MEAN
 #endif
 #include <windows.h>
-
+#include <io.h>
 #ifdef __MINGW32__
 #include <share.h>
 #endif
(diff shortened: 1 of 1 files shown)


gabime:
But the the opposite might be needed too: Inherit the open fd so same logger instance can be used by child procecess

memetea:
maybe add options to control this behavior?   default was inherit.
that take in some complexity

gabime:
@memetea I am not sure if it worth the extra complexity. This is a rare case I think

gabime:
Implemented in 92db8115b77d2a63ceb8fa6741f5d7650c1ab299
To enable go to tweak,me.h and uncomment:
```c++
#define SPDLOG_PREVENT_CHILD_FD
```

PR #302, opened 2016-10-28
fix missing mutex and template parameter in the wincolor_sink

lexxxel:


lexxxel:
the problem is, that spdlog won't compile with the mingw32 5.3 32bit, qt bundles.
I saw what you mean and remove the additional mutex

gabime:
Implemented in wincolor_sink

PR #305, opened 2016-11-03
Fix Klockwork compilation warning

chenhayat:


chenhayat:
Can you please use the latest commit of fmt library?
Now it contains fixes for klockwork.
Thanks.

PR #306, opened 2016-11-03
Windows XP compatibility hack

emazv72:
This tries to fix a corner case, it avoids the dreadful `_ecvt_s could not be located` runtime error message when running on old xp machines.
Unfortunately many SMB still run on it.

Thanks

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -30,6 +30,14 @@
 #include <share.h>
 #endif
 
+#define SPDLOG_SECURE_API
+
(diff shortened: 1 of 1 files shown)


include/spdlog/details/os.h
```diff
@@ -30,6 +30,14 @@
 #include <share.h>
 #endif
 
+#define SPDLOG_SECURE_API
+
+#ifdef __MINGW32__
```
emazv72: Compiled on windows 10 with latest mingw and tested  on windows xp

Issue #321, opened 2016-11-25
Compiler errors in "wincolor_sink.h"


onnodb:
When trying to compile spdlog under either MinGW32 v5.3 or MSVC x64 v14.0 on Windows 7 x64, I'm getting the following compilation errors:
```
In file included from ../../third_party/spdlog/details/spdlog_impl.h:20:0,
                 from ../../third_party/spdlog/spdlog.h:177,
                 [snip]
../../third_party/spdlog/sinks/wincolor_sink.h: In member function 'void spdlog::sinks::wincolor_sink<Mutex>::set_color(spdlog::level::level_enum, WORD)':
../../third_party/spdlog/sinks/wincolor_sink.h:69:37: error: '_mutex' was not declared in this scope
         std::lock_guard<Mutex> lock(_mutex);
                                     ^
../../third_party/spdlog/sinks/wincolor_sink.h: In constructor 'spdlog::sinks::wincolor_stdout_sink<Mutex>::wincolor_stdout_sink()':
../../third_party/spdlog/sinks/wincolor_sink.h:94:28: error: class 'spdlog::sinks::wincolor_stdout_sink<Mutex>' does not have any field named 'wincolor_sink'
     wincolor_stdout_sink():wincolor_sink(GetStdHandle(STD_OUTPUT_HANDLE))
                            ^
../../third_party/spdlog/sinks/wincolor_sink.h: In constructor 'spdlog::sinks::wincolor_stderr_sink<Mutex>::wincolor_stderr_sink()':
../../third_party/spdlog/sinks/wincolor_sink.h:108:28: error: class 'spdlog::sinks::wincolor_stderr_sink<Mutex>' does not have any field named 'wincolor_sink'
     wincolor_stderr_sink():wincolor_sink(GetStdHandle(STD_ERROR_HANDLE))
                            ^
../../third_party/spdlog/sinks/wincolor_sink.h: In instantiation of 'spdlog::sinks::wincolor_stdout_sink<Mutex>::wincolor_stdout_sink() [with Mutex = std::mutex]':
```

gabime:
fixed in merge #322

PR #322, opened 2016-11-25
Fix compilation errors in "wincolor_sink.h"

onnodb:
This should fix issue #321 (just reported).


onnodb:
I hope this helps, even if only a teeny-teeny bit, to make spdlog better. We're really enjoying using it, so thanks for your efforts!

gabime:
thanks @onnodb

PR #330, opened 2016-12-04
Use the feature checking macros to detect thread_local support in clang.

PMExtra:
I have re-examined the clang's official documents.

[http://clang.llvm.org/docs/LanguageExtensions.html#builtin-macros](http://clang.llvm.org/docs/LanguageExtensions.html#builtin-macros)
It told me that, the marketing version numbers should not be used to check for language features, as different vendors use different numbering schemes. Instead, use the [Feature Checking Macros](http://clang.llvm.org/docs/LanguageExtensions.html#langext-feature-check).

[http://clang.llvm.org/docs/LanguageExtensions.html#c-11-thread-local](http://clang.llvm.org/docs/LanguageExtensions.html#c-11-thread-local)
There is a official example to determine if support for thread_local variables.

[http://clang.llvm.org/docs/LanguageExtensions.html#has-feature-and-has-extension]
(http://clang.llvm.org/docs/LanguageExtensions.html#has-feature-and-has-extension)
And there is a example to compatibility with non-clang compilers.

Ps: Why the marketing version numbers should not be used to check for language features?
I found a special thing. The llvm-clang official version just updated to 4.0, but the build-in version of Xcode is consistent with Xcode version.

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -48,6 +48,10 @@
 
 #endif
 
+#ifndef __has_feature       // Clang - feature checking macros.
+#define __has_feature(x) 0  // Compatibility with non-clang compilers.
+#endif
+
 namespace spdlog
 {
 namespace details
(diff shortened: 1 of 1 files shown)


gabime:
Thanks for the pull

PR #348, opened 2017-01-11
Add prefix for BUILD_TESTING cmake option

w0land:
This is helpful when using spdlog as a dependency (git submodule) when a
master project is not interested in spdlog tests. Using
"BUILD_TESTING" name may create a confusion.
Extra: BUILD_EXAMPLE variable already have a prefix.


gabime:
Its a good idea. 
Thanks for the pull.

w0land:
No problem. 
Thanks for a great library!

PR #349, opened 2017-01-12
Replace typedef with templated using statements

harrywong:
This helps crate daily_file_sink like `daily_file_sink_mt<dateonly_daily_file_name_calculator>`.


gabime:
Hi, Thanks for the pr. 
Why replace the ```typedef``` with ```using```?

harrywong:
When use typedef, we cannot write code like `daily_file_sink_mt<dateonly_daily_file_name_calculator>`, but need to write `daily_file_sink<std::mutex, dateonly_daily_file_name_calculator>. It need to look into the source code and write more words.

Anyway, just for convenience.

PR #353, opened 2017-01-16
Adding a property-based format flag

andreadps:
Working on a video processing project, I need to give context to log lines with the frame number that generated the messages. I could build my own `spdlog::formatter` but it would be mostly a copy of `spdlog::pattern_formatter`. Extending it is not possible because of private members.

Instead I went with mimicking something I have seem on log4net: the capacity to use dynamic, key-value-based properties inserted in the custom pattern string. The format is `%{property_name}.` Each logger holds a <string, string> map with O(1) access and insertion. Users can handle it freely through a shared_ptr.

I have also added a few lines to example.cpp and updated the Visual Studio solution to VS2015.

Hope you consider this contribution and I am open to discuss modifications.

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -57,6 +57,13 @@ int main(int, char*[])
         spd::set_pattern("*** [%H:%M:%S %z] [thread %t] %v ***");
         rotating_logger->info("This is another message with custom format");
 
+        // Using a per-logger dynamic property map
+        console->set_pattern("*** [%H:%M:%S %z] [thread %t] [%{property_name}] %v ***");
+        auto properties = console->properties();
+        (*properties)["property_name"] = "some property";
+        console->info("This message has a custom property");
+        (*properties)["property_name"] = "new property";
+        console->info("The property has changed");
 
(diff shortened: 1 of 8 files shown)


gabime:
Thanks for the pull (sorry for the late response).
I will review and consider adding this.

andreadps:
Hi Gabi. Any chance you would still consider this PR?

gabime:
@andreadps I am not sure needed this will be. I never came across such mechanism in the past..
Alos, If I recall correctly, there are some thread safety problems in this pull .. but thats a different issue..

What do you think? Is there a big demand for such a feature? Does it worth the extra complexity?

andreadps:
I can't say there is a big demand, but there is certainly some. Log4net has `%property{varname}`, Log4j has `%X{varname}`.

My use case is that of a system where the logger is called from very diverse classes, but all in the same context (or session). Properties are then used to carry some context info to the log. I can easily see others benefiting from this.

I will rebase and check those concurrency problems.

PR #357, opened 2017-01-23
Fix get size of mpmc queue issue

harrywong:
Because the `first_pos` may overflow the size_t and come to zero again. So in that critical condition, `last_pos` is less than `first_pos` and the size is not zero. Simply don't check this comes to the right result.


gabime:
This condtion is to prevent the returned size_t to overflow  (unsigned, cannot be negative).

harrywong:
But when you add many many items, the `first_pos` can be zero, yes, it's unsigned, so it can be reach SIZE_T_MAX and then be zero. In this time, `first_pos` is less than `last_pos` and it's the right situation, and the queue is not empy.

And in this time, `first_pos` minus `last_pos` still will get right result, because those are all unsigned.

PR #366, opened 2017-01-31
README.md homebrew instructions

ahundt:


gabime:
Thanks! I think though it is just enough to mention the 'brew install spdlog' somewhere in the readme. I will take care of that.

PR #370, opened 2017-02-15
Replace static global std::string arrays by Meyer singletons. 

jcelerier:
This improves thread-safety : we have the guarantee from the standard that the variable will be initialized on first use.

See http://stackoverflow.com/questions/1661529/is-meyers-implementation-of-singleton-pattern-thread-safe

### DIFF
--- a/include/spdlog/details/pattern_formatter_impl.h
+++ b/include/spdlog/details/pattern_formatter_impl.h
@@ -80,7 +80,7 @@ static int to12h(const tm& t)
 
 //Abbreviated weekday name
 static const auto& days() {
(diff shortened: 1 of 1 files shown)


gabime:
I am not sure.. They are already static variables. According to [this](url), it is already thread safe in c++11

> The C++ standard guarantees that initialization of local statics, no matter how complex, is thread-safe, in that the initialization code will run exactly once, and no thread will run past the initialization code before initialization is complete.

jcelerier:
Ah, but they are not **local** statics (i.e. inside the code of a function), but global ones. See http://stackoverflow.com/questions/12186857/on-local-and-global-static-variables-in-c

gabime:
Hmm.. ok.. In anyway some of the tests fail on this pull. Can you please fix?

jcelerier:
yes, sorry !

PR #375, opened 2017-02-28
Don't hardcode '.txt.' log file name suffix (resolve #333)

### LINKED ISSUE #333: Unexpected appending of "txt" at the end of a log filename
I was confused by the usage of the filename parameter for `rotating_logger_mt`. It's called `filename`, so I interpreted that as being the absolute path to the file. Instead, it seems like `.txt` is appended onto the end of whatever is provided, so I ended up getting something like `mylog.txt.txt`. Is there a reason that the extension is hardcoded to `.txt`? Why not allow the entire filename to be specified by the caller?

As a result of this behaviour, I have to implement two functions in my code: one to output the desired log filename *without* an extension (to provide to `spdlog`), and one to output the log filename *with* an extension (when I want to operate on the log file myself).

horar:
Signed-off-by: Ľubomír Carik <Lubomir.Carik@gmail.com>

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -39,7 +39,7 @@ int main(int, char*[])
 
 
         // Create basic file logger (not rotated)
-        auto my_logger = spd::basic_logger_mt("basic_logger", "logs/basic.txt");
+        auto my_logger = spd::basic_logger_mt("basic_logger", "logs/basic");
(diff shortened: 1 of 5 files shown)


gabime:
Thanks! Looks good.

PR #379, opened 2017-03-07
add wide string to utf8 string support

sidyhe:
for log different languages at the same time, mbcs can not do that.


sidyhe:
sorry, words below is english.
for some Win32 Applications, use unicode to support multi-byte character set(in the same time), but spdlog not support that write a wchar_t* string to log, i post this patch to solve it, hope you merge my code, thanks!

sidyhe:
sorry, I only test it on win10 x64.

to use it for console:
on windows, console page code will change to 65001, but still something unexpected (it's not matter)
on linux, it should be ok, I think (utf8 is default), not tested.

this PR is to slove my problem just, eg: write chinese and korean to a same file.
PR not use any os api, all code use standard library, so it may be help others who need it.

gabime:
ok, thanks for this.
I will consider to merge it.  Not sure.

DBJDBJ:
Allow me to chime in :) 

http://www.perlmonks.org/?node=871945
http://illegalargumentexception.blogspot.com/2009/04/i18n-unicode-at-windows-command-prompt.html#charsets_unicodeconsole

Perhaps some of our far-east friends can enlighten us, but I would just drop it. Thanks ...

sidyhe:
Happy you chime in, @DBJDBJ 
Also happy you meet your case like this PR,  -_-!
My target is let different "multi-byte character set" get together, same file, same console, and so on...
Unicode is good, but UTF-8 is better(compatible with ASCII), I think
```spdlog``` support utf8 output already, just add a way to accept unicode input.
convert every uncode string to utf8 string and log, is terrible

DBJDBJ:
I think [this ](http://dbj.org/c-windows-unicode-console-output/)is not a "self plug" just a way to focus the debate.

kleuter:
1. Thanks.
2. You shouldn't change the _README.md_ file and the _details/os.h_ file

sidyhe:
@kleuter 
Hi, I was rebuild my fork, so this PR cannot change any more  (or I don't know how to do it).
If you interested it, please look my new fork, branch "dev".

kleuter:
@sidyhe
create then a new pull request, please

sidyhe:
@kleuter #442

Issue #380, opened 2017-03-07
Build error when using compiler wrapper nvcc_wrapper


pkestene:
When building example.cpp with nvcc_wrapper, which wraps nvcc (Nvidia compiler), one gets the following error:

nvcc_wrapper    -I/home/pkestene/install/spdlog/github/spdlog/include  -Wall    --std=c++11 -o CMakeFiles/example.dir/example.cpp.o -c /home/pkestene/install/spdlog/github/spdlog/example/example.cpp

/home/pkestene/install/spdlog/github/spdlog/include/spdlog/fmt/bundled/ostream.h: In function ‘void fmt::format_arg(fmt::BasicFormatter<Char, ArgFormatter>&, const Char*&, const T&)’:
/home/pkestene/install/spdlog/github/spdlog/include/spdlog/fmt/bundled/ostream.h:97:50: **error: ‘ArgFormatter’ is not a template**
     typedef internal::MakeArg< BasicFormatter<Char> > MakeArg;
                                                  ^
Please notice that when using regular nvcc, the code build fine, but when building with nvcc_wrapper, nvcc_wrapper replaces option "-c"  by "-x cu" and goes on as if source code might contain actual cuda code. This wrapper is really useful for using performance portability tool kokkos.

There is a simple fix which consist in replacing ArgFormatter by ArgFormatter_ in template argument.

PR #381, opened 2017-03-07
fix build with nvcc compiler

pkestene:
This pull request adresses issue #380


gabime:
Thanks for the fix.  Please post it to [fmt](https://github.com/fmtlib/fmt) repo instead (spdlog only bundles it).

PR #403, opened 2017-03-27
Disambiguate fmt logging methods that are using variadic templates.

Falconne:
Fix for #256 

As variadic template arguments can be zero length, we need to specify that at least one fmt argument is provided, to distinguish these methods from the existing trivial method that takes no fmt arguments.

Without this, static analysers such as ReSharper flag the logging calls as errors.

### DIFF
--- a/include/spdlog/details/logger_impl.h
+++ b/include/spdlog/details/logger_impl.h
@@ -121,43 +121,41 @@ inline void spdlog::logger::log(level::level_enum lvl, const T& msg)
 }
 
 
-template <typename... Args>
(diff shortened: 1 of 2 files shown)


Falconne:
@gabime, good point, I didn't look at the spdlog::log methods as I'm not using them. But yes, it looks like one of the specialised implementations can be removed.

domehead100:
Noting that the ambiguity still exists when wide strings are being passed to the log functions.

Falconne:
Oh yeah. If you can test it then you should probably make a pull request with a copy of specialisations for wchar.

PR #405, opened 2017-03-29
[#404] Reading past valid address with multisink logger

### LINKED ISSUE #404: Reading past valid address with multisink logger
I am using a multisink logger in a Windows app and testing the app with [dr memory](http://www.drmemory.org) tool.  I get several errors related to reading past allowed memory address.

The code is taken from the [multisink example](https://github.com/gabime/spdlog/blob/master/example/multisink.cpp). Here is a minimal example:

```
std::vector<spdlog::sink_ptr> sinks;
sinks.push_back(std::make_shared<spdlog::sinks::wincolor_stdout_sink_mt>());
sinks.push_back(std::make_shared<spdlog::sinks::rotating_file_sink_mt>("c:\\a", "log",  128, 3));
auto logger = std::make_shared<spdlog::logger>("logger", sinks.begin(), sinks.end());  <<< HERE!
```

I think it is something to do with the `sinks.end()` pointer. According to [`vector::end()`](http://www.cplusplus.com/reference/vector/vector/end/) docs: 

> Returns an iterator referring to the past-the-end element in the vector container.The past-the-end element is the theoretical element that would follow the last element in the vector. It does not point to any element, and thus shall not be dereferenced.

This doesn't crash my app, but it looks like something that needs to be looked at as a potential security concern

```
Error #1: UNINITIALIZED READ: reading 0x03824528-0x03824530 8 byte(s)
# 0 _InterlockedExchange64_INLINE                           [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:96]
# 1 std::_Store_seq_cst_8                                   [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:1834]
# 2 std::_Atomic_store_8                                    [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:1852]
# 3 std::atomic_store_explicit                              [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:507]
# 4 std::atomic_store                                       [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:519]
# 5 std::_Atomic_llong::operator=                           [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:643]
# 6 std::atomic<>::operator=                                [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:197]
# 7 spdlog::logger::logger<>                                [c:\git\vcpkg\installed\x86-windows\include\spdlog\details\logger_impl.h:25]
# 8 std::_Ref_count_obj<>::_Ref_count_obj<><>               [c:\program files (x86)\microsoft visual studio 14.0\vc\include\memory:901]
# 9 std::make_shared<>                                      [c:\program files (x86)\microsoft visual studio 14.0\vc\include\memory:971]
#10 Log::ConfigureLogWithSeverity                           [c:\***\log.h:53] << My logging init code

Note: @0:00:04.391 in thread 23268
Note: instruction: lock cmpxchg8b (%esi) %eax %edx %ecx %ebx -> (%esi) %eax %edx

Error #2: UNINITIALIZED READ: reading register eax
# 0 _InterlockedExchange64_INLINE                           [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:96]
# 1 std::_Store_seq_cst_8                                   [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:1834]
# 2 std::_Atomic_store_8                                    [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:1852]
# 3 std::atomic_store_explicit                              [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:507]
# 4 std::atomic_store                                       [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:519]
# 5 std::_Atomic_llong::operator=                           [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:643]
# 6 std::atomic<>::operator=                                [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:197]
# 7 spdlog::logger::logger<>                                [c:\git\vcpkg\installed\x86-windows\include\spdlog\details\logger_impl.h:25]
# 8 std::_Ref_count_obj<>::_Ref_count_obj<><>               [c:\program files (x86)\microsoft visual studio 14.0\vc\include\memory:901]
# 9 std::make_shared<>                                      [c:\program files (x86)\microsoft visual studio 14.0\vc\include\memory:971]
#10 Log::ConfigureLogWithSeverity                           [c:\***\log.h:53] << My logging init code

Note: @0:00:04.393 in thread 23268
Note: instruction: cmp    %eax 0xfffffff0(%ebp)

Error #3: UNINITIALIZED READ: reading register ecx
# 0 _InterlockedExchange64_INLINE                           [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:96]
# 1 std::_Store_seq_cst_8                                   [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:1834]
# 2 std::_Atomic_store_8                                    [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xatomic.h:1852]
# 3 std::atomic_store_explicit                              [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:507]
# 4 std::atomic_store                                       [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:519]
# 5 std::_Atomic_llong::operator=                           [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:643]
# 6 std::atomic<>::operator=                                [c:\program files (x86)\microsoft visual studio 14.0\vc\include\xxatomic:197]
# 7 spdlog::logger::logger<>                                [c:\git\vcpkg\installed\x86-windows\include\spdlog\details\logger_impl.h:25]
# 8 std::_Ref_count_obj<>::_Ref_count_obj<><>               [c:\program files (x86)\microsoft visual studio 14.0\vc\include\memory:901]
# 9 std::make_shared<>                                      [c:\program files (x86)\microsoft visual studio 14.0\vc\include\memory:971]
#10 Log::ConfigureLogWithSeverity                           [c:\***\log.h:53] << My logging init code
Note: @0:00:04.394 in thread 23268
Note: instruction: cmp    %ecx 0xfffffff4(%ebp)
```

o-mdr:
- Initialising atomic value

See examples at http://stackoverflow.com/q/36320008/706456
This issue was discovered with dr memory tool on Windows platform, Visual Studio 2015 C++ 11

Fixes #404

### DIFF
--- a/include/spdlog/details/logger_impl.h
+++ b/include/spdlog/details/logger_impl.h
@@ -18,11 +18,11 @@ template<class It>
 inline spdlog::logger::logger(const std::string& logger_name, const It& begin, const It& end):
     _name(logger_name),
     _sinks(begin, end),
-    _formatter(std::make_shared<pattern_formatter>("%+"))
+    _formatter(std::make_shared<pattern_formatter>("%+")),
+	_level(level::info),
+	_flush_level(level::off),
+	_last_err_time(0)
 {
-    _level = level::info;
-    _flush_level = level::off;
-    _last_err_time = 0;
     _err_handler = [this](const std::string &msg)
     {
         this->_default_err_handler(msg);
--- a/include/spdlog/details/mpmc_bounded_q.h
+++ b/include/spdlog/details/mpmc_bounded_q.h
@@ -159,9 +159,9 @@ private:
     cell_t* const           buffer_;
     size_t const            buffer_mask_;
     cacheline_pad_t         pad1_;
-    std::atomic<size_t>     enqueue_pos_ {0};
+    std::atomic<size_t>     enqueue_pos_;
     cacheline_pad_t         pad2_;
-    std::atomic<size_t>     dequeue_pos_ {0};
+    std::atomic<size_t>     dequeue_pos_;
     cacheline_pad_t         pad3_;
 
     mpmc_bounded_queue(mpmc_bounded_queue const&) = delete;
--- a/include/spdlog/details/mpmc_bounded_q.h
+++ b/include/spdlog/details/mpmc_bounded_q.h
@@ -159,9 +159,9 @@ private:
     cell_t* const           buffer_;
     size_t const            buffer_mask_;
     cacheline_pad_t         pad1_;
-    std::atomic<size_t>     enqueue_pos_;
+    std::atomic<size_t>     enqueue_pos_ {0};
     cacheline_pad_t         pad2_;
-    std::atomic<size_t>     dequeue_pos_;
+    std::atomic<size_t>     dequeue_pos_ {0};
     cacheline_pad_t         pad3_;
 
     mpmc_bounded_queue(mpmc_bounded_queue const&) = delete;
--- a/include/spdlog/logger.h
+++ b/include/spdlog/logger.h
@@ -87,7 +87,7 @@ protected:
     spdlog::level_t _level;
     spdlog::level_t _flush_level;
     log_err_handler _err_handler;
-    std::atomic<time_t> _last_err_time {0}; // init atomic value to avoid reads of uninitialised memory
+    std::atomic<time_t> _last_err_time;
 };
 }
 
--- a/include/spdlog/logger.h
+++ b/include/spdlog/logger.h
@@ -87,7 +87,7 @@ protected:
     spdlog::level_t _level;
     spdlog::level_t _flush_level;
     log_err_handler _err_handler;
-    std::atomic<time_t> _last_err_time;
+    std::atomic<time_t> _last_err_time {0}; // init atomic value to avoid reads of uninitialised memory
 };
 }
 


o-mdr:
> On closer inspection, I see that the ctor (line 25) do initialize this. So not sure now why the warning appears

Yes, it does. It looks like this boils down to

```
std::atomic<time_t> _last_err_time;
_last_err_time = 0;
```

vs

```
std::atomic<time_t> _last_err_time {0};
_last_err_time = 0;
```

In the former case, (for whatever reason) the memory is read and then assigned the value 0. In the later case, the memory is set to 0 first, then (i think read) and assigned 0 again. I am not sure if it is an actual problem, but with the default assignment (`std::atomic<time_t> _last_err_time {0};`) the dr memory error disappears. 

I am not a big security expert, but if the memory contains some specially crafted value one may perform attacks on the application. 

I will test this with Visual studio 2013 and scan the codebase for any other atomic use, in case there are other places like this.

o-mdr:
I've tested this with VS 2013 Express (platform v120 and v120_xp) - the source code compiles and works as expected.

Also updated another file where a similar construct is used.

o-mdr:
Can we keep both? That is

```
// create logger with given name, sinks and the default pattern formatter
// all other ctors will call this one
template<class It>
inline spdlog::logger::logger(const std::string& logger_name, const It& begin, const It& end):
    _name(logger_name),
    _sinks(begin, end),
    _formatter(std::make_shared<pattern_formatter>("%+")),
    _last_err_time(0),
    _level(level::info),
    _flush_level(level::off)

{
    _err_handler = [this](const std::string &msg)
    {
        this->_default_err_handler(msg);
    };
}

...
std::atomic<time_t> _last_err_time {0};
```

gabime:
why?

o-mdr:
There are several reasons:

- if anyone else runs dynamic memory analysis tool on their apps, spdlogger may come up in their analysis
- it fixes a potential security vuln 
- Improves stability on different platforms, as uninitialised read is undefined behavior 
- without this, a dynamic memory analysis tool produces 3 errors per single call to a constructor 
```
Error #1: UNINITIALIZED READ: reading 0x03824528-0x03824530 8 byte(s)
Error #2: UNINITIALIZED READ: reading register eax
Error #3: UNINITIALIZED READ: reading register ecx
```

It is unusual, that during an assignment in a constructor, there is a prior read operation. This only appears with `std::atomic`. I also understand your point about code style. However, I think the benefits of this should now be clear. I have included an example and it's trivial to replicate this behavior on another Windows machine running free memory analysis tool.

gabime:
Obviously (I think) that there is no way to read uninitialised member if it was initalized using the ctor inialization list. 

If the dr tool still complains,seems like a problem with the tool rather than spdlig. For example I think valgrind never complained about it.

o-mdr:
I've tried to reproduce it with debug compilation of 

```
#include "stdafx.h"
#include <atomic>
int main()
{
	std::atomic<time_t> foo;
	foo = 0;
	return foo;
}
```

Dr memory thinks it's OK. Indeed, looks like a problem with the tool rather than code.

gabime:
Still it would be good if you fixed the code to use ctor initialization. Could you please?

o-mdr:
Oh... I missed that. I assumed it was doing this already. Right... 

I will update 2 files and revert the "field" member initialisation (or whatever it's called)

PR #407, opened 2017-03-29
Update example in README.md

eliaskosunen:
Currently, the example in README.md uses both `spdlog` and `spd` namespaces. In the code, there are three occurrences of `spdlog` compared to 16 of `spd`. This PR changes all of the `spdlog`s to `spd`s.


gabime:
Thanks!

PR #418, opened 2017-04-15
Adding debian support to README file.

kssim:
I've specified a supported debian version in the README file.
I'll ask for confirmation.
Thank you in advance.


gabime:
Why update the readme before the debian is comfirmed ? should be the other way around

kssim:
I've confirmed that it works fine on debian's stratch.

As a result of my checking, spdlog can not support the jessie version in principle,
so I indicated that the README also only supports stretch.

If the stable version is not supported and you do not think it needs to be marked, I'll delete it.

PR #425, opened 2017-04-28
Add an optional final qualifier to types

jcelerier:
When building with GCC's [-Wsuggest-final-types](http://hubicka.blogspot.fr/2014/01/devirtualization-in-c-part-1.html), a lot of types of spdlog
are marked as being more optimizable if they were marked final.

This patch adds a possibility for the user of the library to `#define SPDLOG_FINAL final`
and enjoy potentially better performance : GCC is then able to replace virtual calls by true
function calls if it can ensure that there are no derived types).

By default SPDLOG_FINAL is defined to nothing to not break existing code that
may be inheriting of some of these types for some reason.


jcelerier:
I don't know about setting it by default, I guess some people have been inheriting from stdout_sink maybe ?

gabime:
Maybe , but the vast majority of users might benefit for this. The few users that need virtual can comment out this macro easily.

Issue #424, opened 2017-04-26
andoid_sink retry on -EAGAIN error


alzix:
In cases when log throughput is too high, call to `__android_log_write` will result with `-EAGAIN` return code.
i think it may be worth adding sleep and retry mechanism instead of throwing an `spdlog_ex` exception.

This will throttle message throughput and will cause queue overflow in case asynchronous logger, which is handled according to the policy.

refer to http://androidxref.com/7.1.1_r6/xref/system/core/logcat/tests/logcat_test.cpp

gabime:
Agreed.  could you issue a pr? a single retry after a wait of few millis should be enough i think

alzix:
why do you thing it should be a single retry? I think in cases when logcat daemon is busy spdlog should perform busy wait instead of dropping the message. In case you drop a message you provide no way for application to throttle itself. In opposite case application queue will overflow and application will decide whether to drop in wait according to the async policy.

I think we can reuse `spdlog::details::async_log_helper::sleep_or_yield` for such purpose.

gabime:
Just to avoid waiting too much.. 
I am not very familiar with android logging mechanism - If you think it should retry multiple times (infinite?) then a simple retry loop with small sleep should do. I think the sleep_or_yield is too complicated for this.

alzix:
merged #427

PR #427, opened 2017-04-30
android sink - add retry mechanism

alzix:
in some cases subsequent calls to __android_log_write() may result with -EAGAIN error code.
in such cases spdlog will sleep and try again for number of times defined in tweakme.h

default behavior not changed

issue: #424


alzix:
done

Issue #431, opened 2017-05-04
log message ID formatter


alzix:
When using `async_overflow_policy::discard_log_msg`  some messages could be lost once queue overflows. 
I suggest adding per logger message counter and adding additional formatting pattern for printing message IDs
Message ID will help to analyze logs in case some of them get lost.

gabime:
This will come with a cost of atomic increment for each log call. Not sure it worth it or for most use cases. 

Maybe the logger could count the overflown messages and emits a special message after the overflow is over. 
Something like "Discarded 544 messages due to overflow..."

alzix:
sounds as a great idea but there are two drawbacks:
1. in order to count overflow messages logger would use same atomic increment, although only in overflown state.
2. "Discarded xxx" message can only be printed from within critical section by adding lock on the printing thread
Is it acceptable or am i missing something?

gabime:
You are missing something :) All those actions would be done on the worker thread (each async logger has one) , which means no locking needed whatsoever.

gabime:
Maybe... In any case it should be implemented without degregading performance. Otherwise it is not worth it.

alzix:
i think the solution we have chosen is sub-optimal.
in case logger is initialized with big queue size lets say 4K. In case queue is overflown the warning message will be displaced in the log by 4K messages - since the counter raises for queue tail but the warning is printed when popping from the queue head.
i suggest reverting my latest pull-request an fall-back to the original idea of message count.
i have performed benchmark for atomic increments - as using https://github.com/google/benchmark. My benchmark code can be found below.
The results are:
```
Run on (4 X 2594 MHz CPU s)
05/13/17 23:11:07
------------------------------------------------------------------------------------------------
Benchmark                                                         Time           CPU Iterations
------------------------------------------------------------------------------------------------
regular_increment/real_time/threads:4                             3 ns         10 ns  236224992
atomic_increment/real_time/threads:4                             24 ns         90 ns   36615224
atomic_fetch_add/real_time/threads:4                             23 ns         78 ns   30331140
atomic_fetch_add_emory_order_relaxed/real_time/threads:4         20 ns         61 ns   33459324
```
as you can see add atomic message counter will not produce significant performance hit - but will provide sense of continuous log in case it is really important.

benchmark code:
```c
#include "benchmark/benchmark_api.h"

#include <stdint.h>
#include <atomic>

uint64_t g_regular_counter(0);
std::atomic<uint64_t> g_atomic_counter(0);

[...]

PR #441, opened 2017-05-11
add an option to warn about discarded messages

alzix:
when using async_logger with async_overflow_policy::discard_log_msg each discarded
message will be counted and warning will be printed by the worker thread

this new feature is disabled by default
related issue - #431


alzix:
- fixed 1 and 3 items
- i'll investigate about 2. increment can be performed with `memory_order_relaxed` with no worries but i'm not sure about `exchange(0)`
- i've added a test - it is unstable. seems it uncovers `mpmc_bounded_queue` issue - occasionally it returns failure when populating last element.

PR #444, opened 2017-05-15
Minor fixes

alzix:


gabime:
The copy assignment is deleted, not the move assignment (this wont compile)

alzix:
you are correct - reverted

PR #445, opened 2017-05-16
implement message counter feature

alzix:
messages can be counted by adding %i to logger pattern string


alzix:
#445

alzix:
i'm not sure how to update the wiki page...

include/spdlog/formatter.h
```diff
@@ -33,11 +34,13 @@ class pattern_formatter SPDLOG_FINAL : public formatter
     pattern_formatter(const pattern_formatter&) = delete;
     pattern_formatter& operator=(const pattern_formatter&) = delete;
     void format(details::log_msg& msg) override;
+    bool do_msg_count() override;
 private:
     const std::string _pattern;
     std::vector<std::unique_ptr<details::flag_formatter>> _formatters;
     void handle_flag(char flag);
     void compile_pattern(const std::string& pattern);
+    bool _is_msg_count_requested = false;
```
gabime: Please init it in the ctor to be compatible with rest of code

include/spdlog/details/pattern_formatter_impl.h
```diff
@@ -668,3 +680,7 @@ inline void spdlog::pattern_formatter::format(details::log_msg& msg)
     //write eol
     msg.formatted.write(details::os::eol, details::os::eol_size);
 }
+
+inline bool spdlog::pattern_formatter::do_msg_count() {
```
gabime: I think```is_msg_count_needed()``` is better name

include/spdlog/details/logger_impl.h
```diff
@@ -87,6 +88,9 @@ inline void spdlog::logger::log(level::level_enum lvl, const char* msg)
     {
         details::log_msg log_msg(&_name, lvl);
         log_msg.raw << msg;
+        if (_formatter->do_msg_count()) {
```
gabime: This is nice.. (though I wonder if the cost of calling ```do_msg_count()``` is similar or to the atomic increment  though..)

alzix: i do agree. but you had concerns about performance hit. as i have shown in my benchmark the cost of atomic increment can be neglected.
do you prefer to revert conditional increment logic?

alzix: another option is to add yet another ifdef...

gabime: I think we can do both:

1. Just use atomic increment without the conditional logic
2. Disable by default, and have the option to enable it in tweakme.h with
```c++
#define SPDLOG_ENABLE_MESSAGE_COUNTER
```
If in the future there will be more demand for this, I will enable by default
3. Need to document in the wiki near the %i  option that it is enabled only if defined in tweakme

alzix:
modified sources according to your request.
i have created two separate commits. the second one deals with tweakme.h - in case you will decide to enable this feature by default - it will be pretty convenient to revert the second commit only.

it seems i do not have permissions to commit to your wiki page.  please assist

gabime:
Why cant you edit the wiki? It is public and open for everyone.

lukasgeyer:
@gabime @alzix What is the state of this feature? With 268222e the `i_formatter` has been removed from `include/spdlog/details/pattern_formatter_impl.h` so `spdlog` cannot be compiled if `SPDLOG_ENABLE_MESSAGE_COUNTER` is defined:
```
In file included from ../spdlog/include/spdlog/formatter.h:46:0,
                 from ../spdlog/include/spdlog/sinks/base_sink.h:14,
                 from ../spdlog/include/spdlog/logger.h:15,
                 from ../spdlog/include/spdlog/spdlog.h:14,
                 from main.cpp:3:
../spdlog/include/spdlog/details/pattern_formatter_impl.h: In member function 'void spdlog::pattern_formatter::handle_flag(char)':
../spdlog/include/spdlog/details/pattern_formatter_impl.h:632:71: error: expected primary-expression before '(' token
         _formatters.push_back(std::unique_ptr<details::flag_formatter>(new details::i_formatter()));
                                                                       ^
../spdlog/include/spdlog/details/pattern_formatter_impl.h:632:76: error: expected type-specifier
         _formatters.push_back(std::unique_ptr<details::flag_formatter>(new details::i_formatter()));
```

gabime:
@lukasgeyer It was removed by mistake. Should be merged back in

Issue #215, opened 2016-05-20
Logging UTC time, not local time


eao197:
I have found myself in a situation where I need to use UTC-time in log files. But it seems that standard pattern formatter in spdlog always use details::os::local_time for conversion from std::time_t to std::tm.

So now I have a choice:
- write my own formatter, but it will be half backed and ugly copy of mature implementation of spdlog::pattern_formatter class;
- copy and paste the whole implementation of spdlog::pattern_formatter and change just one line.

I don't like both of them. I think it is possible to slightly improve spdlog::pattern_formatter implementation:
1. Crate a basic implementation for pattern_formatters:

``` cpp
class pattern_formatter_base : public formatter
{
public:
    explicit pattern_formatter_base(const std::string& pattern);
    pattern_formatter_base(const pattern_formatter_base&) = delete;
    pattern_formatter_base& operator=(const pattern_formatter_base&) = delete;
    void format(details::log_msg& msg) =0;
protected:
    // This will have almost all of the current pattern_formatter::format implementation.
    void format(details::log_msg& msg, const std::tm & tm_time);
private:
    const std::string _pattern;
    std::vector<std::unique_ptr<details::flag_formatter>> _formatters;
    void handle_flag(char flag);
    void compile_pattern(const std::string& pattern);
};

template<typename tm_provider>
class pattern_formatter_template : public pattern_formatter_base
{
public :
  ...
  void format(details::log_msg& msg) override {
    format(msg, tm_provider::get_time(log_clock::to_time_t(msg.time));
  }
};

using pattern_formatter = pattern_formatter_template<local_time_provider>;
```

Where local_time_provider will be:

``` cpp
struct local_time_provider {
  static std::tm get_time(std::time_t t) { return details::os::local_time(t); }
};
```

And there also will be gm_time_provider:

``` cpp
struct gm_time_provider {
  static std::tm get_time(std::time_t t) { return details::os::gmtime(t); }
};
```

It makes possible to do something like:

``` cpp
using gmtime_pattern_formatter = spdlog::pattern_formatter_template<
  spdlog::gm_time_provider >;
spdlog::set_formatter(std::make_shared<gmtime_pattern_formatter>());
```

It you decide that this is an appropriate idea I could implement it in my spdlog fork and make pull request.

gabime:
I could change the private members of ```pattern_formatter``` class to be protected and then you could just inherit it and override the ```format``` method (short method, so copy/paste is not so bad):

```c++
[...]

PR #451, opened 2017-05-30
UTC Time

ThePhD:
This issue addresses #215 in the way @gabime stated he'd like it to be handled: with a simple flag that can be passed to `pattern_formatter`'s constructor. It's a smaller, simpler solution that propagates its changes to the virtual functions `_set_pattern` and `set_pattern` in `spdlog::logger` and `spdlog::async_logger`.

The implementation is made as an enumeration simply because there may be more efficient time abstractions that spdlog takes advantage of in the future: the enumeration allows for seamless transition to these while maintaining backwards compatibility by ensuring that all defaulted loggers and pattern formatters use `spdlog::pattern_time::local` by default.

Anyone who has derived from `logger` and `async_logger` will have to change their code to handle the new parameter for the virtual `set_pattern` call, but I believe this is a negligible burden in 100% of the cases where people have actually customized `spdlog` to this degree.

I have not measured the impact of a switch statement, but have done my best to use my general knowledge to ensure it will remain more or less swift, especially in the default case.

Small question: did you mean to never initialize the member variable `_pattern` in `spdlog::pattern_formatter` ? It never gets initialized (and you never use it anywhere?). I would imagine it's accidentally an artifact of slowly changing the pattern formatter to work in a compiled form and actually no longer needing it, so I don't set it in my code, but I do set the `_time` member variable. Shouldn't it be removed, if that's the case? If so, I will be happy to remove it in this branch.

Let me know if you need anything else.

### DIFF
--- a/include/spdlog/async_logger.h
+++ b/include/spdlog/async_logger.h
@@ -71,7 +71,7 @@ public:
 protected:
     void _sink_it(details::log_msg& msg) override;
     void _set_formatter(spdlog::formatter_ptr msg_formatter) override;
-    void _set_pattern(const std::string& pattern, pattern_time ptime) override;
+    void _set_pattern(const std::string& pattern, pattern_time_type pattern_time) override;
 
 private:
     std::unique_ptr<details::async_log_helper> _async_log_helper;
--- a/include/spdlog/async_logger.h
+++ b/include/spdlog/async_logger.h
@@ -71,7 +71,7 @@ public:
 protected:
     void _sink_it(details::log_msg& msg) override;
     void _set_formatter(spdlog::formatter_ptr msg_formatter) override;
-    void _set_pattern(const std::string& pattern) override;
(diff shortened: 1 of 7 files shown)


ThePhD:
1. (Fixed) Oops, that was a mistake from something else.

2. (Fixed) I'll change it to the if statement. I use a lambda so I could early-return and I only used it once, but I can just make it a private member function instead.
2a. (Rationale Explanation, Not Actually Done) If I remove the "default" statement and use a switch statement, then gcc/clang will warn if the switch statement doesn't use all the enum values of `pattern_time_type`. This might be useful for catching unupdated code in the future if more `pattern_time_type` values are added. But, I'm still leaving it as the if-statement as you requested!

3 & 4. (Fixed) I was wondering what would be a good way to name them. Changing to pattern_time_type frees up the variable name pattern_time: done!

Issue #458, opened 2017-06-12
Creating async and sync loggers


cneumann:
Creating a mixture of async and sync loggers in a single program seems to require careful coordination/thread synchronization since calling spdlog::set_async_mode toggles essentially global state. So even if I write:
```c++
spdlog::set_async_mode(4096);
std::shared_ptr<spdlog::logger> l = spdlog::stdout_logger_mt("logger");
// assume scheduler switches to other thread creating a non-async thread here
spdlog::set_sync_mode();
```
There is a short window for a race condition where on a different thread I end up creating an async logger even though the intent was to create a regular one.
Is there a way to create async loggers without touching "global" state? Perhaps add `spdlog::create_async` functions with the extra arguments (queue size, callbacks) that are needed for the async loggers?

gabime:
Sure. You can create them directly 
```c++
auto l = std::make_shared<spdlog::async_logger>(name, qsize, ..)
```
I agree that this global state might not the best idea. Maybe it should be deprecated in favour of spdlog::create_async or similar.

cneumann:
Superseded by PR #459. Thanks.

PR #459, opened 2017-06-15
Add create_async factory functions for async loggers

cneumann:
As mentioned in #458 this adds `create_async` functions to allow convenient creation/registration of async loggers without flipping the global async mode switch.


gabime:
Thanks!

Issue #324, opened 2016-11-28
Several Simple Improvements to Tweaks


batduder:
I am a big fan of spdlog.  A few simple improvements would make it even better:

- Add a tweak for log level case -> log level tags in the output are in lower case.  I would prefer upper case.
- Add a tweak for log level tag length match -> I would prefer all log level strings to be same length so that log text lines up better in the output
- Add a tweak for setting file extension -> for rotating files the hard coded extension is ".txt".  I would prefer ".log"
- Would be nice if level filter was set globally in a global XML file or config file rather than from code.  This is needed for systems where many processes/apps use spdlog and we want the filter to be controlled from one place for all apps
- Add an "always" level at the highest level.  Sometimes in production code when filter is set to "error" there are still information items we would want to see in the log.  An "always" level would permit this.

- Any plans to add a socket appender?

Razzeee:
Cool would also be a fixed lenth for thread id

PR #471, opened 2017-07-05
fixed length and upper case log level tags

asit-dhal:
Hi,

I added the first two enhancement request https://github.com/gabime/spdlog/issues/324  in this PR.

Please check this and let me know if you want something different.

Warm Regards,
Asit


gabime:
Thanks. I wonder though if it would be better just to provide the option define custom  level names in tweakme.h instead. 

So the user could  define in tweakme.h for example:
```c++
#define SPDLOG_LEVEL_NAMES  { "   TRACE", "   DEBUG", "    INFO",  " WARNING", "   ERROR", "CRITICAL", "OFF" };
```
And to support this common.h would contain:
```c++

#if !defined(SPDLOG_LEVEL_NAMES)
#define SPDLOG_LEVEL_NAMES { "trace", "debug", "info",  "warning", "error", "critical", "off" };
#endif
static const char* level_names[] SPDLOG_LEVEL_NAMES 
```

asit-dhal:
Hi,

It is better than my implementation. I will send a new pull request and with some additional features.

Thanks.

Razzeee:
If you can send the lenth fix in it's own PR, that would be nice. Should hopefully lead to a faster acceptance if there are no other changes involved.

asit-dhal:
Hi,

I will do it today and send you a PR.

PR #475, opened 2017-07-13
support variable format padding

Montellese:
This adds configurable pattern flag padding similar to `printf()` to some of the pattern flags. The following features are supported:
* use `-` for left-alignment. default is right-alignment (as with `printf()`)
* use any number to specify the width of the padding
* use a leading `0` to fill the padding with `0`s instead of whitespaces.

The following pattern flags support padding:
* %n: logger name
* %l: level
* %L: short level
* %t: thread id
* %a: abbreviated weekday name
* %A: full weekday name
* %b / %h: abbreviated month name
* %B: full month name
* %p: process id

All other pattern flags (expect for the log message) should already be fixed width and therefore padding doesn't make sense or can be achieved with whitespaces between the pattern flags in the custom format.

The feature is disabled by default and can be enabled by defining `SPDLOG_ENABLE_PATTERN_PADDING`.

Examples:
* `%l %v`
```
info Hello
warning Hello
```
* `%-7l %v`
```
info    Hello
warning Hello
```

To make the code a bit easier to read we could remove the `#ifdef SPDLOG_ENABLE_PATTERN_PADDING` in `compile_pattern()` and `handle_flag()` because that should really have any effect on performance and just keep the `#ifdef`s in the actual format implementations.

**Beware:** As the feature is disabled by default it also isn't really covered by the catch tests. I've added a few simple test cases but we would have to enable `SPDLOG_ENABLE_PATTERN_PADDING` by default for the test project and I wasn't sure if that is acceptable.


gabime:
Thanks. I wonder though if so many changes worth it. I think the main requirement is fixed width of fields, and this could be accomplished in simpler ways..

Montellese:
Fixed width works fine for fixed values like log levels or day /month names but it does not work well for thread and process id which might have different value ranges for different platforms or for logger names which are completely arbitrary.
If you want to support different fixed widths for different fields (e.g. 8 for log levels and 4 for short month names) you end up with one fixed width option per field which probably doesn't result in less code.

But obviously it's your call. Missing support for fixed width log levels (which is possible now with the latest PR that has been merged), thread ids and logger names is currently blocking me from using spdlog in Kodi.

Montellese:
Btw I just tried to integrate spdlog with this PR applied to Kodi but enabling `SPDLOG_LEVEL_NAMES` and `SPDLOG_ENABLE_PATTERN_PADDING` in `tweakme.h` didn't result in the expected behaviour. I had to include `tweakme.h` in `pattern_formatter_impl.h` and in `common.h` to get the expected behaviour. Is that intended or am I doing something wrong?

gabime:
> Is that intended..

No, seems like a bug

Montellese:
I think I figured out the issue with `tweakme.h`. I included the sink header files before `spdlog.h` which resulted in `tweakme.h` being included after the sink header files.

Montellese:
@gabime what's your prefered way of moving forward with this? Do you want me to put together a pull request with configurable fixed width for log level and thread id only?

gabime:
@Montellese I am not sure yet. Can you elaborate on your proposed PR ?

Montellese:
Well IMO there are two options.

The first option is this PR which allows pretty flexible (left and/or right) padding with whitespaces or zeros.

The second option would be adding separate defines through `tweakme.h` for every format pattern that should be paddable like e.g.
```
#define SPDLOG_PATTERN_PADDING_THREADID_LENGTH 10
#define SPDLOG_PATTERN_PADDING_LOGGERNAME_LENGTH 8
```
which will result in less intrusive code but requires additional defines for every paddable format pattern and is not as flexible concerning left/right padding and the padding charachter.

For both approaches I assume that padding support must be optional and disabled by default to keep the performance benefit of spdlog.

gabime:
I think single padding define in tweakme.h (which would be applied to all fields) is good compromise

Montellese:
Sorry for the late reply but I was on vacation.

Unfortunately single padding applied to all fields won't work for us because the fields have completely different maximum lengths. For the level field it's 7 but for the logger name it's at least 15. If we would apply 15 to all fields we would waste 8 spaces on every log level :-/

mdealer:
We need this per-field too.

gabime:
@Montellese fwiw spdlog now supports  padding. 
Pad left (“%8..”), pad right (“-%8..”), and center (“%=8..”). 
It pads only with spacesno zeros) but thats the usual usecase for this lib.

Montellese:
@gabime That's great news, thanks for the heads up.

PR #477, opened 2017-07-19
add hunter to package managers in readme

headupinclouds:
spdlog is also available in the [hunter](https://github.com/ruslo/hunter) CMake package manager.

https://github.com/ruslo/hunter/wiki/pkg.spdlog

```
hunter_add_package(spdlog)
find_package(spdlog CONFIG REQUIRED)
target_link_libraries(... spdlog::spdlog)
```

https://github.com/gabime/spdlog#or-use-your-favourite-package-manager

Thanks for spdlog!


gabime:
Hunter is cool, but I  try to list only the major managers. Otherwise the list would be very too long.

Issue #440, opened 2017-05-11
how can I use c style formatting e.g. %s, %d, etc.


tibabit:
Hi,
Is it possible to use c style formatting, e.g. %s, %d etc,

gabime:
No such support

iamantony:
Same issue: #272

Issue #402, opened 2017-03-24
how can i change the format style to c printf style gracefully ?


yandaren:
I' used to the c style format of string, can i  replace the format style from python style to c style gracefully?

Issue #272, opened 2016-09-01
printf-style formatting of  logs message part


matswebjorn:
Is it possible to change message formatter to use one which supports printf-style formatting (i.e %d) ?
And if so, is there some sample code on how to do this?

gabime:
can you give a concrete example ?

matswebjorn:
Well, I was just curious if it was possible to change the default Python/C# style formatter ("My cats name is {}", "Carl") to printf style ("My cats name is %s", "Carl"). I can't find where the hook to fmt is made.

RaptorFactor:
I would like this too. I have a large pre-existing codebase where I want to change the logging library, but all the current log messages assume printf-style formatting. It appears that fmtlib even supports the printf-style syntax, one would only have to go and change where spdlog invokes the formatting APIs (I assume - I have not yet looked).

http://fmtlib.net/latest/api.html#printf-formatting-functions

gabime:
pull request about this is welcome

gabime:
It's not so easy. I don't see any fmt::BasicWriter support for this.

fwsGonzo:
:+1:

rhd:
This feature would make my life easier too! :+1:

haoxi911:
Want to switch from my own simple logger into spdlog, but put it on hold because of this. +1

haoxi911:
But...we can workaround it by compose the full log via printf with our own approach, and then adapt to spdlog by always printing full string.

vwbij:
I was interesting in printf style formatting to. I have played around with a extra version of the fmt lib. I found the following:

fmt has got the header "printf.h" that supports printf style formatting. There are the functions sprintf, fprintf and printf. The difference between this functions and the standard printf functions is that they want to get the variable parameter list as a object of the type "ArgList". Maybe this could be helpfull.

For the DLL I build I found this not very helpfull. I ended up providing two Ways: for the old code I just use oldstyle printf to a buffer before invoking spdlog logging. For the new code I have created a log builder object that takes its parameters over a streaming style interface. 

I found that there is no easy conversation between the oldstyle printf parameter list and the functionality of "printf.h" - at least if you want/need a C style dll export interface. It is not possible to parse a oldstyle VARARG parameter list without the information about the parameter types that are bound to the format string. And this is a very oldstyle unsafe approach, so I think it's the best to use old style printf functions before logging for old code, and the safe fmt way of formating for new code.

jlhorner1974:
[...]

PR #479, opened 2017-07-24
Add another "tweak" to enable `printf` formatting

arossert:
Add new "tweak" to replace the `python` formatting with `printf` formatting.

This addressed this issues:
#402, #440, #272


gabime:
Thanks for the pull. 
I'd rather not add  another way to do formatting and stick to a single way - This will help keep things simple as possible.

PR #484, opened 2017-07-29
Background color unchanged

vgabi94:
Keep the background color unchaged on Windows console when using colored loggers.


gabime:
What about the critical level's red background?

vgabi94:
It remains red. I'm just removing the default background color which appears to be red. Here is an example:
Without my code
before
With my code
after

Issue #155, opened 2015-12-23
rotating notification / user's callback


azadkuh:
it would be great if `spdlog::rotating_file_sink` constructor accepts an optional callback which will be called when `rotating_file_sink::_rotate()` finishes (similar to `spdlog::set_async_mode` warmup_cb)

by this callback, one can do:
- compress the older logs with his/her favorite algorithm
- move, backup, send, ... the rotating files
- notify a node or an admin user 
- process the latest rotate file and make a report
- handle errors by clean up, ...
- ...

at the right time.

bakinat:
If I may add another related enhancement to also be able to access the new created file (maybe in a separate callback) to do such thing as writing a common header line.
Thanks, Barak

asit-dhal:
Hi @gabime ,

Should that callback  be supplied with any arguments ?

gabime:
which arguments for example?

asit-dhal:
file name or simply a notifcation should be made ?

gabime:
Good idea. I think the both new and old filenames shoud be passed (and maybe also the open fd of the new file if the user wants to write stuff to it at this point)

asit-dhal:
ok thanks

I will add those and send you a pull reuqest.
Please ignore the PR https://github.com/gabime/spdlog/pull/487

PR #489, opened 2017-08-04
rotating file logger callback

asit-dhal:
Hi

I have added callback for rotating file logger. https://github.com/gabime/spdlog/issues/155

Feature implemented

- callback for rotational logger: list of files are notified
- callback for daily file logger: recently closed file is notified

I didn't add open file descriptor because it will expose an open file to the client. It may affect future renaming for long running processes.

Please check this one and let me know if you need any changes.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -67,6 +67,9 @@ using level_t = std::atomic<int>;
 
 using log_err_handler = std::function<void(const std::string &err_msg)>;
 
+using rotating_logger_cb = std::function<void(std::vector<std::string>)>;
+using daily_logger_cb = std::function<void(std::string)>;
+
 //Log level enum
 namespace level
 {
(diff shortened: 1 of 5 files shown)


gabime:
Thanks. I will consider merging (I am not sure because it contains many changes, add 2 different arguments to the loggers creation, and maybe users can do without it anyway)

asit-dhal:
I will recheck everything and will simplify it and send you a new pull request(if possible).

asit-dhal:
Hi @gabime , do you have any specific suggestion for improvement ?

gabime:
I think it is not very important. Il

In anyway, the exact use case trigger should be defined:
Is it that after a file is closed the user wants to process it? or maybe its the other way around-notify about new file so the user will be able to write some header in it..

asit-dhal:
Ok Thanks

PR #507, opened 2017-08-16
Add performance logging macros

aschuess:
Add SPDLOG_PERF_x macros which
only perform a simple if check on
logging level before going to log functions.

This provides near zero cost of run-time configurable
changing of log levels.

See unit test and description in spdlog.h for
full description and performance numbers

### DIFF
--- a/include/spdlog/spdlog.h
+++ b/include/spdlog/spdlog.h
@@ -152,6 +152,48 @@ void drop_all();
 
 
 ///////////////////////////////////////////////////////////////////////////////
+// Lower cost run-time logging macros
+// This elminates any function calls or argument passing if the log level
+// is below required level for logging and only incurs an inplace if check
+//
+// Example:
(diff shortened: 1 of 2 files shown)


aschuess:
The only issue is when the log level IS at or above we do the `should_log` check twice which could be eliminated if we added a `__force_log` method skipping this second if check.

gabime:
Thanks for the pull. 
The lib already have SPDLOG_TRACE and SPDLOG_DEBUG macros which does a similar thing (as long as the log level is defined at compile time). 
Maybe your could  just add SDPLOG_INFO, SPDLOG_WARN, etc.

aschuess:
Agreed it's similar to SPDLOG_TRACE and SPDLOG_DEBUG. The advantage is the _run-time_ log level checking giving debug and trace logs in production code with very little cost.

PR #509, opened 2017-08-17
Issue 497. Add multi-process-friendly rotating_file_sink_mp  

iamantony:
This PR contains code that provides multi-process safe way to write rotating file logs.

*Main idea*: each write to the log file must be atomic so we would not see any intersection of messages from different processes. 

To achieve this we should use low-level OS functions:
- Linux:
  - write (http://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html)
  - writev (http://pubs.opengroup.org/onlinepubs/9699919799/functions/writev.html)

- Windows:
  - _write (https://msdn.microsoft.com/en-us/en-en/library/1570wh78(v=vs.100).aspx)

Also log files rotation process have changed. At the start of rotation, rotating_file_sink_mp create special file - "rotating". This file is empty and have one purpose - it says to other processes that rotation is going on and you should wait some time. As soon as rotation would be performed, this flag-file will be removed and all processes would start to write messages in new log file.

*Pros*:
- every write operation is atomic
- messages in the log are sequential
- number of processes that write messages to log file is not limited
- the rotation of the log files takes approximately the same amount of time as the rotation in rotating_file_sink

*Cons*:
- length of the message is limited by OS
- we should constantly request size of the log file from OS - this is *very long* operation
- at rotation all processes should wait till will be ended

*New classes*:
- rotating_file_sink_mp  - based on rotating_file_sink with changes in file handling and rotation process
- file_helper_mp - based on file_helper

Tested on Linux and Windows.
The Windows implementation must be finalized.

### DIFF
--- a/example/bench.cpp
+++ b/example/bench.cpp
@@ -55,6 +55,7 @@ int main(int argc, char* argv[])
 
         auto rotating_st = spdlog::rotating_logger_st("rotating_st", "logs/rotating_st", file_size, rotating_files);
         bench(howmany, rotating_st);
+        bench(howmany, spdlog::create<spdlog::sinks::rotating_file_sink_mp_st>("rotating_mp_st", "logs/rotating_mp_st", file_size, rotating_files));
         auto daily_st = spdlog::daily_logger_st("daily_st", "logs/daily_st");
         bench(howmany, daily_st);
         bench(howmany, spdlog::create<null_sink_st>("null_st"));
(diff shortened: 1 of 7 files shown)


gabime:
Hi, Thanks for the pull.
I don't think it will work though.. Those checks are not atomic. What happens if 2 processes call "safe_rotation()" at the same time for example ? (it can happen - if they both on same time check the existence of the rotating file and don't find it).

To really synchronize between processes, something like [this](https://linux.die.net/man/7/sem_overview) should probably  be used.

Also, using ```write``` is problematic -this is low system call without the stdlib buffering and will slow things  considerably. It also ignores  spdlog's flush semantics (```logger::flush_on(..)```).

An alternative  approach would be to push to a shared-memory queue of messages, and to spawn a single dedicated process to consume the queue and write the messages to the file . This would be much like the async logging but between processes instead of threads, and the only synchronization should happen when pushing to the queue.

iamantony:
Yeah, unfortunately checks are not atomic.

I think the last variant with shared-memory queue will be the most optimal.

PR #515, opened 2017-08-25
Compiler-dependent line numbering

elelel:
Format source filename/line number in SPDLOG_TRACE output according to the compiler's convention it has been compiled under: filename(line_number) for MSVC, filename:line_number for others (GCC-style).


gabime:
Why is it better?

elelel:
This helps to automatically parse the trace output for the tooling that expect standardized output. It makes integration of log output with IDEs to get clickable links a bit easier.

PR #531, opened 2017-10-05
Support for printf-style logging

fogo:
Hello, everyone

I've recently been researching about loggers and I'm impressed with `spdlog` so far. 

My goal right now is to replace an old implementation in one project I'm working with. However this project has lots of code and it is littered with printf-like logging messages, which makes unfanthomable to propose to replace its backed w/ `spdlog` as it is right now.

Fortunately, the underlying `fmt` lib already allows for printf-like messages and I just needed to tweak code a bit so `spdlog` can have this feature too.

I've implemented in a way that the message style is switched by a compile-time definition, as it seems this will always be a static decision by a project. I've updated the tests and also performed some benchmarking, which showed times were unnafected by these changes.

Thanks in advance for the feeback.


gabime:
@fogo Thanks for this. I appreciate the PR and the tests.
This was already requested several times in the past but I always insisted on having a single way to do things.
So, unless there is really big demand for it, I prefer not to add it in the meanwhile.
(as a side note, the ```#include "spdlog/fmt/bundled/printf.h"``` best be put in the [include/fmt/fmt.h](https://github.com/gabime/spdlog/blob/master/include/spdlog/fmt/fmt.h) file).

arossert:
I also submitted a PR for that feature in the past (this one is much better). 
I will be happy to see this feature added, it will make it much easier to implement in legacy code.

fogo:
@gabime thanks for the feedback. I understand your concern. As someone who programmed a lot of Python in the last few years, I can see the advantages of using the format-style and the pythonic "one way to do things".

However in this case it feels like this a barrier for the adoption of your (pretty good) library. Many projects interested in using it (as mine and others like @arossert it seems) have legacy code that won't have the budget to write all log messages spread out throughout thousands of lines code (worse, it even "fails" silently so you can't even have the compiler to help detect conversion points...)

Even though it is a simple thing to keep rebasing and use a fork instead, I hope you reconsider. The way I implemented already is intended so this change is a non-default, compatibility resource for legacy projects, keeping the format-style as preferred without tweaks.

gabime:
@fogo Ok, lets give it a shot. 
Could you please fix the pull regarding the include? Also, it would be nice if you could add a short section about it in the wiki.

fogo:
@gabime thanks! I guess I can edit the wiki and move `#include "spdlog/fmt/bundled/printf.h"` to include/fmt/fmt.h later today.

fogo:
@gabime include fixed. Once merged, I'll update the wiki.

gabime:
Thanks, looks good. Which version of bundeled printf.h file did you put? spdlog uses fmt version 4.0.0.

fogo:
Sorry, I thought I had listed the fmt version on commit message. Yes, I used 4.0.0, more specifically branch `release-4.0`.

fogo:
I'll update the wiki once I get home from work.

PR #533, opened 2017-10-11
rethrow unwind exception

manuel-schiller:
On Linux with pthread library spdlog causes an SIGABORT and crashes
the application in case it catches a thread specific cancellation
exception in a critical execution phase while in a try/catch block
in spdlog/detail/logger_impl.h
    
The exception is caught by some general catch(...) clause where
it is NOT rethrown.

However rethrowing these kind of exception is mandatory, otherwise
an abort will be caused by the glibc.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -42,6 +42,12 @@
 #define SPDLOG_DEPRECATED
 #endif
 
+#ifdef __linux__
+#include <cxxabi.h>
+#define SPDLOG_CATCH_ALL catch (abi::__forced_unwind&) { _err_handler("Unknown exception"); throw; } catch (...)
+#else // __linux__
+#define SPDLOG_CATCH_ALL catch (...)
+#endif // __linux__
 
 #include "spdlog/fmt/fmt.h"
 
--- a/include/spdlog/details/logger_impl.h
+++ b/include/spdlog/details/logger_impl.h
@@ -11,7 +11,6 @@
 #include <memory>
 #include <string>
 
-
 // create logger with given name, sinks and the default pattern formatter
 // all other ctors will call this one
 template<class It>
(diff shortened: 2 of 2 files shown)


gabime:
Intersting. Why __GNUC__? is it gcc specific?

manuel-schiller:
This is not GCC specific, it is specific for usage of the "Native POSIX Thread Library" (NPTL), see also https://udrepper.livejournal.com/21541.html. 

I guess we should rather use `__linux__ ` instead of `__GNUG__`.

gabime:
Will it compile under clang? also, could you please move this “#define” to spdlog/common.h?

manuel-schiller:
moved to common.h

yes this will compile on clang as long as cxxabi.h is present

manuel-schiller:
the error handler will be called for all exceptions except for `abi::__forced_unwind&`. Do you want th error handler to be called in this case as well and then rethrow?

gabime:
Yes please.

manuel-schiller:
done

alurie:
@gabime we encountered the same issue described in this PR (using spdlog v1.6.1, on centos7). 
We do pthread_cancel of a thread that happens to be in the middle of spdlog sink/flush, exception is caught and not re-thrown, causing the crash.
Looking at the spdlog code changes, the fix introduced in this PR was refactored out. 
(e.g. https://github.com/gabime/spdlog/commit/8ca1d84a324f93d67fea535fa13e53065719c5be removed catch(...),
then https://github.com/gabime/spdlog/commit/fd170b0fe13369f7152c8018cbf7df60500dfe53 added it back with re-throw, then https://github.com/gabime/spdlog/commit/78ea362e27109694933c9d2e41ea77d772be3bfc removed the re-throw).
If possible, could you please shed some light on those changes (I could not locate comments associated with those commits). Most interesting question is whether the re-throw was removed because it was causing some issue?
Thank you.

gabime:
I think there was some problem with it, but I do not recall exactly..

gabime:
@alurie This was probably the cause: https://github.com/gabime/spdlog/issues/549

gabime:
@alurie Fixed again

Axas1986:
Why is this issue associated with v1.9.0? Is the release note of v1.9.0 incorrect?

PR #534, opened 2017-10-12
Added MDC logging feature, working only for synchronous loggers

miqelm:
Added possibility to specify in pattern %X{key} (unlimited times).
Then one can use spdlog::mdc:set_context(key, value) to enable replacing above pattern to specified value. Function spdlog::mdc::remove_context(key) removes key from underlying map.

Unfortunately user must include this line in his source file:
`thread_local std::map<std::string, std::string> spdlog::details::MDC_formatter::mdc_map;`

Example:
```
#include <spdlog/spdlog.h>
#include <spdlog/mdc_context_manager.h>

thread_local std::map<std::string, std::string> spdlog::details::MDC_formatter::mdc_map;

int main() {
	auto console = spdlog::stdout_color_mt("console");
	console->set_pattern("%+ MDC_VALUE=%X{key}");
	console->info("Without specified key");

	spdlog::mdc::set_context("key", "value");
	console->info("With specified key");

	spdlog::mdc::remove_context("key");
	console->info("Without specified key");

	return 0;
}
```

Output:

> [2017-10-12 14:12:14.730] [console] [info] Without specified key MDC_VALUE=
> [2017-10-12 14:12:14.731] [console] [info] With specified key MDC_VALUE=value
> [2017-10-12 14:12:14.731] [console] [info] Without specified key MDC_VALUE=

Does not work under MinGW under Windows.

It's not the best solution and I imagine it could not be merged to master right now, but it's proposition.

### DIFF
--- a/include/spdlog/details/pattern_formatter_impl.h
+++ b/include/spdlog/details/pattern_formatter_impl.h
@@ -19,6 +19,8 @@
 #include <utility>
 #include <vector>
 #include <array>
+#include <regex>
+#include <map>
 
 namespace spdlog
 {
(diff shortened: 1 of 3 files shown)


gabime:
Thanks, thats a start.  Implementing this in the formatter would be more natural to spdlog

PR #535, opened 2017-10-17
Update README.md

vitor-alves:
Updated install section for Arch Linux.
I changed
`pacman -S spdlog-git ` to `yaourt -S spdlog-git` in the readme.

[spdlog-git is in AUR](https://aur.archlinux.org/packages/spdlog-git/), not in the official repositories. Therefore it should be installed with an AUR Helper like yaourt or packer, but not with pacman. I suggest yaourt since it seems to be the most commonly used AUR Helper.

Pacman cant find the package with `pacman -S spdlog-git `. I tested in my own Arch Linux machine.


gabime:
Thanks!

PR #545, opened 2017-11-03
Allow compilation on platforms with unwind (android)

costinm:
I'm building envoy, which uses spdlog, using current android NDK and cmake - ran into this error.


costinm:
Can do that - but maybe other platforms have similar limitations. Your call.

gabime:
Lets try using the standard ```__ANDROID__```.
If more platforms will have such problems, I will use your original suggestion instead.

PR #551, opened 2017-11-11
modified path on quoted #includes

daylanKifky:
Paths pointing to the root of the library where replaced for ones relatives to each file.

For example, inside `/include/spdlog/details/file_helper.h:`
`#include "spdlog/details/os.h"`
This will look for `os.h` in `/include/spdlog/details/spdlog/details/` which doesn't exists.
replaced with:
`#include "../details/os.h"`
and similar for all the files


gabime:
Thanks. I like it better that way. It is clean from where the included files come from..

daylanKifky:
:)

Issue #476, opened 2017-07-16
Formatting in trace and debug C macros


elelel:
The trace/debug output format is hard-coded into the macro. It would be nice if that formatting would be customizable, especially when integrating the trace output to IDEs. For example, the trace macro is used to inject "line #" substring into the output, while Emacs under default configuration expects "source.cpp:line" format. To make the trace output clickable, one either has to rewrite Emacs's compile parser, or to undef the SPDLOG_TRACE macro if it is defined and redefine it from scratch. Perhaps it is possible to make this macro dependant on a formatting macro setting.
This is related to #340

werman:
I would also like to see this feature implemented. I have integrated spdlog in a big project and it was utmost improtant to have logs clikable in visual studio output, visual studio has "file.cpp(123) : message" format to navigate to corresponding line. I have solved it by having custom ::log method but it will be nice to have a good solution available in the library.

gabime:
>  I have solved it by having custom ::log method

Could you please share your solution here (at the least the relevant part), might be good start

werman:
Unfortunately it's not a good solution, just was ok for us since we didn't need flexible one. From memory (I'm not near that code atm):
```
inline void spdlog::logger::log(level::level_enum lvl, const fmt::StringRef& msg_prefix, const char *msg)
{
    ...
        details::log_msg log_msg(&_name, lvl);
        log_msg.raw << msg;
        log_msg.formatted << msg_prefix;

        _sink_it(log_msg);
    ...
}
```
From the first look I like #561 idea (didn't look at its optimality) as a general solution.
Also thanks for this good library!

Issue #340, opened 2016-12-29
how to include "__FILE__" and "__FUNC__" and "__LINE__" into log formatter


neutronsharc:
Hi all,  Is there a way to include these macros into log formatter? This would be extremely useful to help locate exact location of some log output.  Thank you!

gabime:
No.  I didn't find an easy way to implemnet this

neutronsharc:
I added a wrapper layer atop spdlog, and use macros to prepend __FILE__ to log entries.  Now problem solved.  spdlog is really awesome. Great work gabime!

Here is my code in case anyone is interested:
https://github.com/neutronsharc/logtest/blob/master/hcdlog.h#L77

gabime:
Thanks. I think I might add a set of macros to the API to provide this feature.

neutronsharc:
This would be better!  Please send me a notice when you are ready.

carun:
I am wondering if this would work after all with LTO turned on. Any thoughts?

gabime:
why not? it just a macro

htfy96:
[source_location](http://en.cppreference.com/w/cpp/experimental/source_location) would be a more elegant solution. 
However it is badly supported in gcc and not even implemented in clang. Hope that several years later this could be adopted as part of standard.

mythos002:
Hi,
I was searching myself for such a solution and after a bit of playing around I came up with something like this (definition is done in spdlog.h just after the original SPDLOG_TRACE define):

`#define SPDLOG_TRACEF(logger, ...) logger->trace("[" + fmt::format(pathlast(__FILE__)) + "::" SPDLOG_STR_HELPER(__LINE__) "::" + string(__func__) + "] " + fmt::format(__VA_ARGS__))
#define SPDLOG_TRACEPF(logger, ...) logger->trace("[" + fmt::format(pathlast(__FILE__)) + "::" SPDLOG_STR_HELPER(__LINE__) "::" + string(__PRETTY_FUNCTION__) + "] " + fmt::format(__VA_ARGS__))`

I had a few problems getting `__func__` and `__VA_ARGS__` to get interpreted together, so I've been testing around and came up with the above.

As you can see I use the fmt::format function to pass several arguments preformated into the logger, so the logger itself should interpret them as normal text like the "[" , "::" and "]" stuff. Also for the `__func__` and `__PRETTY_FUNCTION__` I encapsulated them with the string() function to get a proper string output.

[...]

PR #561, opened 2017-11-21
WIP tracing symbols in pattern

mateuszzz88:
This pull request is proof of concept, not finished work. I want to know if you are interested in changes like this - I may then finish it. It possibly solves #340 and #476 

This commit adds additional symbols to pattern that allow logging:
- file name (as in __LINE__ or just last part of path)
- line number
- function name (as either __FUNCTION__ or __PRETTY_FUNCTION__

To use it, user must 
1. set pattern, eg. "%u @ %j:%#> %v" is equivalent to message "__FUNCTION__ @ __FILE__:__LINE__> %v"
2. for given logger use exclusively macro invocations (SPDLOG_DEBUG)
(see example in tests/trace.cpp)

Implementation:
spdlog::trace_info struct keeps data when given log line was created. It is updated in spdlog::logger by macros. Logger copies this data to log_msg. 
I don't like the way I keep this data in logger, but I don't see way around it without duplicating whole api (suggestions?)

Downsides:
- if user uses non-macro invocation (logger->debug(...)) AND has one of new symbols in pattern, then logged message will be wrong (either empty or pointing to location where last macro was used)
- every log_msg object is a little bigger (sizeof(int) + 3 std::strings), even if pattern doesn't use this data

Further work needed:
- currently this is implemented for SPDLOG_DEBUG and _TRACE. Macros for all log levels are necessary. Is there good reason why those 2 macros are optional and disabled by default in tweakme?
- perhaps it wolud be good to make this whole feature optional in tweakme.

So, would you like me to continue work on this? Any suggestions are welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,5 @@ install_manifest.txt
 /tests/tests.VC.db
 /tests/tests
 /tests/logs/file_helper_test.txt
+*.kdev4
+build
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -80,6 +80,13 @@ typedef enum
     off = 6
 } level_enum;
 
+#define SPDLOG_LOGLEVEL_CRITICAL 5
+#define SPDLOG_LOGLEVEL_ERROR    4
+#define SPDLOG_LOGLEVEL_WARNING  3
+#define SPDLOG_LOGLEVEL_INFO     2
+#define SPDLOG_LOGLEVEL_DEBUG    1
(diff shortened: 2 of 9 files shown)


werl:
I would be interested!

gabime:
thanks for the pull. I need to think if just giving the option to format the trace info worth all those additions. 
As a side note, those strings can be just “const char*” (and save precious cpu time) since they originate nate from the gcc pre processor, and will be valid in the entire process lifetime.

PR #563, opened 2017-11-24
Per sink formatter

### LINKED ISSUE #421: per sink format pattern
I suggest adding an option to set per sink format pattern.
Some sinks (i.e. `android_logger` and `syslog_sink`) apply their own formatting to every print by adding timestamps, severity, thread and process ids.
In case single logger connected to more than one sink, by using `dist_sink_st` it is not sufficient to set formatting pattern at a logger level, since _"native sink formatting"_ may vary.

alzix:
add per sink formatter

close #421

### DIFF
--- a/include/spdlog/details/file_helper.h
+++ b/include/spdlog/details/file_helper.h
@@ -82,11 +82,8 @@ public:
         }
     }
 
-    void write(const log_msg& msg)
+    void write(const void* data, size_t msg_size)
     {
-
-        size_t msg_size = msg.formatted.size();
-        auto data = msg.formatted.data();
         if (std::fwrite(data, 1, msg_size, _fd) != msg_size)
             throw spdlog_ex("Failed writing to file " + os::filename_to_str(_filename), errno);
     }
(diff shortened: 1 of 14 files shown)


alzix:
@wang-bin please test on Android

gabime:
Thats a lot of changes just to fix a small bug in the android sink :). I prefer fixing the bug directly in the android sink instead of modifying the entire library around it..

alzix:
it is not a bug fix but a whole new feature :)
i can move the new functionality closer to android_sink.h from sink.h - but i think it is a very useful functionality that worth keeping for all sinks.
BTW i think syslog will suffer from the same problem as android

alzix:
changes in pattern_formatter.h and in formatter.h are mandatory anyway

gabime:
This is big feature that need to be defined exactly and I am still not sure it worth the extra complexity and code changes..

alzix:
Well. I'm open for suggestions and for constructive feedback :)

wang-bin:
@alzix It works

alzix:
@gabime - do you want to park this PR? You are welcome to cherry pick the interesting parts (if you find any 😄 ) in case you decide to implement this feature in a different way.
In addition I'm opened for design proposals from your side.

@wang-bin this is a great news, thanks. Although i would suggest not to rush in to production with the new implementation, but wait until the new feature will find its new home on master branch.

gabime:
>  do you want to park this PR?

In the meantime, yes.

> You are welcome to cherry pick the interesting parts

Sure I will do that. Thanks!

Issue #516, opened 2017-08-31
Get log file name


fmilicchio:
Dear all,

I have a simple question: how do I get the logger's *file name*? 

I've searched but I've come to nothing. Currently I am using something like this:

    // Format
    spdlog::set_pattern("[%H:%M:%S] %v");
    
    // Queue size
    std::size_t q_size = 4096;
    
    // Asynchronous file logging
    spdlog::set_async_mode(q_size);
    
    // Init files and console
    outfile = spdlog::daily_logger_st("outfile", outfilename);
    logfile = spdlog::daily_logger_st("logfile", logfilename);
    console = spdlog::stdout_color_mt("console");

and files are create with this pattern: `filename.log_2017-08-31_16-50`, where everything after the `".log"` is generated by `spdlog`.

In alternative, I'd like to customize the file name, if possible.

Thanks!

s850042002:
Same question here.
It seems that we cannot set daily logger files' extension which is quite inconvenient when opening log files.

pszalanski:
Just stumbled across the same issue. An optional file extension would be nice, as well.

uralbash:
In v0.12.0 this feature exist, but this commit https://github.com/gabime/spdlog/commit/5585299b038e0e196bfe43719f81cec3f241dbd3 drop it

pszalanski:
Okay, so the feature would be to bring back the additional parameter for `calc_filename` and make it configurable at runtime.

gabime:
right. pull request would be most welcome..
or maybe just have a new
```c++
#define filename_ext ״.txt”
```
in tweakme.h .

gabime:
Fixed in 3c4a2bf5317219c685667fee65273e68bb4410a5

Now spdlog will automatically detect the extension and append it to rotated or daily files.

For example:
```c++
spdlog::daily_logger_st("logger", "mylog.txt");
```

would generate 


"**mylog**_2017-12-17_23-47.**txt**"

PR #565, opened 2017-11-25
issue-516: log file extension

asit-dhal:
added a macro in tweakme.h for log file extension

The files produced are like the following
async_log_2017-11-25_01-13.txt
async_log_2017-11-25_01-18.txt
basic.txt
daily_2017-11-25_01-13.txt


gabime:
Thanks for the pull. It fixes #516 but I dont think that ‘file_helper’ should be involved..
 ‘file_helper’  class is a small  os abstraction helper that should act exactly according to it’s method  parameters - with no surprises and side effects like altering filenames “on its own”.

asit-dhal:
Ok thanks !!

PR #570, opened 2017-12-06
Don't allow log file handle child process inheritance in Windows

joaomoreno:
Hi!

We started using spdlog in VS Code for a fast logging experience. It has been working great so far.

We did find an issue on Windows in which log file handles were being inherited by child processes https://github.com/Microsoft/vscode/issues/39659. This is problematic if we assume the file handle is released once we drop the logger, since the child processes will still have open handles to the log files. Once we attempt to instantiate a new logger on that same file path, we would get an exception after the 5 `f_fopen_s` attempts.

Our fix for this is to use the `N` mode character on Windows when opening the file. [This prevents file handle inheritance by child processes](https://msdn.microsoft.com/en-us/library/z5hh6ee9.aspx#Unicode%20support):

> N
> Specifies that the file is not inherited by child processes.

We assume that this is the desired behavior anyway in Windows.

cc @sandy081

### DIFF
--- a/include/spdlog/details/file_helper.h
+++ b/include/spdlog/details/file_helper.h
@@ -48,7 +48,7 @@ public:
     {
 
         close();
-        auto *mode = truncate ? SPDLOG_FILENAME_T("wb") : SPDLOG_FILENAME_T("ab");
+        auto *mode = truncate ? SPDLOG_TRUNCATE_MODE : SPDLOG_APPEND_MODE;
         _filename = fname;
         for (int tries = 0; tries < open_tries; ++tries)
         {
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -366,6 +366,8 @@ inline size_t thread_id()
 // wchar support for windows file names (SPDLOG_WCHAR_FILENAMES must be defined)
 #if defined(_WIN32) && defined(SPDLOG_WCHAR_FILENAMES)
 #define SPDLOG_FILENAME_T(s) L ## s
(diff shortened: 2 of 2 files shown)


gabime:
Hi,

Wow, that's great news. 

spdlog already provides an option  to prevent  child process inheritance in Windows. 
You need to uncomment  in [tweakme.h](https://github.com/gabime/spdlog/blob/master/include/spdlog/tweakme.h) (or define it before including spdlog):

```c++
#define SPDLOG_PREVENT_CHILD_FD
```

I hope it will solve your issue  -it implemented in a different way, using 
```c++
SetHandleInformation(file_handle, HANDLE_FLAG_INHERIT, 0)
```

joaomoreno:
Oh, great! We'll just use that instead! Thanks @gabime, great job!

joaomoreno:
@gabime That flag works perfectly, thanks for the tip!

gabime:
btw if you use aync logging, you might want to upgrade to 0.16.0 that will be released in the next few days.
Most notably in async logging, the back thread  busy wait loop (when the queue is empty) will be reduced from 5/sec to 2/sec to reduce cpu waste.

Also, rotated log files will preserve their extension (“log.3.txt” instead of the current “log.txt.3”)

joaomoreno:
Excellent, I created an issue on us to remember this. Thanks!

> Also, rotated log files will preserve their extension (“log.3.txt” instead of the current “log.txt.3”)

It's like you read our minds.

PR #575, opened 2017-12-11
Dirty fix for Intel compiler compatibility.

cynthia:
Intel compiler seems to have slightly different behavior when it comes to handling symbols that reference TLS members than what os.h expects. This is by no means a pretty fix, but more of a quick fix to fail compiling entirely when building against the Intel compiler on macOS.

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -359,8 +359,8 @@ inline size_t thread_id()
 #if defined(SPDLOG_DISABLE_TID_CACHING) || (defined(_MSC_VER) && (_MSC_VER < 1900)) || (defined(__clang__) && !__has_feature(cxx_thread_local))
     return _thread_id();
 #else // cache thread id in tls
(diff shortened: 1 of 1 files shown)


gabime:
so what is the fix exactly? is it removingthe  “inlne”? changing the variable name?

cynthia:
It takes away the inline due to how Intel compiler deals with TLS variables, and works around a variable name collision. It is not a very pretty fix - I need to look into exactly what icpc does differently for a better fix. I'm strongly suspecting a corner case bug that will need to be worked around, but might as well be wrong.

gabime:
ok. i will wait with the merge until you investigate more.
Also removing inline in header file would cause linking problems

cynthia:
Yeah, this isn't the right patch for this problem. Apologies for the noise. I have a strong hunch this is a compiler bug, but I don't have a good idea to fix this from the back of my head.

PR #583, opened 2017-12-17
Fixed SPDLOG_TRACE bug when using a variable for fmt

### LINKED ISSUE #529: SPDLOG_TRACE doesn't compile when given a variable as the `fmt` argument
```
const char* blah = "I am a string";
SPDLOG_TRACE(logger, blah);
```

This fails to compile with the following errors:

```
myfile.cpp: error: expected ‘)’ before ‘blah’
         SPDLOG_TRACE(logger, blah);
                              ^
spdlog/spdlog.h:175:103: note: in definition of macro ‘SPDLOG_TRACE’
   #define SPDLOG_TRACE(logger, ...) logger->trace("[ " __FILE__ ":" SPDLOG_STR_HELPER(__LINE__) " ] " __VA_ARGS__)
                                                                                                       ^~~~~~~~~~~
```

Using the literal explicitly works fine:

```
SPDLOG_TRACE(logger, "I am a string");
```

This compiles without complaint.

Using variables for arguments past the one representing `fmt` also does the job:

```
const char* blah = "I am a string";
SPDLOG_TRACE(logger, "format string {}", blah);
```

This also compiles. It's only when a variable is used as the first argument (besides the logger).

pabloariasal:
Fixes #529

### DIFF
--- a/include/spdlog/spdlog.h
+++ b/include/spdlog/spdlog.h
@@ -167,20 +167,18 @@ void drop_all();
 #ifdef SPDLOG_TRACE_ON
 #define SPDLOG_STR_H(x) #x
 #define SPDLOG_STR_HELPER(x) SPDLOG_STR_H(x)
-#define SPDLOG_CONCAT_STR(str1, str2, res) \
-do { res = (char*) malloc((strlen(str1)+strlen(str2)+1)*sizeof(char)); strcpy(res, str1); strcat(res, str2);} while(0)
 #ifdef _MSC_VER
 #define SPDLOG_TRACE(logger, fmt, ...) \
 do { \
-    char* s; SPDLOG_CONCAT_STR("[ " __FILE__ "(" SPDLOG_STR_HELPER(__LINE__) ") ] ", fmt, s); \
-    logger->trace(s, ##__VA_ARGS__); \
+    auto s = std::string("[ " __FILE__ "(" SPDLOG_STR_HELPER(__LINE__ ") ] ") + fmt; \
+    logger->trace(s.c_str(), ##__VA_ARGS__); \
 } \
 while(0)
 #else
 #define SPDLOG_TRACE(logger, fmt, ...) \
 do { \
-    char* s; SPDLOG_CONCAT_STR("[ " __FILE__ ":" SPDLOG_STR_HELPER(__LINE__) " ] ", fmt, s); \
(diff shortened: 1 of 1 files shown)


gabime:
Thanks! I was looking for a way to concat those strings during compile time though (and avoid creating  2 std::stirings on each trace call).  If you coud find such way, it would be great

pabloariasal:
That was in fact my original intention, but had a hard time figuring out how to do this at compile time. I was thinking on the lines of a `constexpr` function but didn't study the possibility further. Concatenating two string literals forced me to either create a temporary char[] (of which size?) or do a malloc (and worry about memory leaks). I decided to use `std::string` given that the `std::string(const char* c)` constructor is very efficient and seemed a lot safer. I made a mistake, though, as the `+` operator in `basic_string` overloads for (std::string, char *), meaning only one instantiation of `std::string` is required and not two. Will give the compile time concatenation some more thoughts.

godbyk:
I took a shot at it, relying on `fmt::format` (which I think uses `constexpr`, but I haven't checked):

```c++
#ifdef SPDLOG_TRACE_ON
#ifdef _MSC_VER
#define SPDLOG_TRACE(logger, s, ...) logger->trace(fmt::format("[ {}({}) ] {}", __FILE__, __LINE__, (s)).c_str(), ##__VA_ARGS__);
#else
#define SPDLOG_TRACE(logger, s, ...) logger->trace(fmt::format("[ {}:{} ] {}", __FILE__, __LINE__, (s)).c_str(), ##__VA_ARGS__);
#endif
#else
#define SPDLOG_TRACE(logger, s, ...)
#endif
```

This is just a first pass. I haven't tested it much yet, but it does work in the following cases:

```c++
SPDLOG_TRACE(console, "a trace message {} {}", 1, 2);
const char* blah = "I am a string and int {}";
SPDLOG_TRACE(console, blah, 0);
const char* blah2 = "Testing string only";
SPDLOG_TRACE(console, blah2);
const std::string blah3 = "Testing std::string only";
SPDLOG_TRACE(console, blah3);
```

Tested with clang++ and g++, but my understanding is that it should work with MSVC, too.

gabime:
@godbyk Looks good,  can you open a PR?

gabime:
Fixed in de595fe2b782763109d9a3b64aaf8b070a7420ed

Thanks@pabloariasal and @godbyk  for your help!

PR #585, opened 2017-12-20
feature/version file

horar:
We can generate version file from CMake project file. This reduce version string location to one place i.e. main CMake file.

In future, we can improve it so it will handle git hash etc. for non-tagged build etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ example/*
 
 # generated files
 generated
+include/spdlog/version.h
 
 # Cmake
 CMakeCache.txt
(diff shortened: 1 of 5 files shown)


godbyk:
For my projects, I use [GetGitRevisionDescription.cmake](https://github.com/OSVR/SteamVR-OSVR/blob/master/cmake/GetGitRevisionDescription.cmake), [GetGitRevisionDescription.cmake.in](https://github.com/OSVR/SteamVR-OSVR/blob/master/cmake/GetGitRevisionDescription.cmake.in), and code [like this](https://github.com/OSVR/SteamVR-OSVR/blob/master/src/CMakeLists.txt#L24).

gabime:
Thanks. Too much cmake magic for me though :)

horar:
@godbyk yes, I'm using similar approach in openconnect-gui - anyway, there has to be versioned version string somewhere, because when people export the project and then there is no way how to get the info from version control system in this case :/ ...

Issue #580, opened 2017-12-15
Custom filename creation for all sinks


ldionne:
Currently, I think only the daily logger allows customizing the filename format. I'm trying to replace our internal logging framework by Spdlog, and I need to encode the date in the filename (but without using a daily logger, which is too coarse grained). Instead, I want to generate a new filename every time a size-based sink "rotates". I might be able to provide a PR if there's interest for solving this issue.

gabime:
Thanks. That would be useful.

shazni:
Do we have an example of out to achieve this?

gabime:
Do we have an example of how to achieve this?
The [daily sink](https://github.com/gabime/spdlog/blob/master/include/spdlog/sinks/file_sinks.h#L192) already has such possibility

fawdlstty:
How is the progress on this issue?

alokpr:
I need this feature as well - specifically using "datetime" instead of "index" suffix for rotated files - similar to that provided by logrotate dateformat with second resolution.

Is there any progress on this issue or a suggested workaround to achieve this?

fawdlstty:
https://github.com/gabime/spdlog/pull/1847

alokpr:
IIUC daily sink only rotates daily. I need file size based rotation.

KonradJPE:
I also need this feature. Is there any general schedule regarding this topic?

rhaport:
I would need this feature as well for size based rotation

twarg1:
Any updates to the issues?:
1) allow file syncs to customize file names (e.g. add process name, add timestamp, etc.)
2) rotate file based on filesize
3) create new log on process crash
Thanks!

PR #601, opened 2017-12-30
Custom filename creation for rotate file sink.

lisongmin:
impl #580 

i also like size rotating with date, and commit this PR.
This impliment base on my real world requirement.

Size base rotate with the filename format `{base}.{index}.{ext}`,
this implement extend the `{base}` part to `{base}{extra}` which `{extra}` may be date.

Say we has log filename `rotated.txt`, the rotating should be:

* rotated.txt -> rotated_2017-12-30.1.txt

and next rotating on the same date should be:

* rotated.txt -> rotated_2017-12-30.2.txt

I'd like sequence increasement index,
just because I do not delete the file rotated for business needed.
And increasement index is more human readable.

The rotating_file_sink_mt interface is compat to previous version, and I add rotating_file_with_date_sink_mt interface.

### DIFF
--- a/include/spdlog/sinks/file_sinks.h
+++ b/include/spdlog/sinks/file_sinks.h
@@ -58,35 +58,18 @@ private:
 typedef simple_file_sink<std::mutex> simple_file_sink_mt;
 typedef simple_file_sink<details::null_mutex> simple_file_sink_st;
 
-/*
- * Rotating file sink based on size
- */
-template<class Mutex>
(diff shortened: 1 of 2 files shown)


gabime:
I think maybe a more suitable approach is a new kind of logger - rotating and daily combined

lisongmin:
Do you mean do not change the origin rotating_file_sink?

so what about this: https://github.com/lisongmin/spdlog/commit/34c0a0af70fdd93490d39f162eacc92db4540225

I will make a new PR if it is suitable.

PR #604, opened 2018-01-03
Ensure marcos always expand to expressions

sam-lunt:
Currently, the `SPDLOG_TRACE` and `SPDLOG_DEBUG` expand to nothing if they are disabled, but this can lead to problems. `assert` expands to `(void)0` when `NDEBUG` is defined, so I use the same here.

A couple of possible problems that could happen with the current definitions:

This is valid when `SPDLOG_DEBUG_ON` is defined, but expands to `,;` (which is an error) if not:
```
SPDLOG_DEBUG(logger, "x = {}", x), SDPLOG_DEBUG(logger, "y = {}", y);
```

If `SPDLOG_DEBUG_ON` is defined, this is a compilation error, but if not defined then then if statement actually catches the assignment to `b`:
```
bool b;
if (b)
    SPDLOG_DEBUG(logger, "b was true")
b = false;
````

### DIFF
--- a/include/spdlog/spdlog.h
+++ b/include/spdlog/spdlog.h
@@ -165,24 +165,23 @@ void drop_all();
 ///////////////////////////////////////////////////////////////////////////////
 
 #ifdef SPDLOG_TRACE_ON
-#define SPDLOG_STR_H(x) #x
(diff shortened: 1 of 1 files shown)


gabime:
Good catch. Thanks

PR #605, opened 2018-01-03
Add global flush_on function

sam-lunt:
This provides a way to set the flush level for all loggers. It replicates the interface and implementation of the `set_level` free function.

I set the default flush level to `off`, which is what `logger_impl` defaults to, so there should be no change of behavior.


gabime:
Thanks. Looks good

PR #606, opened 2018-01-07
Add customizable rotating file sink which support date in filename.

lisongmin:
Hi, 

I reimplement of #601 by standalone sink, is it suitable?


gabime:
Hi, Thanks for the pull, but I am planning to remove all those templates for custom file names and just just use lambdas provided by the user instead. Should be cleaner, I hope.

PR #610, opened 2018-01-11
Use Sleep in Windows instead of sleep_for

### LINKED ISSUE #609: Windows: sleep_for is dependent on system clock
Hey there! 👋 

We've hit this issue in VS Code: https://github.com/Microsoft/vscode/issues/41136

Ultimately here's the problem: `std::this_thread::sleep_for` is poorly implemented in Windows and depends on the system clock. When the thread is sleeping and the system clock is changed to a previous time, the sleep takes longer than expected, apparently by as much as the system clock was changed.

The bug easily reproduces with this snippet:

```cpp
#include "stdafx.h"
#include <iostream>
#include <chrono>
#include <thread>

int main()
{
	while (true) {
		std::cout << "sleeping for 500ms" << std::endl;
		std::this_thread::sleep_for(std::chrono::milliseconds(500));
	}

    return 0;
}
```

Simply go back in time and one of the 500ms `sleep_for` calls will take much longer.

---

I found a comment in [here](https://developercommunity.visualstudio.com/content/problem/61684/stdthis-threadsleep-for-depends-on-system-time.html) which indicates that `sleep_until` doesn't have the same issue, though I don't really verify that. I tried using both `system_clock` and `steady_clock` to compute the `sleep_until` argument; the bug still happens.

Of course, using the WinAPI call `Sleep` works like a charm.

Would you accept a PR in which I replace `sleep_for` calls for `Sleep` calls in Windows?

joaomoreno:
Fixes #609

@gabime Please give it a proper review... I'm not too comfortable in C++ to make such changes.

### DIFF
--- a/include/spdlog/details/async_log_helper.h
+++ b/include/spdlog/details/async_log_helper.h
@@ -358,7 +358,6 @@ inline void spdlog::details::async_log_helper::set_formatter(formatter_ptr msg_f
 // spin, yield or sleep. use the time passed since last message as a hint
 inline void spdlog::details::async_log_helper::sleep_or_yield(const spdlog::log_clock::time_point& now, const spdlog::log_clock::time_point& last_op_time)
 {
-    using namespace std::this_thread;
     using std::chrono::milliseconds;
     using std::chrono::microseconds;
 
@@ -374,10 +373,10 @@ inline void spdlog::details::async_log_helper::sleep_or_yield(const spdlog::log_
 
     // sleep for 20 ms upto 200 ms
     if (time_since_op <= milliseconds(200))
-        return sleep_for(milliseconds(20));
+        return spdlog::details::os::sleep_for_millis(20);
 
     // sleep for 500 ms
-    return sleep_for(milliseconds(500));
+    return spdlog::details::os::sleep_for_millis(500);
 }
 
 // wait for the queue to be empty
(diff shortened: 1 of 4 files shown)


gabime:
Thanks looks good. Please use  ‘details::os::sleep_millis’ without the spdlog:: prefix to be consistant with the rest of the code in async_helper.

 But why does it fix the problem? the sleep is in different thread so how does it hangs  the main window? The queue gets full?

joaomoreno:
It's only on shutdown, when flush is called. Then, `wait_empty_q` is called and gets stuck looping in [here](https://github.com/gabime/spdlog/blob/master/include/spdlog/details/async_log_helper.h#L387.
 The queue doesn't get empty until the worker thread wakes from that long sleep.

joaomoreno:
Pushed the namespace changes.

gabime:
Thanks @joaomoreno

joaomoreno:
Awesome stuff. Let me know if you'll release the fix soon, I'll make sure we move to 0.16 with the adoption.

gabime:
I will release it in v0.16.3 in the next few days.

gabime:
@joaomoreno v0.16.3 was released with this fix.

joaomoreno:
Excellent, thanks for the ping!

PR #615, opened 2018-01-18
A few small ANSI color improvements

Qix-:
Hey there 👋 

This PR does the following:

- Renames `gray` to `black` as it is more accurate (`30m` is officially specified as "black")
- Removes needless zeros from the `reset` code
- Clears the line after writing a log message, which fixes `->critical()` log messages bleeding their background color onto the new line. This is because the log messages have the newline _before_ the reset code, thus forcing most xterm-compliant emulators to paint the entire next line red before writing the next log message.

Before:
image

After:
image

### DIFF
--- a/include/spdlog/sinks/ansicolor_sink.h
+++ b/include/spdlog/sinks/ansicolor_sink.h
@@ -49,7 +49,7 @@ public:
     }
 
     /// Formatting codes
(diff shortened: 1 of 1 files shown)


gabime:
Thanks! I always wondered what cause the bakckground bleed,  but never had the time to address this.

PR #619, opened 2018-01-21
Implemented delay open option for simple and size rotating file sinks

emruiz81:
This patch allows to create the files on demand for the mentioned file sinks.
The application setups spdlog with all the corresponding parameters, the file name, the log level, etc. And starts working. With this parameter set to true, the log files won't be created unitl the first real message arrives to the file sink. 
This is based on the delay open parameter of the Python logging library. And allows, for example, to avoid having empty log files when the applicaton wasn't logging anything.

### DIFF
--- a/include/spdlog/sinks/file_sinks.h
+++ b/include/spdlog/sinks/file_sinks.h
@@ -29,9 +29,17 @@ template<class Mutex>
 class simple_file_sink SPDLOG_FINAL : public base_sink < Mutex >
 {
 public:
-    explicit simple_file_sink(const filename_t &filename, bool truncate = false):_force_flush(false)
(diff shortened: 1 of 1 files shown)


gabime:
@emruiz81 Thanks for the pull (and sorry for late response).

For simplicity sake, I prefer to have clear picture when the file is opened (in the sink ctor)  - one can always defer creating the logger to later if this is a problem for him,

gabime:
@emruiz81  I think your original pr was ok.. No need for new macros.
I will try several approaches and decide.

emruiz81:
Ok. Let me know if I can help you with something else.
Thanks!

PR #622, opened 2018-01-25
New function to convert level_enum from string

fegomes:
## Requeriment
Configure log level using name. (trace, debug, info...)

## Solution
Creation a new function to convert level_enum from string  - **to_level_enum(const char*)** 

## Example
``` cpp
spdlog::level::level_enum lvl = spdlog::level::to_level_enum("debug");
```

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -51,6 +51,9 @@ namespace spdlog
 
 class formatter;
 
+template<class T, size_t N>
+constexpr size_t size(T(&)[N]) { return N; }
+
 namespace sinks
 {
 class sink;
@@ -98,6 +101,18 @@ inline const char* to_short_str(spdlog::level::level_enum l)
 {
     return short_level_names[l];
 }
+inline spdlog::level::level_enum to_level_enum(const char* name)
+{
+	for (size_t level = 0; level < size(level_names); level++)
+	{
+		if (!strcmp(level_names[level], name))
+		{
+			return (spdlog::level::level_enum) level;
+		}
+	}
+	return (spdlog::level::level_enum) 0;
+}
+
 } //level
 
 
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -56,9 +56,6 @@ namespace spdlog
 
 class formatter;
 
-template<class T, size_t N>
-constexpr size_t size(T(&)[N]) { return N; }
-
 namespace sinks
 {
 class sink;
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -87,16 +87,6 @@ enum level_enum
     off = 6
 };
 
-static std::unordered_map<std::string, level_enum> name_to_level = {
-                                                                        { "trace"   , level::trace },
-                                                                        { "debug"   , level::debug },
-                                                                        { "info"    , level::info },
-                                                                        { "warning" , level::warn },
-                                                                        { "error"   , level::err },
-                                                                        { "critical", level::critical },
-                                                                        { "off"     , level::off }
-                                                                   };
-
 
 #if !defined(SPDLOG_LEVEL_NAMES)
 #define SPDLOG_LEVEL_NAMES { "trace", "debug", "info", "warning", "error", "critical", "off" }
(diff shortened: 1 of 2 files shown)


gabime:
Thanks for the pull.  Why is the use case  here?

Also, A lookup table(```unordered_map<string, level>```) should be considered for better performance (but need to be benched - for such small array not user it is faster)

fegomes:
## Example:

``` ini
// log.ini
[General]
path=/home/user/log
pattern=%+
level=debug -- it is more simple using name for users
```

I didn`t make with unordered_map because you maked rule to redefine the SPDLOG_LEVEL_NAMES macro

gabime:
>I didn`t make with unordered_map because you maked rule to redefine the SPDLOG_LEVEL_NAMES macro

How about:
```c++
inline spdlog::level::level_enum from_name(const char* name)
{
	static std::unordered_map < std::string, level_enum> name_to_level
	{
		{ level_names[0] , level::trace },
		{ level_names[1] , level::debug },
		{ level_names[2] , level::info },
		{ level_names[3] , level::warn },
		{ level_names[4] , level::err },
		{ level_names[5] , level::critical },
		{ level_names[6] , level::off }
	};

	auto lvl = name_to_level.find(name);
	return lvl != name_to_level.end() ? lvl->second : level::off;						
}
```

fegomes:
It's so good, but the users can't customize SPDLOG_LEVEL_NAMES macro. How you can see in your code

**spdlog/common.h:85**

``` cpp
#if !defined(SPDLOG_LEVEL_NAMES)
#define SPDLOG_LEVEL_NAMES { "trace", "debug", "info",  "warning", "error", "critical", "off" }
#endif
```

fegomes:
I think this function is simple and I don't imagine call many times during application running, so I preserved the compatibility with your SPDLOG_LEVEL_NAMES macro.

gabime:
> It's so good, but the users can't customize SPDLOG_LEVEL_NAMES macro. How you can see in your code

Why not? I tried and it worked, You need to define SPDLOG_LEVEL_NAMES before including spdlog.h

fegomes:
Yeah it's works, but the function "from_name" or "to_level_name" doesn't works, when user define SPDLOG_LEVEL_NAMES in your code.

### example

``` cpp
#define SPDLOG_LEVEL_NAMES { "customize1", "customize2", "customize3" }
#include <spdlog/spdlog.h>
...
```
Could you create a  protection if user define SPDLOG_LEVEL_NAMES, How about in spdlog.h

``` cpp
// spdlog.h
#pragma once

#ifdef SPDLOG_LEVEL_NAMES 
    #error "Error: You must not define SPDLOG_LEVEL_NAMES" 
#endif
```

include/spdlog/common.h
```diff
@@ -51,6 +51,9 @@ namespace spdlog
 
 class formatter;
 
+template<class T, size_t N>
+constexpr size_t size(T(&)[N]) { return N; }
```
boimart1: VC 2013 doesn't support `constexpr`

fegomes:
@gabime,  What do you think about commits?

gabime:
@fegomes Looks good. What about the SPDLOG_LEVEL_NAMES  ?

fegomes:
Are you saying about compile protection or name of variable?

gabime:
To be able to handle custom log levels which might be defined in SPDLOG_LEVEL_NAMES

fegomes:
I didn't see benefits to customize **SPDLOG_LEVEL_NAMES**  because your enum level_enum is static.

And more, if user not define 7 values, It's causes runtime error.

gabime:
> I didn't see benefits to customize SPDLOG_LEVEL_NAMES because your enum level_enum is static.

Doesn't matter - we need to decide whether to take into account that someone would use it. Do we want ```from_name``` respect it or not? I think yes

> And more, if user not define 7 values, It's causes runtime error.

This would happen anyway - regardless of the ```from_name``` function

fegomes:
> Doesn't matter - we need to decide whether to take into account that someone would use it. Do we want from_name respect it or not? I think yes

Ok! ;).. I changed the code.

PR #635, opened 2018-02-11
CMake explicit language

DanielChabrowski:
Hey, a few changes to cmake files.

Marking project as CXX will disable detecting C compiler and other C related checks.

'INCLUDES DESTINATION' adds paths to 'INTERFACE_INCLUDE_DIRECTORIES' which is where include paths already are in spdlog target.

Bumped example's cmake version to 3.1.

Both test and example CMakeLists were using CRLF, changed to LF to match other source files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 #
 
 cmake_minimum_required(VERSION 3.1)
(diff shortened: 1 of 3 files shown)


gabime:
Thanks!

Issue #643, opened 2018-02-20
Support for writing binary data to a logger


rob-p:
Hi @gabime,

  Thank you for this fantastic library!  We've been using it extensively in our lab's projects, and it's been awesome.  I've been particularly surprised with how versatile `spdlog` is; as it allows us to quite cleanly separate the concerns of what we want to write out from _how_ we want to write it (e.g. [here](https://github.com/COMBINE-lab/pufferfish/blob/develop/src/PufferfishAligner.cpp#L1139), where we generalize over stdout vs. a file and uncompressed vs. compressed).  We're particularly fond of using `spdlog` as a means to sync up output of messages from many different threads to a file in a threadsafe manner.

  However, there's one thing we'd like to do, and it's not clear to us how to achieve this (or if it is even possible) in spdlog.  Basically, we're using a log with an `ostream` sink as a way to write output files from our program.  This is actually not a "log", but the main output of the tool that records data about how sequencing reads (from DNA and RNA) map to a target genome.  The output of this tool will be consumed by another tool we have written for further processing.  Because we know exactly how we want to represent each output record, it would be _much_ more efficient to do this in a binary format rather than a text-based one.  We'd still like to do this using `spdlog` as our entire output system is based around its excellent multithreaded, asynchronous logging to allow writing to file from many threads at once.  It's not clear how to do this (or if it is even possible).  I realize that this may be outside of the supported scope of spdlog, but it would be great to get your feedback / input on if there is a way to do this, and if so, what the best design might look like.

Thanks!
Rob

gabime:
Hi,
Thanks for the nice feedback. I appreciate it!

Regarding your question, one solution the comes to mind is to define a custom logabble class:
```c++
class my_binary_record
{
private:
	const std::vector<char> _bin_data;

public:
	//copy bin data to this record
	my_binary_record(const std::vector<char>& bin_data) :_bin_data(bin_data) {};
	//or just move if possible
	my_binary_record(std::vector<char>&& bin_data) :_bin_data(move(bin_data)) {};

      //support for logging directly from spdlog
      template<typename OStream>
      friend OStream& operator<<(OStream& os, const my_binary_record &bin_record)
      {        
  		std::ostream_iterator<char> out_iter(os);		
		std::copy(bin_record._bin_data.begin(), bin_record._bin_data.end(), out_iter);		
		return os;
     }
};
```

And use it like this:
```
bin_logger->set_pattern(""); // disable formatting - will not be used
[...]

PR #646, opened 2018-02-23
Adding extra meta information to log records

lion10243:
This PR adds the optional, OFF-by-default SPDLOG_ENABLE_LOG_ATTRIBUTES feature to spdlog.

It enables log messages to store extra data (attributes) in the same style as Boost::Log Attributes. Attributes are stored as a key-value map (string -> variant) against each `detail::log_msg`. Attribute values can be: bool, int, long, float, double, std::string. See common.h for declaration. More types can be added.

While std::variant is officially c++17, the author of std::variant (MPark) has made available a version compatible with c++11 and c++14.

Comments, suggestions welcome.


gabime:
Thanks..
Where is the code that consumes(and logs?) those values? I couldn't find it..

include/spdlog/logger.h
```diff
@@ -44,6 +44,17 @@ class logger
     template <typename Arg1, typename... Args> void error(const char* fmt, const Arg1&, const Args&... args);
     template <typename Arg1, typename... Args> void critical(const char* fmt, const Arg1&, const Args&... args);
 
+#ifdef SPDLOG_ENABLE_LOGMSG_METADATA
+    template <typename... Args> void log(level::level_enum lvl, attrmap_type& ma, const char* fmt, const Args&... args);
+    template <typename... Args> void log(level::level_enum lvl, attrmap_type& ma, const char* msg);
```
boimart1: shouldn't these all be `attrmap_type&&`? Since you're moving from `ma` inside the implementations.

lion10243:
I need to provide tests to demonstrate that. Earliest will be next week.
Currently the attributes are only accessible if one writes a custom sink
and implements the `log(details::log_msg const&)` method.

Am 23.02.2018 22:34 schrieb "Gabi Melman" <notifications@github.com>:

lion10243:
I do agree with you. My first implementation locally used move semantics.
However I experienced a number of annoying compile errors where clang (3.9)
failed to deduce the correct template specialisation ("attrmap_type ma"
passed as Val, instead of "attrmap_type&& ma" as move), thus in the
callstack between first `spdlog::log()` call in user's code to the
internal `details::logger_impl::log()` method there were between 2-4 times
std::move on the `ma` param.

Pragmatically speaking, i see the use of reference `ma` param, followed by
one final move, easier and cleaner to implement, however you are right in
stating that move semantics is technically correct.


Am 23.02.2018 22:39 schrieb "Martin Boisvert" <notifications@github.com>:

*@boimart1* commented on this pull request.
------------------------------

In include/spdlog/logger.h
<https://github.com/gabime/spdlog/pull/646#discussion_r170374765>:

> @@ -44,6 +44,17 @@ class logger
     template <typename Arg1, typename... Args> void error(const char*
fmt, const Arg1&, const Args&... args);
     template <typename Arg1, typename... Args> void critical(const
char* fmt, const Arg1&, const Args&... args);

+#ifdef SPDLOG_ENABLE_LOGMSG_METADATA
+    template <typename... Args> void log(level::level_enum lvl,
attrmap_type& ma, const char* fmt, const Args&... args);
+    template <typename... Args> void log(level::level_enum lvl,
attrmap_type& ma, const char* msg);

shouldn't these all be attrmap_type&&? Since you're moving from ma inside
the implementations.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/gabime/spdlog/pull/646#pullrequestreview-99060787>, or mute
the thread

tests/test_log_attrs.cpp
```diff
@@ -0,0 +1,71 @@
...
+    SECTION("attributes test") {
+        // supported types: int, double, float, bool, std::string
+        spdlog::attrmap_type attrs {
+            {"param_int", 42},
+            {"param_double", 42.5},
+            {"param_float", 42.5},
+            {"param_bool", true},
+            {"param_string", "spdlog feature test"},
+        };
+        REQUIRE(log_to_str(attrs, "attributes enabled") == "attributes enabled");
```
fegomes: Where is the example using it? you set it up attrmap_type but I didn't see the example using this case. Could you better your test?  or  exemplifly the use?

lion10243: It was in the log_to_str() func. I've moved it out and added a comment. The actual Catch checks for the attribute values are in NullSink::log().

fegomes:
I think your pull request is too big, you could to simplify your request to @gabime analysis better.

lion10243:
@fegomes @gabime   too big in the sense that a number of new method signatures were added in `logger.h` ? Arguably Yes. These are necessary to cover the different forms of `logger::log()` api.

Potentially these extra methods in `logger.h` could be extracted into a subclass, e.g. `attribute_logger.h`, or even mostly removed, apart from the 2-3 core `logger::log()` signatures, which need overrides to accept the `attrmap_type` attribute map type.

What I really specifically need from the library is the support for the attributes in `detail::log_msg`, and the related 2-3 `logger:log()` funcs which accept a map of key-val data. The rest could be removed.

I'm happy to compromise and try to reach a consensus about what would be acceptable, providing we can get the core 'attributes' feature into the library. I really need spdlog to support the log attributes feature that Boost::Log has.

gabime:
@lion10243 I agree with @fegomes . I am  reluctant in adding so many changes..Especially since it requires a new header (pre cpp17 variant support header), and that the gain is not very obvious..

Also, if the main need is to add context info, maybe there is no need for variant at all - a simple object pointer could be passed around.

lion10243:
@gabime @fegomes thanks for the feedback so far.

The aim is to implement extra context data (key-val, where vals can be
different data types) for a log message, data which might not be encoded
directly in the log msg string but is required by the log sink.

I don't know if you're both familiar with FluentD logging daemon and
Kibana?  We've written a FluentD spdlog sink and need to write extra
context meta data (shop, partner, market etc) which is required by FluentD
daemon for the Kibana backend but should not exist in the log message text.
The idea is directly comparable to a feature in boost::log called
'Attributes'. Contextual log message metadata, basically.

@gabime I'm not sure how to implement the key-val map with a flexible data
types for the Val field without effectively reimplementing a type like
std::variant to some extent. In worst case a string-string key-val map
would suffice. Perhaps you could exand upon your suggestion with a little
code?

Anyway, does it make sense, what I'm trying to achieve?

Let me rethink the PR, make it simpler. Having seen the implementation of
the async logger, I reckon a similar technique would work; subclass the
logger (e.g. attribute_logger) and implement a logger-specific log_msg
strict  (e.g. attribute_log_msg). Then no need to modify logger.h directly
etc.

Thanks for the attention after a long comment :)


Am 26.02.2018 15:19 schrieb "Gabi Melman" <notifications@github.com>:

@lion10243 <https://github.com/lion10243> I agree with @fegomes
<https://github.com/fegomes> . I am reluctant in adding so many
changes..Especially since it requires a new header (pre cpp17 variant
support header), and that the gain is not very obvious..

Also, if the main need is to add context info, maybe there is no need for
variant at all - a simple object pointer could be passed around.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/gabime/spdlog/pull/646#issuecomment-368517615>, or mute
the thread

lion10243:
Having tried to refactor the PR, there are too simply too many consequential refactorings for methods/functions in registry/core necessary to make the introduction of another, subclassed logger type successful - fewer refactoring than simply building the feature into logger.h. I will close the PR and maintain this feature privately.

gabime:
Ok. Thanks for the suggestion anyway. As a reference you might want to consider issue #643 - It has similiar aim (but not identical).
I think you could take the example there (defining loggable class and a custom sink) and twaek it to your needs.

PR #647, opened 2018-02-25
Applied some of clang-tidy fixes

DanielChabrowski:
Hello,
These are the fixes for some of the checks, not all of the checks, not all of the fixes. There is a ton of warnings coming from fmt library as it seems to support cpp98 and from catch.

btw. I know you said that astyle is enough for you but the project could really use clang-format.


gabime:
Thanks! Very interesting fixes. Why does it recommends passing strings by values and doing move ?
Why is 
```c++
spdlog::logger::logger(std::string name, const It& begin, const It& end):
    _name(std::move(name))
   ...
``` 
better than the original
```c++
spdlog::logger::logger(const std::string &name, const It& begin, const It& end):
    _name(name),
   ...
```

I assume it is good practice because that way it is more apparent that the class never really needs a string reference as it just copy the input right away ?

Anyway - please post here (or even better a PR) on the command line and options to produce all those tidy warnings.

> btw. I know you said that astyle is enough for you but the project could really use clang-format.
Agreed - please post a PR

DanielChabrowski:
@gabime 
About the move - https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html

`run-clang-tidy-5.0.py -p="build" -header-filter="include/spdlog/.*" -checks="-*,readability-else-after-return"`
Used it from the repo root with the checks as in commit messages. The problem is that it also reports warnings from fmt unless SPDLOG_FMT_EXTERNAL is defined. I'll see what I can do about the PR, no promises though.

I can see that you fixed gcc warnings after a merge. Checks that introduced those changes were cppcoreguidelines-pro-type-member and readability-inconsistent-declaration-parameter-name.

PR #649, opened 2018-02-28
Add clang format

DanielChabrowski:
Added and applied clang-format-5.0 for spdlog, tests and benchmarks. It has to be a specific version of clang-format, otherwise the output may vary. It has to check for both clang-format and clang-format-5.0 because of how package managers distribute it.

We may as well format the fmt library to get rid of that list filtering.

To format the code you have to call:
```
$ cmake
$ make format
```

Let me know what you think.


gabime:
@DanielChabrowski Thanks ! Very useful . 

Which style of formatting did you put in .clang-format ?
Can you please resend the PR but only withe the formatting addition ? (without applying it)

DanielChabrowski:
@gabime 
It's kind of a mix, I made it so long ago I don't even remember what I based it on.

Are you going to run it over the code?

gabime:
@DanielChabrowski  Yes- see latest commits

DanielChabrowski:
@gabime 
How about changing "AlwaysBreakTemplateDeclarations" to true?

```
-template <typename T> inline std::string format(const T &value)
+template <typename T>
+inline std::string format(const T &value)
```

I would also change CompactNamespaces to false and PointerAlignment to left but I guess these are just personal preferences. What do you think?

gabime:
Ok for the templates. As for other two, we’ll see.. Not sure.

PR #651, opened 2018-03-01
Create formatting_sink.h

hoerenj:
Proposing a decorator sink to address issue #421 per sink format pattern


hoerenj:
I have no clue how the Travis CI build problem is related to the proposed sink.

PR #652, opened 2018-03-03
Python

bodgergely:
Adding to README how to install python language binding.


gabime:
Please put it in the wiki pages instead

PR #653, opened 2018-03-05
Fix implicit conversion warnings (-Wsign-conversion)

tbastos:
Fix for some annoying warnings when using `-Wsign-conversion` with clang


gabime:
Thanks @tbastos

PR #654, opened 2018-03-06
Don't use fstat64 when _LARGEFILE_SOURCE is defined

cygnus9:
Using fstat64 on ancient distributions like RHEL5 causes errors because
fstat64 is declared inline, but there's no definition. Modern compilers
don't like that without the -fgnu89-inline flag.

> gcc-4.9.4/lib/gcc/x86_64-unknown-linux-gnu/4.9.4/include-fixed/sys/stat.h:255:16: error: inline function 'int fstat64(int, stat64*)' used but never defined
> __inline__ int fstat64 (int __fd, struct stat64 *__buf) __THROW __nonnull ((2));

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -248,7 +248,7 @@ inline size_t filesize(FILE *f)
 #else // unix
     int fd = fileno(f);
     //64 bits(but not in osx or cygwin, where fstat64 is deprecated)
-#if !defined(__FreeBSD__) && !defined(__APPLE__) && (defined(__x86_64__) || defined(__ppc64__)) && !defined(__CYGWIN__)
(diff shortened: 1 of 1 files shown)


gabime:
Not sure I understand. Why if _LARGEFILE_SOURCE is defined means that ```fstat`` and not ```fstat64``` should be used ? I can't find any documentation on this online

cygnus9:
To be honest I applied this patch on a local repo over a year ago and had to do it again after updating spdlog (which made me think it could be worth a PR). I don't remember how I came to this conclusion, but just grepping through some system headers leads me to be quite confident that all the size related fields in ```struct stat``` are actually 64 bit versions when ```_LARGEFILE_SOURCE``` is defined.

I think it's actually about ```__USE_FILE_OFFSET64``` being defined, but that seems to be defined transitively.

gabime:
@cygnus9 This is problematic to apply this if we not 100% sure about it or how would it affect other  platforms

PR #656, opened 2018-03-08
Fix a conflict of argument and namespace

k1988:
when use SPDLOG_WCHAR_TO_UTF8_SUPPORT macro, i got a compiler error C2760.
the argument name 'fmt' conflict with the namespace 'fmt'，so rename the argument is my choice.

The above paragraph may not be very clear, because it is automatically translated.

### DIFF
--- a/include/spdlog/details/logger_impl.h
+++ b/include/spdlog/details/logger_impl.h
@@ -58,7 +58,7 @@ inline void spdlog::logger::set_pattern(const std::string& pattern, pattern_time
 }
 
 template <typename... Args>
-inline void spdlog::logger::log(level::level_enum lvl, const char* fmt, const Args&... args)
+inline void spdlog::logger::log(level::level_enum lvl, const char* fmtStr, const Args&... args)
 {
     if (!should_log(lvl)) return;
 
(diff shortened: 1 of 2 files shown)


gabime:
why only this fumction then? there are more functions with this arg name

k1988:
but fmt::WMemoryWriter wWriter only  used in this function。other SPDLOG_WCHAR_TO_UTF8_SUPPORT functions not contains fmt::xxx

include/spdlog/fmt/bundled/printf.h
```diff
@@ -632,7 +632,7 @@ inline void printf(WWriter &w, WCStringRef format, ArgList args)
 {
     PrintfFormatter<wchar_t>(args, w).format(format);
 }
-FMT_VARIADIC(void, printf, WWriter &, WCStringRef)
+FMT_VARIADIC_W(void, printf, WWriter &, WCStringRef)
```
gabime: This is part of fmtlib - not spdlog - please issue pr about this in the fmtlib repo

include/spdlog/details/logger_impl.h
```diff
@@ -58,7 +58,7 @@ inline void spdlog::logger::set_pattern(const std::string& pattern, pattern_time
 }
 
 template <typename... Args>
-inline void spdlog::logger::log(level::level_enum lvl, const char* fmt, const Args&... args)
+inline void spdlog::logger::log(level::level_enum lvl, const char* fmtStr, const Args&... args)
```
gabime: We use camel_case, not snakeCase - please use "fmt_str" instead of "fmtStr"

include/spdlog/details/logger_impl.h
```diff
@@ -217,11 +217,16 @@ inline void spdlog::logger::log(level::level_enum lvl, const wchar_t* msg)
 }
 
 template <typename... Args>
-inline void spdlog::logger::log(level::level_enum lvl, const wchar_t* fmt, const Args&... args)
+inline void spdlog::logger::log(level::level_enum lvl, const wchar_t* fmtstr, const Args&... args)
```
gabime: We use camel_case, not snakeCase - please use "fmt_str" instead of "fmtStr"

k1988:
OK，Thank you for your patience.

Issue #147, opened 2015-11-19
Flush before close loses logs in async mode


oconnore:
Perhaps similar to #145, but in async mode. I haven't figured out a way to force all logs to write + flush to file before closing the program. 

The patch would likely involve changing the busy wait behavior here: https://github.com/gabime/spdlog/blob/cbc8ba72035a3828f3a03e425e7f0d91da6c2ba3/include/spdlog/details/async_log_helper.h#L265-L267 . Using a per-log [condition variable `wait_for`](http://en.cppreference.com/w/cpp/thread/condition_variable/wait_for) instead of a non-interruptible thread `sleep_for` would allow a log-flush to immediately drain the log queue, and then call per-sink flushes. 

Would you be open to a patch? Is there a reason to be busy waiting here at all?

gabime:
This is because the queue is lockfree, so using condition vaiables would defeat this purpose (user thread whould need to acuire the mutex of the condion var).

The cost of this is that when the queue is empty, the backend thread  would spin for 10 times/sec or so.. 

I think it is small price to stay lockless, but I could be wrong of course..

Perhaps overriding the logger::flush in the async logger to drain the queue and only then performing the  flush is a solution for this.

Another option would be chenge the destructor in async_log_helper to process all the messages in the queue before destruction..

gabime:
I think I modify the destructor to process all the messages in the queue before destruction..

gabime:
Fixed in latest commit. Now the logger will first process all messages in the queue before terminating

rob-p:
Awesome; thanks for this!

tcwalther:
I am still encountering this problem in my production app. It occurs when you terminate the program before flushing:

```cpp
#include <iostream>
#include <chrono>
#include <spdlog/spdlog.h>
#include <csignal>


int main() {
  try {
    size_t q_size = 1048576; //queue size must be power of 2
    spdlog::set_async_mode(q_size, spdlog::async_overflow_policy::block_retry,
                           nullptr, std::chrono::seconds(1));
    auto logger = spdlog::daily_logger_mt("main", "log");
    spdlog::set_level(spdlog::level::info);

    logger->info("Test");
    
    // now terminate program
    raise(SIGABRT);
  } catch (const spdlog::spdlog_ex& ex) {
    std::cout << "Log failed: " << ex.what() << std::endl;
  }
  return 0;
}
```

[...]

Issue #145, opened 2015-11-14
Blocking flush in multithreaded logger


rob-p:
Hi,

  First, thanks for spdlog.  It's a fantastic library and has become my go-to tool for logging in all of my projects.  It's lightweight header only nature, fast speed, and rich feature set make it a near-perfect choice!  I have a situation in which I'm using a multi-threaded logger, and I want to report a certain error condition to the user before the program exits.  The problem that I'm encountering is that *sometimes* the message appears and *sometimes* it does not (the program terminates before the message is written).  I'm making the call to flush the logger before I call `std::exit()`.  In the multithreaded logger, is there any way to *block* until the log has been flushed?  Right now, I'm reduced to the very inelegant solution of using `cerr` in the cases where I want to report an error right before program termination.

gabime:
Thanks for the feedback, I appriciate it.
Regarding the question, I assume you are using the sync mode with cerr logger?
I am not sure why flush doesnt help, but please try calling spdlog::drop_all() before exiting.

rob-p:
I'm not quite sure what you mean by the `sync` mode with cerr logger.  I create the logger in the following manner:

```
auto fileSink = std::make_shared<spdlog::sinks::ostream_sink_mt>(logFile);
auto consoleSink = std::make_shared<spdlog::sinks::stderr_sink_mt>();
auto consoleLog = spdlog::create("stderrLog", {consoleSink});
auto fileLog = spdlog::create("fileLog", {fileSink});
auto jointLog = spdlog::create("jointLog", {fileSink, consoleSink});
```

I then write to `jointLog`, and call `jointLog->flush_all()` before exiting.  I also tried additionally calling `spdlog::drop_all()` before exiting, but it doesn't seem to help.  I *think* that the issue may have something to do with the semantics of logging directly before `std::exit()`.  If I place the call in a separate function and force the function to return directly before a call to `exit()`, the message seems to be written to the log.  However, this solution is clearly not always possible.  Is there something specific about the way that the compiler handles `std::exit()` that may cause the logger to not be appropriately flushed?

gabime:
This is very strange indeed.. which os/compiler do you use?

rob-p:
I'm on ubuntu 14.10.  The result of `uname -a` is:

```
Linux feynman 3.16.0-23-generic #31-Ubuntu SMP Tue Oct 21 17:56:17 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
```
[...]

PR #664, opened 2018-03-19
Improve async flush

sam-lunt:
Using flush with an async logger is problematic, because the call can hang forever. This is due to the fact that flush waits for the queue to be empty, instead of waiting for just the messages in the queue at the time of the call to flush to be removed. If there is a logger on another thread writing to the queue faster than messages are being pulled off, the call to flush will never return.

This is particularly problematic because most code cannot know if a reference to a `spdlog::logger` is an async logger or not (it depends if `set_async_mode` was called). If a call to flush can cause deadlock if the logger is async and if there is no way to know (short of a `dynamic_cast`), then it is never safe to call flush on a logger reference.

The expectation for a call to flush is not "block until no messages are waiting to be written"; it's "block until all the messages currently on the queue are written". This change does that: instead of blocking until the queue is empty, it blocks until the dequeue position catches up to what the enqueue position was when flush was called.

As long as messages are being pulled off the queue, this guarantees that flush will eventually exit.

A few notes about the implementation:

- The waiting logic checks if `write_pos > read_pos` before setting `last_op`, which can save a call to `now` if no waiting is necessary. This should be slightly more efficient when the queue is already empty.

- This adds two accessors to the queue class, which is not ideal but better than making async_log_helper a friend class.

- Since `wait_empty_q` was private and only used in the flush method, I removed it.

- The queue function `is_empty` introduces a race condition and shouldn't be used, so I marked it as deprecated. It's not used anywhere in the spdlog project, but it's public, so it didn't seem wise to just delete it, as another project might be using it. The race condition in `is_empty` could lead to `is_empty` never returning true, even though the queue is empty at points in time.

For example:
thread 1 calls `is_empty` and reads `front = front1 = 1` and `back = 0`. `is_empty` returns false;
thread 2 pulls off a message, so now `dequeue_pos_` (aka `back`) is `1`. The queue is empty
thread 3 pushes a new message on to the queue, so now `enqueue_pos_` (aka `front`) is 2
thread 1 calls `is_empty` again and reads `front = front2 = 2` and `back = 1`. `is_empty` returns false

So a thread waiting on `is_empty` might do so forever, even though the queue is empty at several points in time. Even if it doesn't wait forever, it will end up waiting significantly longer than it likely wants to.

### DIFF
--- a/include/spdlog/details/async_log_helper.h
+++ b/include/spdlog/details/async_log_helper.h
@@ -182,8 +182,6 @@ private:
     // sleep,yield or return immediately using the time passed since last message as a hint
     static void sleep_or_yield(const spdlog::log_clock::time_point &now, const log_clock::time_point &last_op_time);
 
-    // wait until the queue is empty
-    void wait_empty_q();
 };
 } // namespace details
 } // namespace spdlog
@@ -248,7 +246,19 @@ inline void spdlog::details::async_log_helper::flush(bool wait_for_q)
     push_msg(async_msg(async_msg_type::flush));
     if (wait_for_q)
     {
-        wait_empty_q(); // return when queue is empty
+        auto write_pos = _q.enqueue_pos(std::memory_order_acquire);
+        auto read_pos  = _q.dequeue_pos(std::memory_order_relaxed); // synchronized by read of enqueue_pos_
+
+        if (write_pos > read_pos)
+        {
+            auto last_op = details::os::now();
+            do
+            {
+                sleep_or_yield(details::os::now(), last_op);
+                read_pos = _q.dequeue_pos(std::memory_order_acquire);
+            }
+            while (write_pos > read_pos);
+        }
     }
 }
 
@@ -375,16 +385,6 @@ inline void spdlog::details::async_log_helper::sleep_or_yield(
 
     // sleep for 500 ms
     return details::os::sleep_for_millis(500);
-}
-
-// wait for the queue to be empty
-inline void spdlog::details::async_log_helper::wait_empty_q()
-{
-    auto last_op = details::os::now();
-    while (!_q.is_empty())
-    {
-        sleep_or_yield(details::os::now(), last_op);
-    }
 }
 
 inline void spdlog::details::async_log_helper::set_error_handler(spdlog::log_err_handler err_handler)
--- a/include/spdlog/details/mpmc_bounded_q.h
+++ b/include/spdlog/details/mpmc_bounded_q.h
@@ -144,6 +144,13 @@ public:
         return true;
     }
 
+#if __cplusplus >= 201402L
+    [[deprecated("introduces data race")]]
+#elif defined(__clang__) || defined(__GNUC__)
+    __attribute__((deprecated("introduces data race")))
+#elif defined(_MSC_VER)
+    __declspec(deprecated("introduces data race"))
+#endif
     bool is_empty()
     {
         size_t front, front1, back;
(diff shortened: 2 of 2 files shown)


gabime:
When you think about it - there might be an easier, less intrusive way to achieve something similar:

1. push "flush" message to the queue (done in flush impl already))
2. wait until "_flush_requested" flag is false again (which means flush occurred)


Of course - this should deal with data races around the ```_flush_requested``` flag and to decide what do do with concurrent requests to flush() - but it seems like more elegant way

sam-lunt:
There could definitely be a better solution. I just modified the existing behavior to work correctly, I didn't look too closely at the other functionality.

However, while I haven't looked at the current implementation of `_flush_requested` yet, it does seem like this could still lead to unnecessarily long waits or even deadlock. If I have one thread that calls flush with `wait_for_q = true` and another thread that is frequently calling flush but with `wait_for_q = false`, then the first thread could be stuck waiting for a very long time (or forever); at the very least, it will have waited for messages added to the queue after it called flush the first time.

The flush message could have some data added to it to enable the flush call to track when _its_ flush message is pulled off, but that might not be any less intrusive than what I have here

gabime:
Maybe, just maybe, the best would be not wait at all n async mode. Just push the flush message to the queue and retunrn immdiatly. This was the old behavior, but changed to handle destruction properly.

sam-lunt:
Maybe, but that wouldn't make as much sense to me.

The synchronous version of flush blocks until all data currently available can be written to the physical file. Asynchronous mode is replacing the OS-level write buffer with an application level message buffer. Analogously, I would expect that the asynchronous version of flush would wait for all the data in that message buffer to be written to the file (but maybe not wait for the physical file to flush).

If async flush did not actually wait for the data to be pulled off, that could lead to problems if the message buffer is near full, especially if the overflow handling is to discard messages.

gabime:
In a typical scenario, one would not want to hang and wait just for flush to finish. In sync mode flush is  fast so its ok, but in asunc mode, if the q is full, it can take long time.
For the special cases, a new (non virtual) flush_q() could be added to the async logger.

sam-lunt:
In the scenario you're proposing, what would calling `flush` on an asynchronous logger do exactly? It ensures that the physical file will get flushed by the time that message is pulled off, but for an async logger that's less important.

For a sync logger, I can see two reasons why someone might call flush:

1. To ensure that all data written so far is visible in the file (e.g. so that new data shows up in `tail` or that no data is lost in case of an unexpected shutdown).
2. To ensure that a latter call to log doesn't block due to the file buffer being full (i.e. the write call would overflow the buffer, so the system needs to flush the data to file before the log call can complete). This would probably be to reduce latency.

I think that both use cases apply to the async case, perhaps more so. Also, I don't think there's an expectation that flush completes quickly, since it requires writing data to the hard disk, which is magnitudes slower than writing to memory (plus, currently, the async logger can deadlock on a flush call, so any finite time for flush is an improvement there)

gabime:
Related #147  and #145

sam-lunt:
Looking through the code it seems like each async_logger is registered with its own async_log_helper, each of which in turn starts up a thread to pull data from the queue. So unless I've misunderstood, each new async logger also starts a new background thread. This seems problematic, since that could easily lead to quite a few threads being spawned in what I wouldn't consider abnormal use patterns (e.g. each instance of an object has its own logger).

I'm wondering what the intended use pattern is for async logging. Maybe I've misunderstood something

gabime:
The main rational is simplicity, and that typical program would not have hundreds of loggers. 
That being said it is on my todo list review this again.  pull #278 is a good start (each logger has its own queue, with a coomom thread pool) .

PR #679, opened 2018-04-06
Color formatting

gabime:
Support for custom color ranges like  %^text%$ in formatter


gabime:
Now only of the log line can be colored using the %^ and %$ pattern flags

PR #683, opened 2018-04-11
Let's add a new type of logger 'step_logger' :)

Puasonych:
Hello.
Quite often I meet now systems that write logs, csv, etc. with a specified time step (fairly short) and the file is limited in volume (large enough). Also has protection from access to the log not yet completed. It consists in the fact that we first log into a temporary file, and then save it with the specified extension. This is necessary in order for monitoring systems, etc. took away the accumulated logs.

It's very sad when at work I meet several products with the same loggers, but with different implementations. And I really like this logger. I wanted to contribute. :)

Constructive criticism is welcomed.


gabime:
@Puasonych Thanks for this. Looks nice.
Few comments:
1. I've created a **contrib/sinks/**  for user contributions. Popular contribs will be moved to the main tree after stabilization phase. Please put the sink in there ("contrib/sinks/step_file_sink.h" ). 
2. Please remove the changes to spdlog.h. Will be added later if moved from contrib to main tree.
3. why the duplication of code in ```~step_file_sink()``` why not just call ```close_current_file()``` ?
4. I think adding a ".tmp" extension to the file while in step whould help. It would be much more  understandable to anyone looking at the folder while logger is active. This will also solve the problem with file names with no extension (where ```_ext``` is empty).

Puasonych:
@gabime Duplication is due to the fact that it is not good to leave the possibility of calling an exception from the destructor.

gabime:
>Duplication is due to the fact that it is not good to leave the possibility of calling an exception from the destructor.

Well you could wrap the call to ```close_current_file()``` with ```try{..}  catch(...){}```

PR #690, opened 2018-04-20
Add allocator support

Bousk:
Hi,

I was trying to use spdlog, but evolving in a tight memory industry, I had to add allocator support first.
That being done, here is my modest contribution back to you. I could try only on PC though.

- Allocators default to std::malloc & std::free so if you don't care or are already using spdlog, it makes no difference to you
- To override allocators, use 
> spdlog::allocator().allocate = xx; // or lambda [](size_t size) { return MyOwnMalloc(size); };
> spdlog::allocator().deallocate = yy; // or lambda [](void* ptr) { MyOwnFree(ptr); };

Cheers,
Cyrille // Bousk

### DIFF
--- a/include/spdlog/async_logger.h
+++ b/include/spdlog/async_logger.h
@@ -33,19 +33,19 @@ class async_logger SPDLOG_FINAL : public logger
 {
 public:
     template<class It>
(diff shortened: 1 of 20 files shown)


Bousk:
Damn, didn't see you did changes as I'm not using git as a repo.
I'll get the latest version and get back to you when everything is done.

PR #691, opened 2018-04-20
Add allocator support

Bousk:
Hi,

I was trying to use spdlog, but evolving in a tight memory industry, I had to add allocator support first.
That being done, here is my modest contribution back to you. I could try only on PC though.

Allocators default to std::malloc & std::free so if you don't care or are already using spdlog and want to keep using default allocators, it makes no difference to you
To override allocators, use
> spdlog::allocator().allocate = xx; // or lambda [](size_t size) { return MyOwnMalloc(size); };
> spdlog::allocator().deallocate = yy; // or lambda [](void* ptr) { MyOwnFree(ptr); };

Cheers,
Cyrille // Bousk


gabime:
Thanks for the pull. spdlog almost never allocates (at least not in async mode), so i am not sure where it takes effect.

Also, this pull removed the std:: prefixes everywhere, and it is hard to understand what really changed.
Please reissue with the only real changes  related to the allocator.

Bousk:
Actually there are a lot of allocations in lots of places : static strings, shared_ptr etc...
Also fmt is used with default allocator by default.
The main change can be found in common.h where the allocator is implemented and types are defined to use it instead of using basic std types with default std allocators.

gabime:
> Actually there are a lot of allocations in lots of places : static strings, shared_ptr etc...

Those are all allocated once only in initialization phase, not while the loggers do their thing. If you spot allocations in the hot path, please let me know.

Bousk:
It doesn't really matter how much allocations are done when you have strict allocation rules all you want is to control them. We have such rules because with "only" 8GB on consoles, we do want to keep track of them to control our memory budget.

gabime:
Right. I see your point. For which consoles  are you developing?

Bousk:
Right now XOne & PS4. I have troubles with exceptions now though as those are disabled. I'm thinking adding a SPD_LOG_NO_EXCEPTION in tweakme.h but it could be quiet a lot of work to make it work properly. I'm still estimating how many hours it would take.

gabime:
There have numerous [discussions](https://github.com/gabime/spdlog/pull/231) about it in the past. The conclusion was that it is really not a simple task to disable exceptions

Issue #693, opened 2018-04-25
Windows: Symbol visiblity of registry_t class.


xaqq:
Hello,

I've been using spdlog for a while and I recently had to port a project from Linux to Windows.
I've encountered various crashes that I believe come from a symbol visibility issue.

The project does the following:
  +  A shared library has a functions "init_loggers()" that create various loggers using the
      `spdlog::stdout_color_mt("root");` functions and similar.
  + A main program that links with my shared library, call `init_loggers` then (attempt to) use the loggers.

It turns out that the program will crash when using the logger after a `spdlog::get_logger()` call.

spdlog being header only, and `registry_t` class using a singleton, my program is most likely violating the One Definition Rule because both my shared library and the main program seems to get their own
definition of the `registry_t::get_instance()` function (and its internal `static registry_t instance` object). I think this then causes UB and therefore crash.

The issue seems to be fixable by using the classic BLABLA_API macro. Something like

```
#ifdef SPDLOG_EXPORTS
#define SPDLOG_API __declspec(dllexport)
#else
#define SPDLOG_API __declspec(dllimport)
#endif
```

Using the `SPDLOG_API` macro on the declaration of the `registry_t` class will fix the issue if I also compile my my shared library with `-DSPDLOG_EXPORTS=1`.

I would be interested in knowing if other people faced related issue. I'd also like feedback about the fix, especially if there are other place in the library that may need to explicitly export symbols.

Thank you,

gabime:
This is very important and a FAQ.. Any solution or help will be greatly appreciated. 

Note that any soltution  need to support both linux(gcc and  clang) and windows msvc.

gabime:
@xaqq I think that using the compiled version of spdlog will resolve this issue too. Could you confirm?

PR #695, opened 2018-04-26
Symbol visiblity

xaqq:
See #693

I tested this on a simple project with a shared library and a main program and the simple case
worked. I believe this will help with using spdlog from shared libraries.

I don't know if this is enough to fix the whole issue.
Users need to enable SPDLOG_ENABLE_DLL_IMPORT_EXPORT in tweakme.h and define SPDLOG_EXPORTS when building shared library.

### DIFF
--- a/include/spdlog/details/registry.h
+++ b/include/spdlog/details/registry.h
@@ -25,7 +25,7 @@
 namespace spdlog {
 namespace details {
 template<class Mutex>
-class registry_t
+class SPDLOG_API registry_t
 {
 public:
     registry_t<Mutex>(const registry_t<Mutex> &) = delete;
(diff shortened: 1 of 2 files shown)


gabime:
Thanks. What about gcc and clang? We must be cross platform.

xaqq:
Its my understanding that things work as expected for clang / gcc as it is.
`SPDLOG_ENABLE_DLL_IMPORT_EXPORT` is not supposed to be enabled for those compiler because, AFAIK, exporting / importing symbols is the default behavior.

So I believe nothing should be done for those compilers. However, defining `SPDLOG_ENABLE_DLL_IMPORT_EXPORT` for clang/gcc would indeed cause compilation error. What we can do is add a condition so that defining `SPDLOG_ENABLE_DLL_IMPORT_EXPORT` is noop for clang and gcc.

DasRoteSkelett:
This should be done with cmake, there are cross plattform makros for it.

gabime:
@DasRoteSkelett Can you provide more details or example?

DasRoteSkelett:
I am preparing a pull request. Might not be perfect though, especially, since I do not have Windows at all. Is visibility the issue why shared library is not working with Windows?

DasRoteSkelett:
Please see PR #1257

Issue #708, opened 2018-05-22
Upgrade to fmt 5.0


rivertam:
The glorious fmt library just [released 5.0.0](https://github.com/fmtlib/fmt/releases/tag/5.0.0).

I'd love to see a version of spdlog that utilizes the awesome new features. I'd make a PR but I don't exactly know how to manage the update.

rivertam:
Additionally/alternatively, is there a way to provide your own version of fmt to spdlog?

gabime:
You can define SPDLOG_FMT_EXTERNAL before including spdlog.h (not sure if spdlog will compile with fmt 5.0 though).

stkw0:
I used SPDLOG_FMT_EXTERNAL, but seems that fmtlib removed fmt::MemoryWriter so spdlog will need to update their code before updating to fmt 5.0

aldanor:
@gabime Is it anywhere in the plans to upgrade to fmt 5.0? (e.g. if you have a project that needs to make use of *both* fmt and spdlog)

The main thing is that `fmt::MemoryWriter` no longer exists (it's `memory_buffer` or something), so obviously it fails to compile.

arthur-tacca:
The `MemoryWriter` class was removed because it is "superceded by the compile-time format API" (as stated [the message for this commit](https://github.com/fmtlib/fmt/commit/bb738c4c264cb4e860e09160b00e5536a310c854)). It's true that there is a `memory_buffer` class that seems to mostly serve the job that the `MemoryWriter` used to, but it looks like it is intended as more of an implementation detail than a user-facing API. Assuming your format string is not known at compile time, it seems like the intended API now is just `fmt::format()`.

gabime:
@aldanor The plan is to upgrade spdlog v1.x to fmt 5.0. Release date- not sure but hopfully soon.

arthur-tacca:
Just to add to my previous comment: I was pretty much saying spdlog will have to switch from MemoryWriter to format() because you might not know the format string at compile time. But actually requiring a pattern format string to be a compile time constant is probably a sensible idea. Note that even if you choose a format string at runtime from a small list of strings known at compile time, you can still use the fmtlib compile time format routines. This is likely to be fine for the vast majority of users, and those that don't can create their own spdlog::pattern_formatter child class as a last resort.

[...]

PR #709, opened 2018-05-23
Update fmt to v5.0.0

rivertam:
I won't actually actively recommend merging this (though I certainly won't recommend *against* merging it), but it does pass the tests.

Some notes:

-  I couldn't run the benchmarks because glog wouldn't run, so I don't know how this affects performance. fmt v5 claims performance improvements, so it should improve it, but I have no numbers to back this up whatsoever.
- This does pass the mostly-unmodified tests (they were only modified in a handful of cases where the tests were using fmt directly, and I believe it was to create fixtures rather than as an actual part of spdlog), but I haven't tried to actually merge this into any real project.
- I basically took the quickest route to getting the tests to pass in most cases. This includes writing a somewhat hacky `operator<<` to avoid rewriting many lines in `pattern_formatter_impl.h`. While I don't think this will harm performance or really even readability, it's probably unideal.
- Sort of an addendum to the previous point, fmt now offers additional features which may improve performance or at least correctness or perhaps ergonomics. Chiefly, it now offers a `fmt` function which wraps templates and evaluates them **at compile-time**. I don't believe this occurs if you don't wrap them, so we're potentially losing out on significant performance on C++14 and beyond.

Hope this is helpful!

#708

### DIFF
--- a/include/spdlog/details/async_log_helper.h
+++ b/include/spdlog/details/async_log_helper.h
@@ -88,8 +88,8 @@ class async_log_helper
             msg.level = level;
             msg.time = time;
             msg.thread_id = thread_id;
-            msg.raw.clear();
-            msg.raw << txt;
+            msg.raw = fmt::memory_buffer();
+            fmt::format_to(msg.raw, txt);
             msg.msg_id = msg_id;
         }
     };
--- a/include/spdlog/details/logger_impl.h
+++ b/include/spdlog/details/logger_impl.h
@@ -64,7 +64,7 @@ inline void spdlog::logger::log(level::level_enum lvl, const char *fmt, const Ar
 #if defined(SPDLOG_FMT_PRINTF)
         fmt::printf(log_msg.raw, fmt, args...);
 #else
-        log_msg.raw.write(fmt, args...);
+        fmt::format_to(log_msg.raw, fmt, args...);
 #endif
         _sink_it(log_msg);
     }
@@ -89,7 +89,7 @@ inline void spdlog::logger::log(level::level_enum lvl, const char *msg)
     try
     {
         details::log_msg log_msg(&_name, lvl);
-        log_msg.raw << msg;
+        fmt::format_to(log_msg.raw, msg);
         _sink_it(log_msg);
     }
     catch (const std::exception &ex)
@@ -113,7 +113,7 @@ inline void spdlog::logger::log(level::level_enum lvl, const T &msg)
     try
     {
         details::log_msg log_msg(&_name, lvl);
-        log_msg.raw << msg;
+        fmt::format_to(log_msg.raw, "{}", msg);
         _sink_it(log_msg);
     }
     catch (const std::exception &ex)
--- a/include/spdlog/details/pattern_formatter_impl.h
+++ b/include/spdlog/details/pattern_formatter_impl.h
@@ -22,6 +22,13 @@
 
 namespace spdlog {
 namespace details {
+
+template <typename T>
+inline fmt::memory_buffer &operator<<(fmt::memory_buffer &w, const T &v) {
+  fmt::format_to(w, "{}", v);
+  return w;
+}
+
 class flag_formatter
 {
 public:
@@ -114,17 +121,15 @@ class B_formatter : public flag_formatter
 };
 
 // write 2 ints separated by sep with padding of 2
-static fmt::MemoryWriter &pad_n_join(fmt::MemoryWriter &w, int v1, int v2, char sep)
+static fmt::memory_buffer &pad_n_join(fmt::memory_buffer &w, int v1, int v2, char sep)
 {
-    w << fmt::pad(v1, 2, '0') << sep << fmt::pad(v2, 2, '0');
-    return w;
+    return w << fmt::format("{:02}", v1) << sep << fmt::format("{:02}", v2);
 }
 
 // write 3 ints separated by sep with padding of 2
-static fmt::MemoryWriter &pad_n_join(fmt::MemoryWriter &w, int v1, int v2, int v3, char sep)
+static fmt::memory_buffer &pad_n_join(fmt::memory_buffer &w, int v1, int v2, int v3, char sep)
 {
-    w << fmt::pad(v1, 2, '0') << sep << fmt::pad(v2, 2, '0') << sep << fmt::pad(v3, 2, '0');
-    return w;
+    return w << fmt::format("{:02}", v1) << sep << fmt::format("{:02}", v2) << sep << fmt::format("{:02}", v3);
 }
 
 // Date and time representation (Thu Aug 23 15:35:46 2014)
@@ -142,7 +147,7 @@ class C_formatter SPDLOG_FINAL : public flag_formatter
 {
     void format(details::log_msg &msg, const std::tm &tm_time) override
     {
-        msg.formatted << fmt::pad(tm_time.tm_year % 100, 2, '0');
+        msg.formatted << fmt::format("{:02}", tm_time.tm_year % 100);
     }
 };
 
@@ -169,7 +174,7 @@ class m_formatter SPDLOG_FINAL : public flag_formatter
 {
     void format(details::log_msg &msg, const std::tm &tm_time) override
     {
-        msg.formatted << fmt::pad(tm_time.tm_mon + 1, 2, '0');
+        msg.formatted << fmt::format("{:02}", tm_time.tm_mon);
     }
 };
 
@@ -178,7 +183,7 @@ class d_formatter SPDLOG_FINAL : public flag_formatter
 {
     void format(details::log_msg &msg, const std::tm &tm_time) override
     {
-        msg.formatted << fmt::pad(tm_time.tm_mday, 2, '0');
+        msg.formatted << fmt::format("{:02}", tm_time.tm_mday);
     }
 };
 
@@ -187,7 +192,7 @@ class H_formatter SPDLOG_FINAL : public flag_formatter
 {
     void format(details::log_msg &msg, const std::tm &tm_time) override
     {
-        msg.formatted << fmt::pad(tm_time.tm_hour, 2, '0');
+        msg.formatted << fmt::format("{:02}", tm_time.tm_hour);
     }
 };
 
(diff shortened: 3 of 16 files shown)


include/spdlog/details/async_log_helper.h
```diff
@@ -88,8 +88,8 @@ class async_log_helper
             msg.level = level;
             msg.time = time;
             msg.thread_id = thread_id;
-            msg.raw.clear();
-            msg.raw << txt;
+            msg.raw = fmt::memory_buffer();
```
rivertam: I don't think this will be an issue, but `memory_buffer` doesn't seem to include any destructive methods like `.clear` aside from this operator. I believe this will result in a clean deallocation and essentially nothing else.

include/spdlog/details/pattern_formatter_impl.h
```diff
@@ -114,17 +121,15 @@ class B_formatter : public flag_formatter
...
+    return w << fmt::format("{:02}", v1) << sep << fmt::format("{:02}", v2);
 }
 
 // write 3 ints separated by sep with padding of 2
-static fmt::MemoryWriter &pad_n_join(fmt::MemoryWriter &w, int v1, int v2, int v3, char sep)
+static fmt::memory_buffer &pad_n_join(fmt::memory_buffer &w, int v1, int v2, int v3, char sep)
 {
-    w << fmt::pad(v1, 2, '0') << sep << fmt::pad(v2, 2, '0') << sep << fmt::pad(v3, 2, '0');
-    return w;
+    return w << fmt::format("{:02}", v1) << sep << fmt::format("{:02}", v2) << sep << fmt::format("{:02}", v3);
```
rivertam: So fmt removed `::pad`, but `::format` should now be of equal performance (if it wasn't before), at least with precompilation.

include/spdlog/details/pattern_formatter_impl.h
```diff
@@ -461,13 +466,21 @@ class full_formatter SPDLOG_FINAL : public flag_formatter
...
 
         // Faster (albeit uglier) way to format the line (5.6 million lines/sec under 10 threads)
-        msg.formatted << '[' << static_cast<unsigned int>(tm_time.tm_year + 1900) << '-'
-                      << fmt::pad(static_cast<unsigned int>(tm_time.tm_mon + 1), 2, '0') << '-'
-                      << fmt::pad(static_cast<unsigned int>(tm_time.tm_mday), 2, '0') << ' '
-                      << fmt::pad(static_cast<unsigned int>(tm_time.tm_hour), 2, '0') << ':'
-                      << fmt::pad(static_cast<unsigned int>(tm_time.tm_min), 2, '0') << ':'
-                      << fmt::pad(static_cast<unsigned int>(tm_time.tm_sec), 2, '0') << '.'
-                      << fmt::pad(static_cast<unsigned int>(millis), 3, '0') << "] ";
+        msg.formatted << '[';
```
rivertam: I split this line up because I was getting compilation issues I didn't understand.

include/spdlog/details/pattern_formatter_impl.h
```diff
@@ -483,7 +496,7 @@ class full_formatter SPDLOG_FINAL : public flag_formatter
         msg.color_range_start = msg.formatted.size();
         msg.formatted << level::to_str(msg.level);
         msg.color_range_end = msg.formatted.size();
-        msg.formatted << "] " << fmt::StringRef(msg.raw.data(), msg.raw.size());
+        msg.formatted << "] " << fmt::string_view(msg.raw.data(), msg.raw.size());
```
rivertam: It is very much possible we could be using this throughout the codebase rather than `std::string &` or similar for added benefit.

include/spdlog/details/logger_impl.h
```diff
@@ -64,7 +64,7 @@ inline void spdlog::logger::log(level::level_enum lvl, const char *fmt, const Ar
 #if defined(SPDLOG_FMT_PRINTF)
         fmt::printf(log_msg.raw, fmt, args...);
```
rivertam: Travis is failing on this line. I didn't test with any compile flags locally, which is why I didn't catch this. I genuinely don't know what the solution is because I don't know what the distinction between `printf` and `.write` was. I suspect we can/should just remove this branch, but I don't know.

gabime:
Thanks @rivertam !
 This is very good starting point for migrating to the great fmt 5.0. 
I will probably merge fmt 5.0 to spdlog v1.x branch after some stabilization phase in both fmt 5.x and spdlog 1.x , so it might take some time..

I noticed that the example compiles with huge amount of warnings.

I compared performance and found a regression in the synchronous logging.  
The async performace on the other hand  is better with 5.0 -  so I think the cause is the ```pattern_formatter``` changes.

spdlog master branch (with fmt 4.0.1):

```
gabi@gabi-pc ~/devel/spdlog/example $ ./bench 
*******************************************************************************
Single thread, 1,000,000 iterations
*******************************************************************************
rotating_st...		Elapsed: 0.394751	2,533,240/sec
daily_st...		Elapsed: 0.387373	2,581,491/sec
null_st...		Elapsed: 0.318579	3,138,939/sec

*******************************************************************************
10 threads sharing same logger, 1,000,000 iterations
*******************************************************************************
rotating_mt...		Elapsed: 0.731929	1,366,252/sec
daily_mt...		Elapsed: 0.767222	1,303,404/sec
null_mt...		Elapsed: 0.338751	2,952,024/sec

*******************************************************************************
async logging.. 10 threads sharing same logger, 1,000,000 iterations 
*******************************************************************************
as...		Elapsed: 0.449787	2,223,274/sec
as...		Elapsed: 0.405909	2,463,608/sec
as...		Elapsed: 0.421645	2,371,661/sec
```

spdlog with  fmt 5.0 branch:
```
gabi@gabi-pc ~/devel/spdlog-fmt-5.0/example $ ./bench 
*******************************************************************************
Single thread, 1,000,000 iterations
*******************************************************************************
rotating_st...		Elapsed: 1.89739	527,040/sec
daily_st...		Elapsed: 1.904	525,211/sec
null_st...		Elapsed: 1.82554	547,784/sec

*******************************************************************************
10 threads sharing same logger, 1,000,000 iterations
*******************************************************************************
rotating_mt...		Elapsed: 0.915919	1,091,799/sec
daily_mt...		Elapsed: 0.910556	1,098,230/sec
null_mt...		Elapsed: 1.92459	519,591/sec

*******************************************************************************
async logging.. 10 threads sharing same logger, 1,000,000 iterations 
*******************************************************************************
as...		Elapsed: 0.405947	2,463,373/sec
as...		Elapsed: 0.362881	2,755,726/sec
as...		Elapsed: 0.377215	2,651,009/sec
```

include/spdlog/details/pattern_formatter_impl.h
vitaut: This is not very efficient because every call to `fmt::format` will construct a string. I suggest merging into one call `fmt::format("{1:02}{0}{2:02}{0}{3:02}", sep, v1, v2, v3)`.

Milerius:
any news ?

include/spdlog/details/pattern_formatter_impl.h
iillyyaa: Probably makes sense to remove the performance note from the comment above, since the implementation has changed. Better yet, re-time the two versions with fmt 5.0.0, pick the fastest and update the comment.

gabime: I don't think there is a way in ftm 5.0 to do it any way.
@vitaut Is there a way in fmt 5.0 achieve something similiar to (without using format strings):
```c++
fmt::memory_buffer buf;
buf << fmt::pad(static_cast<unsigned int>(tm_time.tm_min), 2, '0') << ':' <<..
```

vitaut: > Is there a way in fmt 5.0 achieve something similiar to (without using format strings)

No, the write API is gone for good so you'll have to use the format API or write to `memory_buffer` directly e.g. `buf.push_back(':')`.

gabime: @vitaut I did find a workaround which is twice as fast (I checked).
Would be happy if you could quickly review is see if reasonable (espescailly regarding the  the use of the internal ```format_int``` class):
```c++
void append_int(int n, fmt::memory_buffer &dest)
{
    fmt::format_int i(n);
    dest.append(i.data(), i.data() + i.size());
}

inline void append_and_pad2(int n, fmt::memory_buffer &dest)
{
    if (n < 0) // should not happen in this formatter, but just in case, let fmt deal with negatives correctly
    {
        fmt::format_to(dest, "{:02}", n);
        return;
    }
    if (n < 10)
    {
        dest.push_back('0');
    }
    append_int(n, dest);
}

void append_and_pad3(int n, fmt::memory_buffer &dest)
{... }
...
...
        msg.formatted.push_back('[');
        append_int(tm_time.tm_year + 1900, msg.formatted);
        msg.formatted.push_back('-');

        append_and_pad2(tm_time.tm_mon + 1, msg.formatted);
        msg.formatted.push_back('-');

        append_and_pad2(tm_time.tm_mday, msg.formatted);
        msg.formatted.push_back(' ');

        append_and_pad2(tm_time.tm_hour, msg.formatted);
        msg.formatted.push_back(':');

        append_and_pad2(tm_time.tm_min, msg.formatted);
        msg.formatted.push_back(':');

        append_and_pad2(tm_time.tm_sec, msg.formatted);
        msg.formatted.push_back('.');

        append_and_pad3(static_cast<int>(millis), msg.formatted);
        msg.formatted.push_back(']');
        msg.formatted.push_back(' ');
...

vitaut: Looks fine to me. `format_int` is experimental (and undocumented), but still part of the public API and OK to use.

rivertam:
I don't want to be seen as a jerk, but I'd just like to comment that I have neither the time nor the expertise to continue to work on this, and I hope someone can take it from here. So if anyone's waiting on me to make changes, they should rather pick up where I left off. The only reason I add this comment is so no one thinks I'm working on this when I'm not.

gabime:
Branch v1.x is already using fmt 5. To be released soon..

vitaut:
BTW I'm about to release fmt 5.1 with a bunch of bug fixes (mostly related to iterator support).

PR #719, opened 2018-06-01
Closes #717

### LINKED ISSUE #717: Compile failure on sparc Solaris 9
Compile fails on Solaris 9 as Solaris 9 does not have `strerror_r`:
```
[solaris_sparc] /home/jenkins/build/workspace/Fractions_XPlatform@2/3rdparty/spdlog_1fcc89/include/spdlog/details/os.h:430:36: error: no matching function for call to ‘errno_to_string(char [256], fmt::internal::Null<>&)’
[solaris_sparc]      return errno_to_string(buf, err);              // use overloading to select correct stringify function
```
this is technically a failure in `fmt` for not falling back to a thread-unsafe version on systems where there is no thread-safe version perhaps?

However, this can be fixed in `spdlog` by adding an inline overload for `::Null<>` - will submit a PR soon.

JamesReynolds:
This fix uses the fmt::safe_strerror instead of the underlying
strerror_r functionality. I think this is better - but I'm not
sure if there is a reason for avoiding the use of this function.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -143,11 +143,6 @@ enum class pattern_time_type
 //
 // Log exception
 //
-namespace details {
-namespace os {
-std::string errno_str(int err_num);
-}
-} // namespace details
 class spdlog_ex : public std::exception
 {
 public:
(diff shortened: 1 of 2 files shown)


JamesReynolds:
As per comments on #717 I've moved the code all into common.h. Are there any standards that I should follow that I've missed? I've also not rebased to squash this into a single commit - happy to do that if it helps the history.

gabime:
Thanks @JamesReynolds

gabime:
@JamesReynolds I commited some minor code simplification in common.h around the use of ```fmt::safe_strerror```

PR #727, opened 2018-06-13
add remove_all_sinks method to dist_sink

jwnimmer-tri:
This allows users to set exactly the sinks they want, even if other unknown application code has added bespoke sinks in the meantime.


jwnimmer-tri:
Hi @gabime -- when looking to upgrade the tag of spdlog that I'm using, it seems like this commit has gone missing -- maybe there was a force push to master that removed it?  I can't see it on https://github.com/gabime/spdlog/blob/master/include/spdlog/sinks/dist_sink.h anymore.  Maybe I'm looking at the wrong branch or something, but I think I've got it right.  Any ideas?

gabime:
Yes. Could be due to force push I did. Care to commit again?

jwnimmer-tri:
=> #735.  Thanks!

PR #753, opened 2018-07-14
Add logger name for macro "SPDLOG_CATCH_AND_HANDLE"

yhchen:
1.add logger name for exception handler macro "SPDLOG_CATCH_AND_HANDLE"
2.change exception handler to macro "SPDLOG_CATCH_AND_HANDLE" in function "spdlog::async_logger::_sink_it"

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -176,13 +176,13 @@ using filename_t = std::wstring;
 using filename_t = std::string;
 #endif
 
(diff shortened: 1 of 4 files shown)


gabime:
The handler is used in the loggers, no need to pass the name. It is already an accesible member. It is enough just to append the _name.

gabime:
> The handler is used in both logger and async_log_helper. the member "name" are not the same("name_" and "_logger_name")

Sorry,  I was talking about the v1.x branch.. - no async_log_helper anymore..

PR #789, opened 2018-08-11
Modify travis-ci, fix compilation issues

### LINKED ISSUE #783: Cannot find source file: bench.cpp
in examples CMakeLists.txt was mentioned bench.cpp , but it is not in the folder.

DanielChabrowski:
Hey,
I got some time and cleaned up travis-ci, as you can see thread sanitizer detected *some* issues, address sanitizer also reported a leak.

Current issues:
* Always building with -O3 (is that really an issue though?)
* Always building with c++11
* Valgrind is run on examples only
* ASAN and TSAN are not run at all

Possible improvements:
* Move tsan/asan handling to CMake

Added:
* Release/Debug job for gcc 7 and clang 3.5
* Debug asan gcc 7 job
* Debug tsan gcc 7 job
* Disabled extensions
* Added a spdlog::spdlog alias for tests

Removed:
* Valgrind workaround, the current version is 3.10.1
* install_libcxx
* Makefiles

Fixed:
* examples build ( fixes #783 )
* multisink build

Workarounds:
* gcc7 with tsan and asan needs gold linker, otherwise linking
  fails becase of unrecognized option '--push-state'


DanielChabrowski:
So apparently tsan doesn't work well with lock-free. Don't know about asan. It may be worth to investigate both though. Let me know if I should remove the asan/tsan jobs.

gabime:
@DanielChabrowski Thanks for the pull. 
Indeed, I'll have  to investigate the asan/tsan errors before merging. I will update here my findings.

gabime:
I investigated. Seems like false positives by gcc (there are numerous reports about similiar false positives - for example [here](https://stackoverflow.com/q/28719796)):

I run tests on the tests exe:
- no errors reported if complied with clang++  (whether compiled with or without -O2
- no errors reported if complied with gcc  - but with -O2
- no errors reported by valgrind

So I think it is better to remove those options - they just cause confusion

Also, can you please put back the removed tests/makefile ? I know it is redundant, but I like having it around when cmake is not used/available..

DanielChabrowski:
I moved handling of asan/tsan flags to cmake since that allowed me to easily run these builds locally. "sudo: required" fixed the asan job that was failing because of "Could not attach to thread" error.

I read your message while writing this one - how about running tsan/asan jobs with clang then? I will bright back the Makefile in a second.

gabime:
ok, lets try with clang

DanielChabrowski:
"sudo required" basically means the jobs will be executed in a VM on Google Cloud Engine as opposed to a container. The issue is that asan requires some capsys privileges that containers are probably not run with, as a security concern I think.

https://docs.travis-ci.com/user/reference/overview/

gabime:
Still errors.. valgrind, tasn, etc..

DanielChabrowski:
Changing the test_sink.h:
`
    size_t msg_counter_{0};
    size_t flush_counter_{0};
`
to:
`
    std::atomic<size_t> msg_counter_{0};
    std::atomic<size_t> flush_counter_{0};
`

Fixes the tsan issue.

gabime:
hmm.. TSAN fails on my machine too (clang 6.0.0), while ASAN passes..

gabime:
>Changing the test_sink.h:
size_t msg_counter_{0}; size_t flush_counter_{0};
to:
std::atomic<size_t> msg_counter_{0}; std::atomic<size_t> flush_counter_{0};

Fixes the tsan issue.

That's problematic finding... They should have been protected by std::mutex

DanielChabrowski:
Adding std::atomic breaks gcc 4.8 builds though. :(

https://travis-ci.org/DanielChabrowski/spdlog/builds/415160620

gabime:
> Adding std::atomic breaks gcc 4.8 builds though. :(

An in anyway, I don't understand why it solves it..

DanielChabrowski:
Then I propose two solutions:
1) Allow tsan job to fail so that we can always check if something fixes it. In that case the result of tsan job won't affect the commit status.
2) Drop the tsan job. We could still leave cmake configuration though.

Up to you.

gabime:
I suggest removing the TASN from the tests. Seems like yet another false positive (at least I hope so

gabime:
Also, while you are at it, could you please remove from the matrix gcc 4.9 and clang 3.5 ? There are too many builds configurations.

DanielChabrowski:
I don't mind gcc 4.9 but I would leave clang 3.5 as the README claims that spdlog should also work with it. We can either have issues reported or CI automatically checking that. I would opt for a second option. Range-v3 has 27 configurations so I wouldn't call 10 too many.

It's really annoying that travis-ci fails so often - have you considered gitlab-ci as an alternative?

DanielChabrowski:
I have yet to setup a gitlab-ci with github but it doesn't seem hard.
https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html

It's been a while since I played with it but I don't remember it failing because of the issues that travis currently has. If you setup a dummy CI that will already be connected to gitlab I can write the gitlab-ci yaml file. Then we can compare.

gabime:
Travis is not  bad actually - it gets the job done usually. But if problems continues, i will sure try gitlab-ci

DanielChabrowski:
@gabime Great! You also removed valgrind from all builds - was that intended? I would also add ```export TSAN_OPTIONS=verbosity=1``` as it gives more information about why tsan failed.

Looking at commits, out of last 12 builds, 6 failed because of the apt failure. Not very stable...

gabime:
> You also removed valgrind from all builds - was that intended?

No, but maybe if asan tsan are enable, it is not needed anyway

DanielChabrowski:
@gabime Valgrind also detects uninitialized memory reads, which asan does not. Msan on the other hand will but as it says on wiki msan only "implements a subset of functionality found in Valgrind".

PR #791, opened 2018-08-12
Add gitlab-ci.yml

DanielChabrowski:
As discussed in #789 it may be worth checking out gitlab-ci instead of travis-ci.

https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html#connect-with-github-integration
I enabled it on my branch, it took like 3 clicks to do it.

Notes:
- clang-3.5 package is not available in ubuntu:bionic
- gcc-7 upgraded to gcc-8
- The build takes 4-5 minutes but could potentially be faster with premade docker images.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,94 @@
+image: ubuntu:bionic
+
+stages:
+  - build
+
+variables:
+  CPP: 11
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @DanielChabrowski 
I will consider this. I am leaving this PR open for now..

PR #793, opened 2018-08-13
Replace direct usage of std::mutex,std::lock,std::thread,std::chrono …

gajanak:
Replace direct usage of std::mutex,std::lock,std::thread,std::chrono by  'imported' namespaces. By importing this in common.h, it is now possible to use  boost::thread and boost::chrono as replacement of std::thread/mutex/chrono.

Current MinGW compiler comes with mutex headers, but without and implementation. and boost::thread only works with boost::chrono.

Now there is a central place to decided which implementation is used, kind of abstraction for lazy.

Additional this commit removes all double includes, which is already included in common.h and os.h -> which is currently included in all paths. This pretend including useless mutex/chrono headers and speedup compile time a little bit.

This is a really fast simple made change - and can be improved.
The changes are only made to the header only libs.  Change to the testsuite and examples are separated.

PS: this is my first PR in GitHub -> hope it works ;)


gabime:
Thanks for the PR. 
While it might be useful for some specific cases, I dont think it worth the big change and the additional abstraction layer.  I try to stick with simple and standard c++11 code.

PR #799, opened 2018-08-16
Locking vector in dist_sink::sink_it_ and dist_sink::flush_

surkamp:
`dist_sink` needs locking to read the `sinks_` vector because another thread may `add_sink()` or `remove_sink()` while `sink_it_()` or `flush_()` is running. Changing the container while another thread is reading may lead to container corruption.

```
Data races
 If a reallocation happens, all contained elements are modified.
```
http://www.cplusplus.com/reference/vector/vector/push_back/


gabime:
sink_it_ and flush_ are already locked when called from the base public functions.

PR #806, opened 2018-08-23
WIN32: Don't deny access to file sinks from other processes

emruiz81:
Hello! I don't know why is the file open mode set to _SH_DENYWR in Windows, but it prevents to read the file when it is being used by the logger.
I have just changed the mode to _SH_DENYNO and it seams to work perfectly in all my production environments.


gabime:
> I don't know why is the file open mode set to _SH_DENYWR in Windows

To pevent writing by any other process. I guess it is problematic with some editors., so I dont mind the fix.

mmakhalaf:
We used to rely on the behaviour `_SH_DENYWR` to stop multiple file sinks from opening the same log and handled the exception to create/open a different log file. Is there any way this could be a global setting / flag?

PR #813, opened 2018-08-29
Add set_sinks method to dist_sink

jwnimmer-tri:
This allows users to set exactly the sinks they want, even if other unknown application code has added bespoke sinks in the meantime.

(This is a re-submission of #735, which was merged against the master (v0.x) branch.)


gabime:
Maybe even better would be to change to “set_sinks()” ? (which would remove and add atomically the new sinks).

jwnimmer-tri:
Ah, so that we'd be guaranteed not to drop any messages.  Sure, I'll rework it.

jwnimmer-tri:
@gabime What do you think of this PR now?

gabime:
Thanks @jwnimmer-tri

PR #814, opened 2018-08-29
Pessimizing move removed

maciekgajewski:
An instance of pessimizing move removed.

It actually makes code less performing, as is prevents copy elision.
It was also causing compilation error with -Weverything enabled in clang.


maciekgajewski:
It fails to compile only because I'm compiling my project with -Werror, so the warning is treated as error.

PR #823, opened 2018-09-06
add argument to rotating file sink for rotate_on_open

pwm1234:
when true, the log file will be rotated when it is opened so the newly constructed file will start off being empty


gabime:
I think a “truncate” param is what you want. like in [basic_file_sink](
https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/sinks/basic_file_sink.h).

pwm1234:
First, let me say THANK YOU for your work on spdlog and making it available!

Second, I did not know about the "truncate" param in basic_file_sink. But since truncate means to shorten or replace, which is different than "rotating," do you really want the argument to be named truncate? I am happy to do so, I just wanted to double-check.

Third, is this a pull request that you would accept once we iron out the details? If so, do you want me to make it using HEAD or am I okay at the 1.0 tag? (I chose this because that is the tag that vcpkg is using.)

pwm1234:
> I was referring to the actual implementation. It depends the purpose - if the purpose is is just starting with empty file, then maybe truncating is good enough.. If the purpose is to rotate first, and lose an old file, then "rotate_on_open" is fine.

> Either way, the program is going to lose a log file..

Yes, the purpose and implementation is to rotate first. An old one will be lost only if the maximum number has already been reached.

I will update the pull request to use the HEAD of the 1.x branch.

pwm1234:
Just checking on this pull request. Is there something I can do to get this change merged?

gabime:
@pwm1234  Please open a new pr. I see there are conflicts

PR #831, opened 2018-09-12
Change the default value for SPDLOG_BUILD_xxx

### LINKED ISSUE #830: Change default value for SPDLOG_BUILD_TESTING
Consider the following usage scenario: _spdlog_ is added as a Git submodule and then in the project CMakeLists included using: 

```cmake
add_subdirectory(third-party/spdlog)
```

This will result in _spdlog_ test targets being created because `SPDLOG_BUILD_TESTING` defaults to `ON`. One can force-reset this variable after the `add_subdirectory` call, but this is not convenient.

I propose to chose the default value based on whether _spdlog_ is used as a third-party dependency or as a standalone project. In other words, if _spdlog_ is included through `add_subdirectory` as in my scenario, the tests are disabled by default, and if _spdlog_ is configured standalone, then the tests are enabled by default (like they are now).

Some other projects follow this approach, for example [_pybind11_](https://github.com/pybind/pybind11/blob/ef13fb2e1c89f3e67f83eb860e8f6f37954a8cb1/CMakeLists.txt#L22-L29). Here is how they do this:

```cmake
# Check if pybind11 is being used directly or via add_subdirectory
set(PYBIND11_MASTER_PROJECT OFF)
if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
  set(PYBIND11_MASTER_PROJECT ON)
endif()


option(PYBIND11_INSTALL "Install pybind11 header files?" ${PYBIND11_MASTER_PROJECT})
option(PYBIND11_TEST    "Build pybind11 test suite?"     ${PYBIND11_MASTER_PROJECT})
```

I can submit a PR if you support such a change.

taketwo:
The value is based on whether _spdlog_ is used as a third-party dependency or as a standalone project. If _spdlog_ is included through `add_subdirectory`, the tests/examples/benchmarks are disabled by default, and if _spdlog_ is configured standalone, then the tests are enabled by default.

Closes #830. (I did not talk about examples/benchmarks in that issue, however the logic is the same, these are not needed when _spdlog_ is a third-party dependency.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,11 +45,17 @@ include(cmake/sanitizers.cmake)
 add_library(spdlog INTERFACE)
 add_library(spdlog::spdlog ALIAS spdlog)
 
-option(SPDLOG_BUILD_EXAMPLES "Build examples" ON)
-option(SPDLOG_BUILD_BENCH "Build benchmarks" ON)
+# Check if spdlog is being used directly or via add_subdirectory
+set(SPDLOG_MASTER_PROJECT OFF)
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+  set(SPDLOG_MASTER_PROJECT ON)
+endif()
+
+option(SPDLOG_BUILD_EXAMPLES "Build examples" ${SPDLOG_MASTER_PROJECT})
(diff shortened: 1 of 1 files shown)


gabime:
Merged. Thanks @taketwo

PR #836, opened 2018-09-17
handle failure of file removal

rafiw:


gabime:
The problem is that it will probably fail(line 93 already removes it).

This code was added as a wraparound to a weird issue in windows +antivirus under heavy rotation rates (many rotations per second).

It just for retries to remove and rename again, even though was already removed. Maybe the correct fix would be
```c++
 if (details::file_helper::file_exists(target)) && details::os::remove(target) != 0)
{
    throw(..)
} 
```

gabime:
Fixed in 2fa7410c0ecf61c4270b9aa562380a6be4c25280

PR #837, opened 2018-09-17
ansicolor_sink.h - add missing sink include

myrgy:
Hello!

Looks like this include was missed!

Thanks!


gabime:
Merged. Thanks @myrgy

PR #862, opened 2018-10-05
Develop daily rotating file sink

Asperger:
A sink combined with daily_file_sink and rotating_file_sink, which creates new file each day and also when the file size exceeds.
Move `daily_filename_calculator` into sink to avoid conflict with `daily_file_sink`.
In addition, no limit of the amount of file.

### DIFF
--- a/include/spdlog/details/file_helper.h
+++ b/include/spdlog/details/file_helper.h
@@ -22,50 +22,6 @@
 namespace spdlog {
 namespace details {
 
-/*
- * Generator of log file names
- */
-struct filename_calculator
-{
-    // Create filename for the form basename.YYYY-MM-DD
-    static filename_t calc_filename(const filename_t &filename, const tm &now_tm)
-    {
-        filename_t basename, ext;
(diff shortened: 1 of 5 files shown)


gabime:
Thanks for the PR.
Few comments: 
1. Max files is important feature imho. Need it back
2. There are too mant args to the ctor. I think its time to mve them to a struct (“daily_rotating_config”).
3.  There is code duplication with the rotating and daily sinks. Lets consider how this duplication can be reduced.

Asperger:
> Just pass `std::numeric_limits<std::size_t>::max()` to max_files
Indeed. I will bring them back.
> About the "truncate" param, It would better if it decided about rotation on the construction (see PR #823).
Sorry. I don't quite get it. Would you mind explain a bit further?
> We should use `std::chrono::hours`, `std::chrono::minutes`.
I will work on them and also other issues later on.

Asperger:
> Just pass std::numeric_limits<std::size_t>::max() to max_files

Still got a problem when I do this. `rotate_()` will literally iterate 18446744073709551615 times to rename file. Maybe we should check the amount of current copies to prevent those kind of wasting.

PR #864, opened 2018-10-10
fix namespace of make_unique

DimRochette:


gabime:
Thanks @DimRochette,.

PR #868, opened 2018-10-13
Add a high time resolution tweak with default pattern

future0906:
Hi gabime,
I added a simple tweak to make default pattern with higher timer resolution, which in some cases is needed.

Sure we can use custom formatter or pattern to archive this, but we will lost the ability of full formatter's second cache, which might be lost some performance.

Hope this help.


gabime:
Thanks. Sounds reasonable, but I need to think twice before adding yet another ifdef to the code. I will put this on hold, and if there is big demand, I will merge.

In the meanwhile, if yoy want to use default pattern and see micros , you could set the pattern to “%+ [%f]”

PR #876, opened 2018-10-18
Use fmt::string_view when logging a C string to avoid unnecessary copy

peergynt:
This PR proposes an optimization when logging C strings.

The existing implementation copies the null terminated C string into the raw buffer.
The changes proposed add a fmt::string_view to log_msg which directly points to the C string.
Whenever the string_view is populated, it shall be used instead of the raw buffer as the actual message.
Another logging method was also added for when the client already knows the size of the C string or if the C string is not null terminated.

I also added a test case and extended the benchmark.
Running the benchmark shows noticeable improvements when logging large buffers.

### DIFF
--- a/bench/bench.cpp
+++ b/bench/bench.cpp
@@ -29,6 +29,7 @@ using namespace utils;
 void bench(int howmany, std::shared_ptr<spdlog::logger> log);
 void bench_mt(int howmany, std::shared_ptr<spdlog::logger> log, int thread_count);
 void bench_default_api(int howmany, std::shared_ptr<spdlog::logger> log);
+void bench_c_string(int howmany, std::shared_ptr<spdlog::logger> log);
 
 int main(int argc, char *argv[])
 {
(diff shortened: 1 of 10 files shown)


gabime:
@peergynt Thanks for the PR. 
It's does save about 10 nanoseconds per each 100 bytes on my machine. When exceeding  500 bytes, the  improvement is much bigger since  memory_buffer allocates when it exceeds 500 bytes.

However it's a major change (and a breaking change - any 3rd party code that rely directly on ```log_msg.raw``` is now broken).
I've created a new [branch](https://github.com/gabime/spdlog/tree/peergynt-sized_buf) for this and commited numerous changes.

The most important one is that  ```log_msg``` contains a string_view, instead of memory_buffer, which works nicely both with c_str and memory_buf.

I suggest  proceed with this branch until It feels ready for merge - there are lots of changes.

peergynt:
@gabime thanks for the pull and the additional cleanup. Using a "string_view" in the log_msg looks much cleaner imho.
Btw, I think that you can remove the log_msg.h include from fmt_helper.h now.

PR #889, opened 2018-10-30
change the logger's sink_it_ parameter to rvalue reference

rwen2012:
There seems not safe for moving (std::move) a lvalue reference,  change the parameter to rvalue should be better.

=======================
inline void spdlog::async_logger::sink_it_(details::log_msg &msg)
{
#if defined(SPDLOG_ENABLE_MESSAGE_COUNTER)
    incr_msg_counter_(msg);
#endif
    if (auto pool_ptr = thread_pool_.lock())
    {
        pool_ptr->post_log(shared_from_this(), std::move(msg), overflow_policy_);
    }
    else
    {
        throw spdlog_ex("async log: thread pool doesn't exist anymore");
    }
}

### DIFF
--- a/include/spdlog/async_logger.h
+++ b/include/spdlog/async_logger.h
@@ -58,7 +58,7 @@ public:
     std::shared_ptr<logger> clone(std::string new_name) override;
 
 protected:
-    void sink_it_(details::log_msg &msg) override;
+    void sink_it_(details::log_msg &&msg) override;
     void flush_() override;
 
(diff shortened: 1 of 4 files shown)


gabime:
Thanks for noticing. 
I think a simpler approach is just to chage the post_log and the fubctions in thread_pool to accept log_msg& instead of log_msg&&. There is no real move anyway in log_msg , so moving is not needed and just confusing imo.

rwen2012:
thanks,  
change to log_msg& is more clear.

PR #892, opened 2018-11-04
Global option that disables global registration of loggers

### LINKED ISSUE #712: Create logger without automatic registration
It is treacherous that if one wants to create loggers on the spot or have loggers with the same name, they have to do manually everything that `create` does (choose async or not, set level, flush policy, etc). Same things for many other functions whose documentation starts with "Create and register"...

Personally I do not foresee a case when I will need to register my loggers, so automatic registration is just an impediment. 

Some proposition for backward compatible changes:
- a global flag (`set_automatic_registration`) to disable registration in convenience functions
- an optional argument added to `create` (and maybe every other function automatically registering)
- extra functions which would not register, naming is tricky, `create_local` ? `create_noregister`?
- have the constructor of `logger`lookup the different global variables like `create` does (most intuitive, but casting for `async_logger` would be an issue)
... ?

pabloariasal:
- fixes #712 
- If option is set, all loggers (both sync and async) created via `create()` won't be added to the registry
- if option has been set, loggers don't need to have unique names
- loggers will be initialized according to the global settings
- given that loggers are not global, further changes to global settings, e.g. via `set_formatter` won't apply to existing local loggers


gabime:
Thanks for the PR !.
Useful feature indeed, however I was thinking about implementing this using some kind of API instead of macro. 
Not sure exactly, but I think @leolchat proposal of having ```set_automatic_registration(bool)``` is good direction.

pabloariasal:
I made a revision of my changes and replaced the macro via a API call, `set_automatic_registration(bool)`. I also added a unit test.

include/spdlog/details/registry.h
```diff
@@ -269,6 +271,7 @@ class registry
     std::shared_ptr<thread_pool> tp_;
     std::unique_ptr<periodic_worker> periodic_flusher_;
     std::shared_ptr<logger> default_logger_;
+    bool automatic_registration_ = true;
```
gabime: must be atomic<bool>

pabloariasal:
Thank you for your comments! Yes, was indeed missing the mutexes

pabloariasal:
Oops yeah that was not the smartest :sweat_smile:

gabime:
Thanks. Merged

leolchat:
Nice, thank you @pabloariasal

Issue #897, opened 2018-11-06
pattern_formatter is not thread-safe; should it be?


aschran:
spdlog's `pattern_formatter` is not thread-safe due to its use of cached pre-computed values that are not protected by a mutex or thread-local.

For example, see `last_log_secs_` and `cached_tm_` here:
https://github.com/gabime/spdlog/blob/v1.2.1/include/spdlog/details/pattern_formatter.h#L559

Was this class intended to be thread-safe? (Or, more broadly, are formatters in spdlog supposed to be thread-safe?) If so, I am happy to send a PR for a fix based on either mutexes or making the cached value thread-local.

gabime:
They should not be thread safe. They are owned and protected by their users (thread safe sinks).

gabime:
Exactly. This also prevents un needed locking when non thread safe sinks are used (the dont lock it).

aschran:
Though formatters are not thread-safe by design right now, would you be amenable to a change like the one I suggested to make them so? Using thread_local preserves the property that no locking is required and should be essentially no change for a single-threaded program, while still making it possible for thread-safe sinks to not lock every call to a formatter. (Of course, if you don't want something like that in spdlog that's certainly your prerogative and then I guess we'd add it locally to the project where we want to use it.)

gabime:
No, sorry. I dont see the point (and it wont even prevent the sinks from locking, since they lock anyway to protect their destination).

aschran:
Many sinks don't otherwise need a lock because the destination is itself already thread-safe. For example, maybe we're appending logs to a file and the file append is atomic. Or maybe we're outputting logs via a network RPC and the RPC library is thread-safe.

So, that would be the reason. But it's no big deal to fork and implement our own thread-safe formatter.

gabime:
Well, in those cases the sink is free to only lock around the formatter . Pattern formatter is being actively developed and improved so forking is not best choice imo.

PR #898, opened 2018-11-06
pattern_formatter: use thread_local for cached values to make the class thread-safe

aschran:
Proposed fix for https://github.com/gabime/spdlog/issues/897.

This PR is not complete. Let me know if you are okay with this overall approach and I can fix the other instances of thread-unsafe cached values in this file?


gabime:
Thanks for the PR,  but there is no real issue.

PR #903, opened 2018-11-12
Ability to run periodic operation in logger thread

levishai:
Hi Gabi,

I think that the user ability to code and run periodic operation in the context of the logger thread is pretty useful and this is what we're introducing in this change.

It started with the will to flush periodically without creating a thread to do it (flush_every creates such thread) but then we decided to make it more generic.

By inheriting/implementing the async_logging_periodic_worker interface the user can define in do_periodic_work what should be done periodically in the logger thread/s context.

In case no periodic_worker provided via the thread_pool constructor then there will be no effect to our change and the wait on the queue will be 10 seconds as it was coded originally,

Thanks,
Shai


gabime:
Thanks for the PR. Why do you need to run periodic in same thread as the logger ?  What are the use cases?

levishai:
> Thanks for the PR. Why do you need to run periodic in same thread as the logger ? What are the use cases?

For example, when we want to flush periodically without creating specific thread for periodic flushing (flush_every() in spdlog create separate thread for it)... Or for checking periodically if there was a change in our category2priority config file (we use this file to map every category to its logging prioirty and the user can modify it)...

gabime:
Are you running in somekind of embedded? Creating single additional thread fir flushing for is not expensive imo.

PR #911, opened 2018-11-18
For Windows, make sure the file handle created by os::fopen_s() is never inherited by child processes launched using CreateProcess().

PCMan:
By default, the file handle under FILE is inheritable by child processes in Windows. If the current process creates a child process using CreateProcess() API, and set inherit handles to TRUE, then the log file of spdlog will be kept open by the child process and is never closed.
In this case, log file rotation will always fail since you cannot do rename() because another process is still using the file.

### DIFF
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -153,6 +153,13 @@ inline bool fopen_s(FILE **fp, const filename_t &filename, const filename_t &mod
 #else
     *fp = _fsopen((filename.c_str()), mode.c_str(), _SH_DENYNO);
 #endif
+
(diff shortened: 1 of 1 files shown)


gabime:
Doenst the. ode already doezthis in line 167?

#ifdef SPDLOG_PREVENT_CHILD_FD
    if (*fp != nullptr)
    {
        prevent_child_fd(*fp);
    }

PCMan:
Oh! I didn't see that `#ifdef SPDLOG_PREVENT_CHILD_FD`. Thanks for the info. Withdrawl the PR.

PR #913, opened 2018-11-18
fix: addition standard equivalent of "pragma once"

Apitronix:
Replacement of "pragma once" by its standard equivalent, to insure the compilation of the code by all compilers.


gabime:
which compilers dont support this?

gabime:
Because its a header only lib, I am a bit reluctant to add so many macros - they will all be added  to the  user’s environment.

gabime:
The problem is spdlog already defines  ~20 macros (try auto-complete in any ide for "SPDLOG_ " and it will give you list of the relevant macros).

 I don't want to add  40 more - just for a compiler that not even supported by spdlog.

Issue #638, opened 2018-02-16
SPDLOG_TRACE and _DEBUG evaluate their arguments even when logging is off


sherm1:
When these macros are expanded (because NDEBUG is _not_ defined), they evaluate their arguments unconditionally, _before_ checking the logging level. When arguments are expensive to evaluate (common in code intended for debugging output), that can be very costly.

Please consider having these macros perform a runtime check of the level prior to invoking logger->trace() or logger->debug(). That is a very small extra cost that avoids evaluating the arguments. For example:
```c++
#ifdef SPDLOG_DEBUG_ON
  #define SPDLOG_DEBUG(logger, ...) \
  do {                                             \
    if (logger->level() <= spdlog::level::debug) { \
      SPDLOG_DEBUG(logger, __VA_ARGS__);           \
    }                                              \
  } while (0)
#else
  #define SPDLOG_DEBUG(logger, ...)
#endif
```

gabime:
> because NDEBUG is not defined

NDEBUG is not used. Spdlog expands the macros only if SPDLOG_DEBUG_ON is defined..

sherm1:
Thanks for that clarification! Would be great not to evaluate the arguments when SPDLOG_DEBUG_ON is defined but logging is off.

gabime:
I agree. It might be worth the extra cost.
There is a handy ```logger->should_log(level)``` that can be used for this.

chrahunt:
When/if this is implemented, heed the lessons learned [here](https://github.com/wxWidgets/wxWidgets/blob/20a98406f06f0392f0b303c0deb6c630e17704fd/include/wx/log.h#L1271).

arthur-tacca:
Just a note that spdlog only formats a log message if it is going to be logged by the current logger. This means that a workaround (or arguably even correct solution) is to use a struct or class with an `operator<<()`:

    // bad: getDescriptionString() is called even if not logged, even if we use SPDLOG_DEBUG instead
    mylogger->debug("Starting {}", getDescriptionString(foo));
    // good: Foo::operator<< only called if it will be logged
    mylogger->debug("Starting {}", foo);

If you don't want to add an `operator<<` to the thing you want to log, or you want to log the result of an expensive function rather than an object, you can always create a little wrapper struct for the sole purpose of having an insertion operator:

    struct FooPrinter {
        explicit FooPrinter(const Foo& foo): foo(foo) {}
        const Foo& foo;
[...]

PR #915, opened 2018-11-19
Conditional evaluation of macro arguments

DanielChabrowski:
Implements #638 

Hey,
This PR introduces the lazy evaluation of macro arguments in cases where logging won't even happen. The number of macros increased since the issue was created so I kept the behaviour constant across all of them.

I changed `err` to `error` as a convenience since logging functions are called `error`, but it's not necessary, additional parameter to `*_GENERIC` macros can be added.

This unfortunately breaks the code that somehow relies on evaluation in macros but the exact same thing would happen if `SPDLOG_ACTIVE_LEVEL` was changed.

Let me know if this enhancement is still desired.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -96,7 +96,7 @@ enum level_enum
     debug = SPDLOG_LEVEL_DEBUG,
     info = SPDLOG_LEVEL_INFO,
     warn = SPDLOG_LEVEL_WARN,
-    err = SPDLOG_LEVEL_ERROR,
+    error = SPDLOG_LEVEL_ERROR,
     critical = SPDLOG_LEVEL_CRITICAL,
     off = SPDLOG_LEVEL_OFF,
 };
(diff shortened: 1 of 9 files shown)


DanielChabrowski:
Looks like 7068c45 broke the osx build.

gabime:
Thanks @DanielChabrowski . 
I think this is redundant with the new way that SPDLG_ACTIVE_LEVEL is implemented. spdlog  [initialize](
https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/details/logger_impl.h#L332) the level of all loggers to the value of this macro. This means that most of the time if a logging macro is enabled, then the logger is also enabled, and vise versa. There mght be exceptions of course if the logger’s level is set at runtime, but for most of cases, I dont think it worth the exrtra cost(another if).

DanielChabrowski:
Sure, then it's in the history if anyone needs it. The issue can be closed as well?

gabime:
> The issue can be closed as well?

Yes. I think so.

PR #921, opened 2018-11-20
refactor:Split main example into several expamples

Apitronix:
Split the file named "example.cpp" into several examples for
"console / basic / daily / rotating logs" to increase the code
readability, and add a directory to store binaries,
to clean up the "example" directory.


gabime:
Thanks for the PR. The test fails..Also the cmake  must be updated.
Anyway, this is just a quick and dirty example folder. I don't think it worth having all those changes.

Apitronix:
Yes, but I don't know why. The binaries path seems to be wrong, I
mean...
Anyway, this will be useful for all developers who want to use this
library, so if my work is not good, I think it would be nice if we
could have something like that.

Le mardi 20 novembre 2018 à 09:22 -0800, Gabi Melman a écrit :

PR #926, opened 2018-11-27
Fix exceptions on file size calculation on Windows XP x64 and Windows Server 2003 x64

lestera:
For some reason, `_fstat64` fails with errno 22 on Windows Server 2003 x64 (and probably Windows XP x64 as well) when compiled using the `v141_xp` toolset.
Using `_filelengthi64` instead solves this issue.


gabime:
Thanks !

PR #936, opened 2018-12-06
Corrected the text alignment in the example

bzindovic:
I've changed the alignment character in the example to illustrate left alignment of text. Originally, the alignment was right.


gabime:
Thanks. Merged.

PR #945, opened 2018-12-17
Fix handling of external fmt lib

cneumann:
Using an external fmt lib should cause the spdlog::spdlog target to have
a dependency on fmt lib - so that a consuming project does not need
to call find_package(fmt) and target_link_libraries(... fmt::fmt).

To this end a new cmake option SPDLOG_FMT_EXTERNAL is introduced which
makes spdlog depend on fmt lib and defines the SPDLOG_FMT_EXTERNAL macro
to avoid using the bundled fmt lib. The value of SPDLOG_FMT_EXTERNAL is
also stored in the installed spdlogConfig.cmake and if it is ON
find_dependency() is used to ensure the fmt::fmt target is imported.

This was discovered in the context of vcpkg which always uses external fmt lib, but required consuming projects to handle the dependency - see Microsoft/vcpkg#4994


gabime:
Thanks.Merged.

PR #956, opened 2019-01-08
Allow filename/line number at all levels. Add function name %! support.

possiblyhuman:
I extended the changes (that allowed the use of filename/line numbers) done to the SPDLOG_TRACE macro to the same macros for all the other logging levels (SPDLOG_CRITICAL, SPDLOG_INFO etc).

I also extended the filename/line number macro stuff to also be able to use function names, and added some extra defines in tweakme.h for customisation.

I picked %! for writing the function name. I don't know if that would be a good choice or not.


gabime:
Thanks for the pull. Not sure I want each of those macros emit the source info - seems it mostly needed in trace. 
Also, let's avoid defining too many macros in the spdlog.h. After all, this is a header only lib, and all those defines end up everywhere (The SPDLOG_LOG_MACRO_.. are redundant and can be removed with use of direct calls like current version)

possiblyhuman:
In hindsight, I agree.

I removed the commit with the excessive use of SPDLOG_LOG_MACRO_... macros. I thought I was going to need the option to selectively choose when to have the features, but after using the library some more and thinking about, I won't.

We could also get rid of the SPDLOG_SOURCE_MACROS_ON and SPDLOG_LOG_MACRO macros that remain. I'll do this if you think the other changes are actually worth adding.

It's definitely useful to me to have filename, line number and function name available at all levels. Otherwise I'd have to add SPDLOG_TRACE in places where I already have SPDLOG_DEBUG, SPDLOG_INFO etc. I guess that's kinda the point of TRACE, but having it available to all just seems simpler and easier to me.

With spdlog::source_loc struct being constexpr, there shouldn't be any additional runtime overhead in having filename, line number and function name available at all levels.

gabime:
Thanks. Merged

PR #963, opened 2019-01-14
Extended rotation file sink (with timestamp and daily rotation)

NGamzakov:
Can behave as rotating_file_sink.
Extended with:
1. Can accept timestamp format to append to the file name.
Can be omitted - passed as empty string or wstring. In this case the file will have default day time format (the same way in daily_file_sink).
If set the max_files will work only with timestamp unique files (if timestamp is with minute precision max_files will apply to current minute. After format returns different string, the max_files counter is as good as reset).
2. Can accept rotation_hour and rotation minute in addition to the size.
Can be omitted.
If set upon reaching rotation hour and minute, the last file will be closed and new file will be opened.

Example added with proper description (more or less)

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -21,6 +21,7 @@ void user_defined_example();
 void err_handler_example();
 void syslog_example();
 void clone_example();
+void rotating_example_extended();
 
 #include "spdlog/spdlog.h"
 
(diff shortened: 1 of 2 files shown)


NGamzakov:
I`ve needed this kind of sink so I`ve extended the rotation_file_sink and decided to make a pull request.
But I have no idea whats the build error above, so if anyone can lend me a hand please?

gabime:
“use of undeclared identifier 'calc_filename'; did you mean
      'calc_filename_'?
        file_helper_.open(calc_filename(0)); “

PR #969, opened 2019-01-16
Update log_msg.h

scribam:
`SPDLOG_NO_THREAD_ID` should only affect thread retrieval.


gabime:
Thanks for spotting this @scribam

PR #978, opened 2019-01-23
Add method to filesinks to return filename

### LINKED ISSUE #968: Getting filename sink is pointed to
Hello, I am trying to share the same file sink between multiple loggers, but I would like to use the sink pointers that are held by a given logger instead of maintaining a list of my own.  I know I can iterate the vector of sinks in a logger, but cannot find out what the filename it is using because it's private.  Can that detail be exposed in the API?  I'm happy to submit a PR to do it if it's something desirable.

markniebur:
Resolves #968


gabime:
Thanks @markniebur

PR #981, opened 2019-01-28
CMAKE_BINARY_DIR -> CMAKE_CURRENT_BINARY_DIR

ksergey:
The patch fix directory where version config file and pkg config file produced.


gabime:
What problem does it solve ?

ksergey:
I use spdlog as submodule as part of bigger project. Without this patch spdlog's cmake produce `spdlog*` files in main project build root instead of own build directory.

gabime:
Merged. Thanks @ksergey

PR #985, opened 2019-02-01
Allow the inclusion of an external fmtlib as as sub_directory

### LINKED ISSUE #982: Including both fmt and spdlog as subdirectories
I have a project where I want to include both spdlog and fmt as subdirectories, and use both of them be installed as subdirectories.

This "should" be trivial:
```
SET(SPDLOG_FMT_EXTERNAL ON CACHE BOOL "")

add_subdirectory(path_to_fmt)
add_subdirectory(path_to_spdlog)
```

But there's a problem. When using SPDLOG_FMT_EXTERNAL, spdlog always looks for fmt through `find_package()`, which will fail.

This should be fixable by changing [this](https://github.com/gabime/spdlog/blob/ac8a7bc12d389c72b508a499efbb5e9b594ef3d5/CMakeLists.txt#L57) check to:

```
if( SPDLOG_FMT_EXTERNAL AND NOT TARGET fmt::fmt )
```

(That is all unless there's some cmake magic I'm not aware about that can be used to prevent `find_package()` from running from my top-level project, of course...)

FrancoisChabot:
Fixes #982 

Tested to work well on Windows and Ubuntu and the newest versions of fmtlib. But coverage has been limited. Really curious what the CI will tell...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ option(SPDLOG_BUILD_BENCH "Build benchmarks (Requires https://github.com/google/
 option(SPDLOG_BUILD_TESTS "Build tests" ${SPDLOG_MASTER_PROJECT})
 option(SPDLOG_FMT_EXTERNAL "Use external fmt library instead of bundled" OFF)
 
-if(SPDLOG_FMT_EXTERNAL)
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. Merged

PR #986, opened 2019-02-03
Allow custom formatter in sink's ctor

DanielChabrowski:
Sink's constructor should accept formatter interface just like set_formatter member function.


gabime:
Thanks @DanielChabrowski . Good idea

PR #988, opened 2019-02-06
Remove unnecessary log function

DanielChabrowski:
Looks like one of the log functions is unnecessary since the "enable_if" check is later performed for log(source_loc...) version.


gabime:
Thanks @DanielChabrowski . Merged.

PR #990, opened 2019-02-06
Add CMake option for conditional installation

mnemotic:
This change makes it cleaner to use `spdlog` with CMake's `add_subdirectory` command (e.g. including it in the project tree as a Git submodule). Previously, `spdlog` headers etc. would be installed along with the outputs from the super-project.

Testing in my own project, this change doesn't seem to break anything, but that's hardly proof-positive that it won't ruin someone's day, somewhere, sometime.


gabime:
Merged. Thanks @mnemotic

PR #993, opened 2019-02-10
custom thread id types

alexezeder:
These changes allows to use custom types as thread identifiers to get more information inside formatters.
I faced with problem in spdlog when I tried to print thread names, which were named by pthread_setname_np function. Because thread identifiers (under Linux) received by syscall there is no way to get pthread identifier to use it in pthread_getname_np function. But now thread identifiers can have any type, can be received from any function and can be printed with any custom code (this is the most confusing one).
Some examples (all in tweakme.h):
_to use pthread on Linux_
`#define SPDLOG_CUSTOM_THREAD_T pthread_t`
`#define SPDLOG_CUSTOM_THREAD_ID_GETTER pthread_self()`
SPDLOG_CUSTOM_THREAD_ID_PRINTER can be not defined if pthread_t is integral, but if it is not then this macro can be defined but be empty - in case if you don't want to use pattern formatter

_to show what you can do with this_
`struct custom_thread_t
{
    some_type some_variable;
    int value = 42;
};`
`#define SPDLOG_CUSTOM_THREAD_T custom_thread_t`
`#define SPDLOG_CUSTOM_THREAD_ID_GETTER custom_thread_t()`
`#define SPDLOG_CUSTOM_THREAD_ID_PRINTER { fmt_helper::append_int(msg.thread_id.value, dest); }`

I think you should look at this as just a suggestion of using custom thread types, because I hope someone have better ideas to implement this. This feature will be really useful.

### DIFF
--- a/include/spdlog/details/log_msg.h
+++ b/include/spdlog/details/log_msg.h
@@ -41,7 +41,7 @@ struct log_msg
     const std::string *logger_name{nullptr};
     level::level_enum level{level::off};
     log_clock::time_point time;
-    size_t thread_id{0};
+    details::os::thread_t thread_id{};
     size_t msg_id{0};
 
     // wrapping the formatted text with color (updated by pattern_formatter).
--- a/include/spdlog/details/os.h
+++ b/include/spdlog/details/os.h
@@ -310,38 +310,49 @@ inline int utc_minutes_offset(const std::tm &tm = details::os::localtime())
 #endif
 }
 
-// Return current thread id as size_t
+#if defined(SPDLOG_CUSTOM_THREAD_T)
+using thread_t = SPDLOG_CUSTOM_THREAD_T;
+#else
+using thread_t = size_t;
+#endif
+
+// Return current thread id as thread_t
 // It exists because the std::this_thread::get_id() is much slower(especially
 // under VS 2013)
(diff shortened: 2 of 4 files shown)


gabime:
Thanks. Will keep this open for a while.

alexezeder:
I just want to say why, on my opinion, this is not the best solution. As I said before my task was to print thread names in log, so I tried to pass thread identifier (received by spdlog) to pthread and get a name of this thread. When I realized that log_msg receive thread id using `syscall(SYS_gettid)` call my first try was to add some `SPDLOG_PTHREAD_LINUX` define to tweakme and receive identifiers from pthread, but only if `defined(_POSIX_THREADS) && (_POSIX_THREADS > 0)` macro condition is true. This check means that basic support for POSIX threads is available and more importantly `pthread_self()` function should be present. But there was one flaw of my plan - `pthread_self()` returns `pthread_t` which can be any type, because 

> "POSIX.1 allows an implementation wide freedom in choosing the type used to represent a thread ID; for example, representation using either an arithmetic type or a structure is permitted"

 As you can see in my PR `pthread_t` can be passed to formatter (instead of `size_t`), there was not problem with that, but the real problem was to find the right way to print `pthread_t` in formatter. This is why I have decided to make more general solution, but this solution forces user to write some additional code on its side.

PR #996, opened 2019-02-18
Add tweak support for user short level names

MFornander:
We wanted the short names to be in sync with the long names and this is the minimal PR. Since the short names are already strings and not just chars we allow users to use longer short names.  Short names (%L) becomes more of an "alt-names" actually.


gabime:
Thanks. It was on my todo list.

PR #998, opened 2019-02-19
Add custom flags

MFornander:
This is an updated version of @fegomes PR from last year.  I made an effort to give as much credit as possible by merging in his code before updating and modifying it.

I updated the implementation to adhere to the current coding standards, added test, and added support for alignment of custom tags.  Your hesitations to accept the MR may still be valid but I thought I would open a new MR for others to see there is an implementation that compatible with the current Feb2019 spdlog.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ install_manifest.txt
 
 # idea
 .idea/
+
+# vscode
+.vscode/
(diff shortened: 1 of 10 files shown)


gabime:
Implemented new feature. Please see example.cpp for an example

PR #1034, opened 2019-03-22
List headers as sources

raitraak-rrk:
Visual Studio Intellisense and Resharper have issues recognizing headers from header only project. Proposed solution is to add the headers into the main library based on the option switch. Also this removes an extra target.


raitraak-rrk:
No idea why the test is failing under Mac. Can it be test machine load issue or too tight timing requirement?

gabime:
why is it better than the current approach (spdlog_headers_for_ide target)?

raitraak-rrk:
I was trying to replicate the issue I had with VS functionality, but am unable to do so after the latest update.. Maybe they actually fixed something for a change  :) 
So the only remaining argument for the PR is one less target. If you are using multiple libraries, the numbers can pile up.
If that's not enough, feel free to reject.
Thanks for the awesome library!

gabime:
That’s good news. Thanks anyway !

Issue #1025, opened 2019-03-14
Do not apply global setting when creating logger from std::make_shared


tt4g:
I am creating loggers dynamically with sinks divided by usage, but don't applied global settings to this loggers.

I set log level `spdlog::level::trace` globally and noticed this problem.

I noticed this problem by setting log level `spdlog::level::trace`globally with code like:

```cpp
spdlog::set_level(spdlog::level::trace); // set global log level

auto consoleSync = std::make_shared<spdlog::sinks::stdout_sink_st>();
consoleSync ->set_level(spdlog::level::trace); // set sync log level.
auto consoleLogger = std::make_shared<spdlog::logger>("console", consoleSync);
spdlog::register_logger(consoleLogger); // I thought that the log level was set to the logger here.

consoleLogger->info("first log"); // output console.
consoleLogger->trace("second log"); // no output, consoleLogger is log level spdlog::level::info.
```

If use `spdlog::create(logger_name, sync_args)`, `spdlog::details::registry::initialize_logger(logger)` will be called and global settings will be applied to the logger.
Global settings can be reflected manually after creating sink and logger itself, but it takes a lot of work.

I think public function `spdlog::initialize_logger(std::shared_ptr<spdlog::logger>)` or `spdlog::create(std::string logger_name, std::shared_ptr<Sink> sync)` should be added.

gabime:
register_logger just add the logger to the registry, it doens do anything else.
However, I agree, it can be confusing sometimes.

tt4g:
Is `spdlog::details::registry::initialize_logger(logger)`  possible to add public API?

PR #1035, opened 2019-03-25
Add spdlog::initialize_logger

tt4g:
Related #1025

Add public API `spdlog::initialize_logger` for create loggers manually.


gabime:
Thanks @tt4g

PR #1045, opened 2019-03-31
Namespace qualify stat functions

p-po:
Hi,

This is trivial change. In a large codebase where I use spdlog, I had to do this change. I know this is not at all a bug in spdlog but this could save trouble where people have to deal with large codebases and namespace abuses (In my case, a few _bad_ "using namespace My_NAME" directives, where MY_NAME contains a stat() function and fixing it will take quite some effort).

If nothing else, this makes it explicit that POSIX functions from global namespace are used.

Thanks.


gabime:
Thanks. Merged

Issue #1052, opened 2019-04-04
Log macro issues causing unexpected behaviour


db-panda:
Seemingly the SPDLOG_TRACE/_DEBUG/... macros are causing undefined behaviour, referring to this: https://stackoverflow.com/questions/55509290/c-why-is-my-else-statement-in-an-if-else-statement-ignored

PR #1055, opened 2019-04-04
Fix unexpected log macro expansion

db-panda:
Old macros caused trouble if used in if-statements with an else-block not wrapped in curly braces.

See issue #1052 and referenced stackoverflow posts.


gabime:
Thanks @db-panda

PR #1058, opened 2019-04-07
Add a CMake option to use a header-only external fmt

mnemotic:
With this, people who want to use the header-only variant of external `fmt` library can do so.

This change should be "safe" as the old behavior is preserved.


CMakeLists.txt
```diff
@@ -68,7 +65,16 @@ target_include_directories(
 
 if(SPDLOG_FMT_EXTERNAL)
     target_compile_definitions(spdlog INTERFACE SPDLOG_FMT_EXTERNAL)
-    target_link_libraries(spdlog INTERFACE fmt::fmt)
+
+    if(NOT TARGET fmt::fmt)
```
gabime: why this addition?

mnemotic: This puts everything relating to `SPDLOG_FMT_EXTERNAL` option into a single, neat block. See removed lines 58 - 60. Having it split up like that seemed like bad style to me.

PR #1064, opened 2019-04-17
Remove unnecessary semicolons (-Wextra-semi)

jhasse:
I'm compiling my code with [`-Wextra-semi`](https://clang.llvm.org/docs/DiagnosticsReference.html#wextra-semi) which causes warnings in one of spdlog's headers.

I haven't added `-Wextra-semi` in spdlog's CMakeLists.txt since it's only supported by newer GCC and Clang versions.


gabime:
Thanks @jhasse

PR #1065, opened 2019-04-20
Fixed line 493 error message.

badryasuo:
**Before it says:** `don't know how to format the type, include fmt/ostream.h if it provides an operator<< that should be used`
**Now it says:** `don't know how to format the type; include spdlog/fmt/bundled/ostream.h if it provides an operator<< that should be used.`


badryasuo:
Hello, sorry for not explaining much but  I've got that error recently and I tried to include `fmt/ostream.h` as it says but it didn't work out; So, I went to spdlog include directory and found that `ostream.h` exists inside the `fmt/bundled` directory, not `fmt` so I've pulled a request to fix this as you can see.

gabime:
fmt source belongs to the fmt repo. I try not to change it here. Maybe the doc/wiki should be fixed instead.

Issue #1074, opened 2019-05-09
Add option to force color output without TTY


psalz:
Currently the `ansicolor_sink` [checks for a TTY with color support](https://github.com/gabime/spdlog/blob/366935142780e3754b358542f0fd5ed83793b263/include/spdlog/sinks/ansicolor_sink.h#L41) in order to decide whether it should actually output any color codes. That is fine, however in some situations it is desirable to force color output anyway - for example when piping outputs through another executable.

Many applications thus typically provide a flag to force color output regardless of the destination, for example Clang's `-fcolor-diagnostic`. As far as I can tell there is currently no way (other than providing a custom sink) to enable such a flag using spdlog.

I thus propose adding an optional parameter to the color sink factory functions, i.e., ```stdout_color_mt(const std::string &logger_name, bool force_color = false)``` which will then be passed into the constructor of `ansicolor_sink` (not sure how the behavior for wincolor is or what the conventions are).

Alternatively / additionally a member function could be added to the sink, e.g. `ansicolor_sink::force_color_output(bool value)`.

gabime:
Agreed. I think ansicolor_sink::force_color_output is preferable to change to the ctor.

psalz:
What do you think about adding the option to the factory functions though? They could then call `force_color_output` internally before returning the pointer. My concern is that having it only on the sink requires care in user code because the sink returned by `stdout_color_mt` is platform dependent.

gabime:
Yea, sounds good.

gabime:
Closed. Thanks @psalz for fixing.

PR #1075, opened 2019-05-09
Add option to force color output without TTY

psalz:
This adds a member function `ansicolor_sink::force_color_output` which can be used to force the output of color codes even when no TTY with color support is detected, as discussed in #1074. The `std[out|err]_color_[mt|st]` factory functions now support an optional argument `force_color_output` (which defaults to `false`) to make it more convenient to set this option directly upon construction.

As per @gabime's request I didn't add a new parameter to `ansicolor_sink`'s constructor, which makes the factory functions a bit clunky, but I tried to keep everything DRY.

### DIFF
--- a/include/spdlog/sinks/ansicolor_sink.h
+++ b/include/spdlog/sinks/ansicolor_sink.h
@@ -33,12 +33,12 @@ class ansicolor_sink final : public sink
 {
 public:
     using mutex_t = typename ConsoleMutex::mutex_t;
-    ansicolor_sink()
+    ansicolor_sink(color_mode mode = color_mode::automatic)
         : target_file_(TargetStream::stream())
         , mutex_(ConsoleMutex::mutex())
 
     {
-        should_do_colors_ = details::os::in_terminal(target_file_) && details::os::is_color_terminal();
+        set_color_mode(mode);
         colors_[level::trace] = white;
         colors_[level::debug] = cyan;
         colors_[level::info] = green;
@@ -136,6 +136,22 @@ public:
     bool should_color()
     {
         return should_do_colors_;
+    }
+
+    void set_color_mode(color_mode mode)
+    {
+        switch (mode)
+        {
+        case color_mode::always:
+            should_do_colors_ = true;
+            return;
+        case color_mode::automatic:
+            should_do_colors_ = details::os::in_terminal(target_file_) && details::os::is_color_terminal();
+            return;
+        case color_mode::never:
+            should_do_colors_ = false;
+            return;
+        }
     }
 
 private:
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -160,6 +160,16 @@ using level_hasher = std::hash<int>;
 } // namespace level
 
 //
+// Color mode used by sinks with color support.
+//
+enum class color_mode
+{
+    always,
+    automatic,
+    never
+};
+
+//
 // Pattern time - specific time getting to use for pattern_formatter.
 // local time by default
 //
--- a/include/spdlog/sinks/stdout_color_sinks.h
+++ b/include/spdlog/sinks/stdout_color_sinks.h
@@ -31,26 +31,26 @@ using stderr_color_sink_st = ansicolor_stderr_sink_st;
 } // namespace sinks
 
 template<typename Factory = default_factory>
-inline std::shared_ptr<logger> stdout_color_mt(const std::string &logger_name)
+inline std::shared_ptr<logger> stdout_color_mt(const std::string &logger_name, color_mode mode = color_mode::automatic)
 {
-    return Factory::template create<sinks::stdout_color_sink_mt>(logger_name);
+    return Factory::template create<sinks::stdout_color_sink_mt>(logger_name, mode);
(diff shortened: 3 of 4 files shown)


include/spdlog/sinks/ansicolor_sink.h
```diff
@@ -138,6 +139,11 @@ class ansicolor_sink final : public sink
         return should_do_colors_;
     }
 
+    void force_color_output(bool value)
+    {
+        should_do_colors_ = value || can_do_colors_;
```
gabime: A bit confusing. It is called force so should be forced.  I think can_do_colors_ is not needed. just set here should_do_colors_ = value

psalz: The idea here is that if at some point the user decides they no longer want to force color output, the sink should return to its default behavior, i.e., printing color only if a TTY is found. So if the user calls `force_color_output(false)` the variable `should_do_colors` still remains true if there is a TTY with color support. Otherwise it doesn't make sense to have a `value` parameter in the first place (or the function should be called something like `enable_color_output`). I think this interface is more flexible and properly reflects the semantics of "forcing" color output.

gabime: but how would one turn off the color?

psalz: Ah, I see. We obviously had different behaviors in mind for this method. I just assumed if you don't want color output in the first place, you'd use a different sink. The problem I see with the semantics you suggest is that once this function has been called with either `true` or `false`, there is no way to get back to the sink's original "automatic" behavior. It's certainly interesting, as to me "forcing" means that I definitely want a certain behavior, while "not forcing" means that I don't really care (i.e., let the system decide for me).

I'm fine with either interpretation, however if we go with yours I'd suggest to rename it to `enable_color_output`, as in my mind it more clearly reflects what is happening. Also in that case the added parameter for the factory functions wouldn't work, as it would always either enable or disable color output, regardless of TTY or not. I guess we'd then have to provide overloads instead.

godbyk: Or pass in an enum value, one of `ALWAYS_COLOR`, `NEVER_COLOR`, or `AUTO_COLOR`. This would allow you to force color output, force no color output, or auto-detect whether there should be color output, respectively.

Best of both worlds!

psalz: Yes, I'd like that solution as well!

gabime: Thanks @godbyk - thats indeed good and clear solution. Might even be better to pass this in enum to the ctor after all. This solution even enable me to remove the redundant (non color) stdout/stderr sinks in future.

psalz:
Alright, I just pushed the new variant using an enum to control color output behavior, as discussed. It's called `color_mode` and can be one of `always`, `automatic` or `never`. I've also added support for this enum to the `wincolor_sink`, however `automatic` is just an alias for `always` here.

A couple of notes:

- I wasn't sure where (which file and namespace) the enum should live, so I went with common.h and the top-level `spdlog` namespace.
- I've kept the member function to set the mode after initial construction. It is now called `set_color_mode()`.
- Maybe there should also be a way to query the current color mode (e.g. `get_color_mode()`)?

Lastly, I think the `should_do_colors()` function is somewhat misleading now (and I always found the name a bit strange tbh) - maybe this should be deprecated?

psalz:
(Not sure why CI failed - seems unrelated?)

gabime:
Thanks. Looking good

gabime:
fyi. Added mutex protection around set_color_mode()

psalz:
Thanks, makes sense!

sweihub:
I am missing colors with the multiple sinks with zsh, how to solve it? thanks.
` combined_logger->info("Hello world has color");` has color, but `spdlog::info("no color");` is missing color.

```cpp
#include <iostream>
#include <spdlog/sinks/rotating_file_sink.h>
#include <spdlog/spdlog.h>

namespace spdlog {

// redirect logs to file
inline int
open(int daemon, std::string const &file, int max_size = 50 * 1024 * 1024, int max_files = 10)
{
    try {
        std::vector<spdlog::sink_ptr> sinks;

        if (!daemon) {
            // Color is missing
            auto terminal = std::make_shared<sinks::ansicolor_stdout_sink_mt>();
            terminal->set_color_mode(color_mode::always);
            sinks.push_back(terminal);
        }

        auto disk = std::make_shared<sinks::rotating_file_sink_mt>(file, max_size, max_files);
        sinks.push_back(disk);

        auto combined_logger = std::make_shared<spdlog::logger>("name", begin(sinks), end(sinks));
        combined_logger->info("Hello world has color");

        set_default_logger(combined_logger);
        flush_every(std::chrono::seconds(1));
        set_pattern("[%Y-%m-%d %H:%M:%S.%e] [%l] %v");
    } catch (const spdlog_ex &ex) {
        std::cout << "Log init failed: " << ex.what() << std::endl;
        return (-1);
    }

    return (0);
}
} // namespace spdlog
```

PR #1082, opened 2019-05-18
Refactor build system

DavidZemon:
Okay, I've made my changes and performed the following tests:

* Run `make all && make test` with default options
* Run `make all && make test` with a static library (default is shared)
* Updated the Conan recipe to make use of the new configuration and used `conan create ...` to verify that the recipe builds and that the recipe's `test_package` still works for
  * x86-64 Linux shared lib
  * x86-64 Linux static lib
  * ARMv7hf Linux shared lib
  * ARMv7hf Linux static lib
* Updated one of our apps to use this new Conan artifact and verified that it A) compiles and B) still passes unit tests

I did make some changes to your branch - some aesthetic, some functional. Some you might disagree with, others you'll hopefully appreciate :). Please ask if there's anything you're unsure of.

I did my best to split the changes into self-contained commits with detailed commit comments, so I do recommend reviewing this commit-by-commit rather than all at once.


DavidZemon:
I don't know why the build is failing from Windows and don't have a Windows development machine to test this on. Can you provide any insight? https://ci.appveyor.com/project/gabime/spdlog/builds/24635829/job/3yibukc9vkfqk2k7

gabime:
@DavidZemon Thanks ! I've done some changes and merged.

1. Added version file
2. Added spdlogConfig.in
3. Removed shared_lib option for now (I'm not very fond of shared libs in general).
4. Update the cmake example (tries to make the header only usage simpler)

Please review and comment

DavidZemon:
That all looks great. The one thing I'd comment on is the "forced" static library. You haven't actually forced users into a static library - I can still build a shared one by giving `-DBUILD_SHARED_LIBS=ON` on the command line. But you have stopped advertising the ability to do so by removing the `option(...)`. If you prefer static, I'd encourage you to put the `option(...)` back in, but set its default to `OFF`. My team is using spdlog on an embedded system, and currently the entire operating system is only ~140 MB. Linking spdlog in statically to multiple executables would increase that number very quickly.

gabime:
> I'd encourage you to put the option(...) back in, but set its default to OFF

But what's the point if the user can anyway do -DBUILD_SHARED_LIBS=ON from the command line?

DavidZemon:
Having the `option()` in there advertises the feature via `ccmake` and `cmake-gui`, and tells users "I respect your choice to use this option." Because you don't _have_ to respect their choice. The alternative is that you could have used `add_library(spdlog STATIC src/spdlog.cpp)` and then it would be misleading to put `option(BUILD_SHARED_LIBS ...)` because you're explicitly overriding that variable and forcing static.

gabime:
Ok. Sounds reasonable.

I wonder how this static lib improve build times for your projects. Any numbers?

DavidZemon:
With shared library: 3.1s for a recompile after adding a single line of whitespace
With header-only: 5.2s for the same test

If I rip out all of spdlog from the project, the same test is down to 2.3s - so spdlog's penalty has been decreased from 2.9 to 0.8s. Still a significant amount of time compile, but MUCH better.

gabime:
Thanks for the info. That's encouraging.

pwm1234:
Just a quick question: why is BUILD_SHARED_LIBS not supported on windows? I am a big fan of spdlog and work on with VS 2017 a lot. I would like to make this work this work on windows if I can.

DavidZemon:
Can you share the error message or line of code that indicates it is not working? Been a couple months since I messed with this, but I certainly don't recall doing anything that would have prevent shared libs on Windows, or any other platform.

pwm1234:
See the logic in the top-level CMakeLists.txt: https://github.com/gabime/spdlog/blob/v1.x/CMakeLists.txt#L70

DavidZemon:
Interesting.... @gabime will have to answer, as it wasn't added by me.

gabime:
supporting shared lib under windows requires some work(dllexport etc.), so it is not yet supported.

pwm1234:
If it is just work like adding dllexport, I can try to do that. Would you like me to try or is this something you would prefer to do yourself? Should I start with the HEAD of the master branch? Is there a particular test to pass?

gabime:
sure, you can try to use v1.x branch. also you can refer to #1156 that tried something smilier but got many warnings..

tbags:
If one were to not use _cmake_(after building the static lib) what's one good option to compile a program against linking this lib?

I'm wondering if this would pick references from the static lib vis-à-vis compiling from header?
`g++ -std=c++11 -L<PATH TO libspdlog.a> -I<PATH TO spdinclude> -o my_pgm my_pgm.cc -lspdlog` 

Currently, I'm trying something like this and see almost no improvement in compile time. Thank you for taking out some time to read my comment :)

DavidZemon:
Linking against static or shared library shouldn't change compile times. So long as you're using the new version that isn't header-only, you _should_ see improvements I would think.

gabime:
@tbags you  must pass -DSPDLOG_COMPILED_LIB,otherwise it will use the header only version.

After doing make install, you can try use “pkg-config spdlog —cflags —libs” to get the right flags instead of cmake

tbags:
Thank you @gabime  and @DavidZemon, works now with define flag` -DSPDLOG_COMPILED_LIB`, compile-time reduced from 25s to 11s on walltime.

PR #1105, opened 2019-06-10
Do not impose -Wextra -Wconversion etc on library users

### LINKED ISSUE #1104: CMake config imposes -Wextra -Wconversion on library users
In my installation, file `lib64/spdlog/cmake/spdlogConfigTargets.cmake` contains the following code:
```cmake
set_target_properties(spdlog::spdlog_header_only PROPERTIES
  INTERFACE_COMPILE_OPTIONS "-Wall;-Wextra;-Wconversion;-pedantic;-Wfatal-errors"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Threads::Threads"
)
```
This means that simply using `spdlog` in my application, for example like this:
```cmake
find_package(spdlog REQUIRED)
target_link_libraries(foo spdlog::spdlog)
```
...will build my code with `-Wall -Wextra -Wconversion -pedantic -Wfatal-errors`. That is very wrong.

I think that this was introduced in 6636ae6e633b3f888ab5504db09f91091fdab9a5 and 54f1941691ac0f8b51d8122a98c259c87cf8026b. That code looks like something that was probably supposed to kick in only when building `spdlog` itself, but it actually affects the resulting CMake configuration files as well. So, when I use CMake to *build* spdlog, these only-for-spdlog options actually taint the resulting CMake config files, and these extra warnings are imposed on code which only "uses" `spdlog`.

jktjkt:
Fixes #1104

I had to nuke the flags targetting `spdlog_header_only` because in CMake, an interface target apparently really only supports the `INTERFACE` option.


CMakeLists.txt
```diff
@@ -80,8 +80,7 @@ target_link_libraries(spdlog_header_only INTERFACE Threads::Threads)
 #---------------------------------------------------------------------------------------
 if(SPDLOG_MASTER_PROJECT)
     if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang")
-        target_compile_options( spdlog PUBLIC -Wall -Wextra -Wconversion -pedantic -Wfatal-errors)
-        target_compile_options( spdlog_header_only INTERFACE -Wall -Wextra -Wconversion -pedantic -Wfatal-errors)
+        target_compile_options( spdlog PRIVATE -Wall -Wextra -Wconversion -pedantic -Wfatal-errors)
     endif()
     if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
         target_compile_options( spdlog PUBLIC /W3 /WX )
```
gabime: how about msvc?

jktjkt: That's a good suggestion -- added the equivalent modifications.

PR #1114, opened 2019-06-18
Allow flush period to be any duration instead of std::chrono:seconds

moznuy:
I need to flush logs faster than one second, but discovered that it's not possible with current implementation of spdlog::flush_every.
I provided templated version of spdlog::flush_every and periodic_worker::periodic_worker() for support of arbitrary duration from std::chrono


gabime:
Thanks. I prefer just add a simple overload that takes millis instead of all those changes

PR #1122, opened 2019-06-27
Provide source location support for systemd_sink.h

jbelloncastro:
This PR includes support to send source code location information to the journal.
It replaces `sd_journal_print` with `sd_journal_send`, which allows more control over the journal's metadata.

If `spdlog-utets` contains `test_systemd.cpp`, the output may look like:
```
$ journalctl -r -o json-pretty _PID=20564 | jq -C 'select(.CODE_FILE != null)'
{
  "_PID": "20564",
  "_SYSTEMD_OWNER_UID": "1000",
  "_CMDLINE": "tests/spdlog-utests -n systemd",
  "_CAP_EFFECTIVE": "0",
  "_TRANSPORT": "journal",
  "_SYSTEMD_INVOCATION_ID": "db6561d285914cd59c2ee2e55bb786b4",
  "_UID": "1000",
  "_GID": "1000",
  "_COMM": "spdlog-utests",
  "CODE_LINE": "11",
  "_SYSTEMD_USER_UNIT": "gnome-terminal-server.service",
  "_SYSTEMD_CGROUP": "/user.slice/user-1000.slice/user@1000.service/gnome-terminal-server.service",
  "_AUDIT_LOGINUID": "1000",
  "_AUDIT_SESSION": "3",
  "CODE_FILE": "/home/jcastro/dev/libraries/spdlog/tests/test_systemd.cpp",
  "SYSLOG_IDENTIFIER": "spdlog-utests",
  "_SOURCE_REALTIME_TIMESTAMP": "1561643730729194",
  "_SYSTEMD_UNIT": "user@1000.service",
  "PRIORITY": "3",
  "MESSAGE": "test error",
  "__MONOTONIC_TIMESTAMP": "17746029539",
  "_SYSTEMD_USER_SLICE": "-.slice",
  "_SYSTEMD_SLICE": "user-1000.slice",
  "_EXE": "/home/jcastro/dev/libraries/spdlog/build/tests/spdlog-utests",
  "CODE_FUNC": "____C_A_T_C_H____T_E_S_T____0"
}
```

### DIFF
--- a/include/spdlog/sinks/systemd_sink.h
+++ b/include/spdlog/sinks/systemd_sink.h
@@ -7,8 +7,6 @@
 #include "spdlog/details/null_mutex.h"
 #include "spdlog/details/synchronous_factory.h"
 
-#include <array>
-#include <string>
 #include <systemd/sd-journal.h>
 
 namespace spdlog {
@@ -36,16 +34,16 @@ inline int syslog_level(level::level_enum l)
 }
 
 /**
- * Sink that write to systemd using the `sd_journal_print()` library call.
+ * Sink that write to systemd journal using the `sd_journal_send()` library call.
  *
- * Locking is not needed, as `sd_journal_print()` itself is thread-safe.
+ * Locking is not needed, as `sd_journal_send()` itself is thread-safe.
  */
 template<typename Mutex>
 class systemd_sink : public base_sink<Mutex>
 {
 public:
     //
-    explicit systemd_sink(void) {}
+    explicit systemd_sink() {}
 
     ~systemd_sink() override {}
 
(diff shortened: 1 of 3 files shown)


include/spdlog/sinks/systemd_sink.h
```diff
@@ -55,8 +56,26 @@ class systemd_sink : public base_sink<Mutex>
...
     {
-        if (sd_journal_print(syslog_level(msg.level), "%.*s", static_cast<int>(msg.payload.size()), msg.payload.data()))
+        const char* key_msg  = "MESSAGE=%.*s";
+        const char* key_prio = "PRIORITY=%d";
+        const char* key_file = "CODE_FILE=%s";
+        const char* key_line = "CODE_LINE=%" PRIu32;
+        const char* key_func = "CODE_FUNC=%s";
+        // Do not send source location if not available
+        if( !msg.source.filename ) {
+            key_file = nullptr;
```
gabime: is it safe?

include/spdlog/sinks/systemd_sink.h
```diff
@@ -7,10 +7,11 @@
 #include "spdlog/details/null_mutex.h"
 #include "spdlog/details/synchronous_factory.h"
 
-#include <array>
-#include <string>
+#define SD_JOURNAL_SUPPRESS_LOCATION // Don't add code location in sd_journal_send
```
gabime: what if already defined? please wrap with ifndef

jbelloncastro: I think I should add an `#undef` at the end of the header file, in case `systemd/sd-journal.h` is included somewhere else.
In this case however we want this macro to be defined there at all times. The macro avoids `sd_journal_send` to capture the source location where the function is called (`sink_it_`). Here we want to capture the location where spdlog users place their SPDLOG_LOGGER_* macros, and not where our sink sends the messages to the journal. In other terms, not setting the macro would set additional entry values pointing with  `CODE_FILE=systemd_sink.h` file, `CODE_FUNC=sink_it_` and `CODE_LINE=69`.

include/spdlog/sinks/systemd_sink.h
jbelloncastro: They should. I've updated the check.

include/spdlog/sinks/systemd_sink.h
jbelloncastro: I found a way to not depend on this macro definition. Hopefully I will have the change soon.

PR #1126, opened 2019-06-30
Add meson support

mensinda:
This PR adds support for the meson build system. This change allows easy inclusion of spdlog in meson projects. Additionally, most features from the current CMake build should also be supported.

A working example:

```
.
├── main.cpp
├── meson.build
└── subprojects
    └── spdlog
```

meson.build:

```meson
project('mesonTest', ['cpp'])

sub = subproject('spdlog')
dep = sub.get_variable('spdlog_dep')

executable('main', ['main.cpp'], dependencies: [dep])
```

main.cpp:

```cpp
#include <spdlog/spdlog.h>

int main(int argc, char *argv[]) {
  spdlog::info("Hello World");
  return 0;
}
```

Build instructions:
```bash
cd /path/to/source/root
meson build
ninja -C build
./build/main
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ example/*
 !example/example.sln
 !example/example.vcxproj
 !example/CMakeLists.txt
+!example/meson.build
 !example/multisink.cpp
 !example/jni
 
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,9 @@ build/*
 # Codelite
 .codelite
 
+# KDevelop
+*.kdev4
+
 # .orig files
 *.orig
 
--- a/bench/meson.build
+++ b/bench/meson.build
@@ -1,0 +1,13 @@
+benchmark = dependency('benchmark')
+
+bench_matrix = [
+  ['bench',           [spdlog_dep],            []],
+  ['async_bench',     [spdlog_dep],            []],
+  ['formatter-bench', [spdlog_dep, benchmark], ['all']],
+  ['latency',         [spdlog_dep, benchmark], []],
+]
+
+foreach i : bench_matrix
+  bench_exe = executable(i[0], i[0] + '.cpp', dependencies: i[1])
+  benchmark('bench_' + i[0], bench_exe, args: i[2])
+endforeach
--- a/example/meson.build
+++ b/example/meson.build
@@ -1,0 +1,8 @@
+example_matrix =  [
+  ['spdlog-example',    spdlog_dep],
+  ['spdlog-example-ho', spdlog_headeronly_dep],
+]
+
+foreach i : example_matrix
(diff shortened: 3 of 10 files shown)


gabime:
Cool! Thanks  @mensinda 

Would you able to support this if needed? Otherwise I’m reluctant to add a build system that I don't really understand nor have the bandwidth to support properly.
Also could you please put all those files under a separate “meson” folder ?

mensinda:
Yes, I will support this feature if necessary. Just ping me when there is an issue.

However, moving the `meson.build` files into a separate directory isn't that easy. Meson requires that there is a `meson.build` in the project root directory for subprojects to work. Additionally, `meson_options.txt` file has to be located next to the root `meson.build` file. Thus it is also impossible to move this file.

The only thing I could easily move is the `extract_version.py` script (maybe into a `scripts` directory, together with `format.sh` and `clang_tidy.sh`?).

mensinda:
Also, the current CI error seems to be unrelated.

If possible, I would also like to add meson support to Travis, just in case. However, I would like to do this in a separate PR since this requires debugging Travis, and I don't want to spam your build history with tons of debug builds :)

gabime:
> The only thing I could easily move is the extract_version.py script (maybe into a scripts directory, together with format.sh and clang_tidy.sh?).

Sounds good. Please note It would require small changes to the scripts to point to the parent dir.

> If possible, I would also like to add meson support to Travis, just in case. 

Lets not in the meanwhile. I have enough struggle with travis and already :)

Also, I noted  that the example is missing the meson build. Could you please add it?

mensinda:
Oh, I missed the example stuff in the `.gitignore`. Fixed now. I have also moved the scripts.

> > If possible, I would also like to add meson support to Travis, just in case.
> 
> Lets not in the meanwhile. I have enough struggle with travis and already :)

Sounds reasonable. Using a docker image helps a lot in my experience :)

gabime:
Merged.. Thanks @mensinda .

gabime:
@mensinda Seems the meson build fails in building .dll under Windows (`-Dlibrary_type=shared`).  

Could you please take a look ?

mensinda:
Looks like a symbol export issue. What are all the magic MSVC flags required to build spdlog as a dll?

gabime:
@mensinda In CMake it's done like like this:
```cmake
if(NOT SPDLOG_FMT_EXTERNAL AND NOT SPDLOG_FMT_EXTERNAL_HO)
        target_compile_definitions(spdlog PRIVATE FMT_EXPORT PUBLIC FMT_SHARED)
 endif()
```

Which means that when spdlog compiles (with the bundled fmt), it defines the ```FMT_EXPORT``` macro, but when it is being used the ```FMT_SHARED``` is used.

PR #1127, opened 2019-06-30
Fix deprecation warnings in filename_to_str and improve performance of wbuf_to_utf8buf

### LINKED ISSUE #569: VS 2017 Update 15.5: Build Error C4996
With update 15.5, MSVC has drastically changed the behaviour of the std::allocator. Just wanted to give you a heads-up:

`Error	C4996	'std::allocator<char>::allocate': warning STL4010: Various members of std::allocator are deprecated in C++17. Use std::allocator_traits instead of accessing these members directly. You can define _SILENCE_CXX17_OLD_ALLOCATOR_MEMBERS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.	...\include\spdlog\fmt\bundled\format.h	903	`

Lemme know if you need any more info.

sylveon:
This fixes #569

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -418,7 +418,7 @@ SPDLOG_INLINE void wstr_to_utf8buf(basic_string_view_t<wchar_t> wstr, fmt::memor
         return;
     }
 
-    int result_size = target.capacity();
+    int result_size = static_cast<int>(target.capacity());
     if ((wstr_size + 1) * 2 > result_size)
     {
         result_size = ::WideCharToMultiByte(CP_UTF8, 0, wstr.data(), wstr_size, NULL, 0, NULL, NULL);
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -36,6 +36,10 @@
 #include <share.h>
 #endif
 
+#if defined(SPDLOG_WCHAR_TO_UTF8_SUPPORT) || defined(SPDLOG_WCHAR_FILENAMES)
+#include <limits>
+#endif
+
 #else // unix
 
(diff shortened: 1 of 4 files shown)


gabime:
How does it solves the warning? seems to contain some changes  not related to the warning. 
Also, changing the signature  of  wbuf_to_utf8buf to return of allocated std::string instead of filling the given buffer is not good since the buffer already contains stack buffer ready to be filled, so no allocation is done.

sylveon:
Yeah, it also generally improves wbuf_to_utf8buf. I'll push another commit with your suggested approach.

include/spdlog/details/os-inl.h
```diff
@@ -398,28 +403,42 @@ SPDLOG_INLINE bool in_terminal(FILE *file) SPDLOG_NOEXCEPT
...
-    if (wbuf_size == 0)
+    if (wstr.size() > static_cast<size_t>(std::numeric_limits<int>::max()))
     {
+        throw spdlog::spdlog_ex("UTF-16 string is too big to be converted to UTF-8");
+    }
+
+    int wstr_size = static_cast<int>(wstr.size());
+    if (wstr_size == 0)
+    {
+        target.resize(0);
```
gabime: seems like redundant resize

sylveon: I added so that if somebody passes a previously used memory_buffer, it'll be correctly emptied instead of staying filled.

gabime: i meant the other resize , the one below

gabime: anyway, it would be interesting to compare the new performance. could you run your benchmark again?

PR #1128, opened 2019-07-01
Fix rotation condition when SPDLOG_NO_DATETIME is defined.

s-shin:
I noticed daily rotation was not worked when `SPDLOG_NO_DATETIME` is defined.
This PR will fix it :)


gabime:
thanks @s-shin !

s-shin:
@gabime Great! ...but sorry :sob:
I found one more `now.time` should be fixed.
Please check https://github.com/gabime/spdlog/pull/1129.

gabime:
Merged. Thanks.

Issue #1136, opened 2019-07-05
std::string_view instead of const char* and const string&


matt77hias:
Is it possible to replace the following method parameter input types:

- `const char*` and `const std::string&` -> `std::string_view`
- `const wchar_t*` and `const std::string&` -> `std::wstring_view`

Note that `fmt` works out-of-the-box with `std::(w)string_view` and note that this will not break existing code using `spdlog`.

gabime:
In which functions in particular ? and why?

matt77hias:
The main interface log methods:

```
template<typename... Args>
inline void info(const char *fmt, const Args &... args)
```
-->
```
template<typename... Args>
inline void info(string_view fmt, const Args &... args)
```

`std::(w)string_view` is the type [{fmt}](https://fmt.dev/latest/api.html) is using for its formatting and outputting methods. `std::(w)string_view`s unfortunately are not null-terminated and would thus first need to be converted to `std::(w)string`s (and thus allocate) in order to use `spdlog`.

In my own use case, I logged on top of {fmt} using `std::(w)string_view`. If `spdlog` used those as well, switching to `spdlog` is an option.

gabime:
Thats interesting.
 See 
https://github.com/gabime/spdlog/blob/ee87aee4ddde5355a18da49985542728e3d90820/include/spdlog/logger.h#L71

Are you saying that this line(passing the const char *fmt to format_t) is causing fmt to allocate??

matt77hias:
No it does not, if you have a ´const char*´ or ´const std::string&´, one can implicitly cast without allocation to ´std::string_view´ (Note that you still need to scan the ´const char*´ in O(N) to find the length, which will happen as {fmt} expects ´std::string_view´.). If you have a ´std::string_view´ on the other hand, there is no way to implicitly cast to ´const char*´ or ´const std::string&´. Instead a ´std::string´ should be constructed and an allocation will probably take place as the SSO will probably not suffice.

gabime:
Ah ok. Please note that there is an overload in spdlog that accepts a string view as single param ((https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/logger.h#L146),so if your usage is passing a pre formatted string view, it would work too.

matt77hias:
Unfortunately, the `string_view` is most of the time unformatted.
Note that you could route both the single and multiple argument log to fmt in a similar fashion:

```
[...]

PR #1138, opened 2019-07-06
 Added fmt::(w)string_view support.

matt77hias:


matt77hias:
Related to [Issue](https://github.com/gabime/spdlog/issues/1136)

gabime:
Thanks. Please use spdlog::string_view_t instead

matt77hias:
Ok, but I need to add a `spdlog::wstring_view_t` as well for `wchar_t`.

gabime:
ok

PR #1140, opened 2019-07-08
Fix syslog output

mattiasj-axis:
payload does not appear to be reliably null terminated and leaks
data. Use size to the formatter to reliably terminate messages.


gabime:
Good catch. Thanks!

PR #1145, opened 2019-07-13
Fix compiler and linker errors

k0zmo:
This PR fixes two issues I recently encountered:

1. Following code triggers compilation error caused by missing include of `spdlog/details/pattern_formatter.h`.
```cpp
#include <spdlog/spdlog.h>
#include <spdlog/sinks/stdout_sinks.h>

int main()
{
    std::make_shared<spdlog::sinks::stdout_sink_mt>();
}
```

2. Using templated constructor for `async_logger` along with building spdlog as a static library triggers linker error when:

a) We use anything else than `std::vector` iterators:
```cpp
#include <spdlog/async.h>
#include <spdlog/sinks/stdout_sinks.h>
#include <list>

int main()
{
    std::list<spdlog::sink_ptr> sinks = {std::make_shared<spdlog::sinks::stdout_sink_mt>()};
    std::make_shared<spdlog::async_logger>("logger", begin(sinks), end(sinks), spdlog::thread_pool());
}
```
Using vector as a sink container "seems" fine because `async_logger::clone` instantiates the template specialization for `std::vector` iterators.

b) We compile with clang (tested with 6.0) with -O2 and -flto. This makes the aforementioned function `async_logger::clone` to be dead-code eliminated and subsequently no one instantiates vector iterator specialization of templated `async_logger` constructor.

### DIFF
--- a/include/spdlog/sinks/base_sink.h
+++ b/include/spdlog/sinks/base_sink.h
@@ -41,6 +41,6 @@ protected:
 } // namespace sinks
 } // namespace spdlog
 
-#ifndef SPDLOG_COMPILED_LIB
+#ifdef SPDLOG_HEADER_ONLY
 #include "base_sink-inl.h"
 #endif
--- a/include/spdlog/async_logger-inl.h
+++ b/include/spdlog/async_logger-inl.h
@@ -14,14 +14,6 @@
 #include <memory>
 #include <string>
 
-template<typename It>
-SPDLOG_INLINE spdlog::async_logger::async_logger(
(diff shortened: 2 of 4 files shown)


include/spdlog/sinks/base_sink.h
```diff
@@ -41,6 +41,6 @@ class base_sink : public sink
 } // namespace sinks
 } // namespace spdlog
 
-#ifndef SPDLOG_COMPILED_LIB
+#ifdef SPDLOG_HEADER_ONLY
```
gabime: why the change? rest of the code use it already.

gabime:
Thanks. Why the change of the ifded?

gabime:
Thanks! Merged

PR #1155, opened 2019-07-17
Update README.md

raphaelts3:
Add steps to compile and install from source


raphaelts3:
> Maybe more suitable under the install section

Agreed, how about that?

PR #1156, opened 2019-07-18
add spdlog_api macros for visibility

taiyu-len:
This PR adds import/export (for windows) and visibility (for linux) attributes, as well as a few small changes.

adds portable(hopefully) `SPDLOG_API` and `SPDLOG_PRIVATE` macros to mark spdlogs public types+functions and private functions respectively.
adds public compile_definition `SPDLOG_SHARED` when building shared library.
made fmt libs API macro portable, could probably be upstreamed.
made sure `FMT_SHARED`, `FMT_EXPORT`, `FMT_HEADER_ONLY` are defined when spdlogs version of those are defined when using bundled fmt.
added common.h to first of every include to ensure defines are defined when needed.
put console_globals singleton functions into a inline header.
~made stdout_sinks inline header guarded like other sinks.~ in upstream, rebased to remove.

This has only been tested on linux so far, but when using `BUILD_SHARED_LIBS=ON` all tests and examples work with `-fvisibility=hidden` which means all the symbols we need (at least what gets tested) are being exported properly.
hopefully this might fix the shared library issue on windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,12 @@ target_include_directories(spdlog PUBLIC
         "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
         "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
 target_link_libraries(spdlog PUBLIC Threads::Threads)
+set_target_properties(spdlog PROPERTIES DEFINE_SYMBOL SPDLOG_EXPORTS)
+
+if (BUILD_SHARED_LIBS)
+    target_compile_definitions(spdlog PUBLIC SPDLOG_SHARED)
+endif ()
+
 spdlog_enable_warnings(spdlog)
 
 #---------------------------------------------------------------------------------------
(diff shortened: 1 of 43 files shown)


gabime:
Unfortunately under windows it doesn't compile cleanly. There are 720 ```C4251``` warnings..

taiyu-len:
clang does support visibility attributes, and builds and links cleanly (minus one warning about converting signed to unsigned that was already present)

can you post the compile log?
looking it up, that warning seems to be primarly due to using stl in the public interface.

also do those warning show up only when building as a shared lib or as a static lib as well,
since i copied the way fmt sets those attributes by using the export attribute whether its built as shared or static.

gabime:
I don't have the full log at hand, but it happens only for shared lib.

caiazza:
> There are 720 C4251 warnings

As taiyu said this is common issue on windows due to the fact that the STL are normally exported. It is usually not an issue, especially if your STL members are private class members. You can check this SO answer:
https://stackoverflow.com/questions/16419318/one-way-of-eliminating-c4251-warning-when-using-stl-classes-in-the-dll-interface

You can silence it in each header enclosing the offending declarations in the following pragma

```
#if defined(_WIN32) || defined(__WIN32__) || defined(WIN32)
#pragma warning(push)
#pragma warning(disable : 4251)
#endif

Your code

#if defined(_WIN32) || defined(__WIN32__) || defined(WIN32)
#pragma warning(pop)
#endif
```

Another option is to export the public methods one by one rather than the whole class, which includes the private member where you typically use STL data members. You will still have the problem with the smart pointers you use to pass data around though

PR #1161, opened 2019-07-20
meson: Do not install when used as a subproject

mensinda:
With this PR, spdlog is no longer installed when used as a subproject. Additionally, the library type is set to `static` when used as a subproject, and `shared` when not. This behavior can be overwritten with a meson option.


gabime:
> Additionally, the library type is set to static when used as a subproject, and shared when not

Why is that? seems confusing

mensinda:
First of all, when spdlog is used as a subproject in meson, it is generally done as a fallback for the spdlog dependency. Something like this:

```meson
spdlog = dependency('spdlog', fallback: ['spdlog', 'spdlog_dep'])
```

This way, a project can either use the system version of spdlog, or the subproject if spdlog is not installed. Before this PR, when a user or distribution packager installs the project, spdlog is installed with it. Projects installing their own dependencies can be annoying for various reasons, so this should be avoided (that's why spdlog no longer installs anything when used as a subproject).

However, not installing spdlog will cause issues if a shared library is used (since the spdlog.{dll,so} is not installed). That's why static libraries should be used in this case. On the other hand, when spdlog is build as a standalone, it is usually installed system-wide. In that case shared libraries are prefered by most distributions.

gabime:
How about doing static by default in both cases, and providing an option to do shared  if needed? This seems more consistent.

mensinda:
Sure, that also works. Done.

PR #1169, opened 2019-07-24
Add more overloads to spdlog::log and spdlog::logger::log

sylveon:
This adds a couple more overloads:
- One in `spdlog::log` which allows to pass a `source_loc` when passing an arbitrary `T` object
- One in `spdlog::log` which allows to pass a `source_loc` when using a wide character format string
- One in `spdlog::logger::log` which allows spdlog to avoid going through fmtlib when the type it is receiving is statically convertible to `wstring_view_t`.


gabime:
Looks good. Thanks!

PR #1192, opened 2019-08-23
cmake: Use DATAROOTDIR for pkg-config and cmake files.

orbea:
Please correct me if I am mistaken, but spdlog does not install any architecture independent binary files?

 If so these files do not have any architecture independent data and only headers are installed so it is better to use architecture independent install paths like `/usr/share`.

On my system without setting any paths via cmake these files are now installed.
```
usr/share/
usr/share/cmake/
usr/share/cmake/spdlog/
usr/share/cmake/spdlog/spdlogConfig.cmake
usr/share/cmake/spdlog/spdlogConfigVersion.cmake
usr/share/cmake/spdlog/spdlogTargets.cmake
usr/share/pkgconfig/
usr/share/pkgconfig/spdlog.pc
```
This will allow distros to set spdlog to `noarch` or similar instead of `x86_64,` `i686` or etc.

Again please correct me if I made any wrong assumptions.

I was also able to confirm that at least one project (`libixion`) was still able to build with spdlog and this change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,13 +160,13 @@ if (SPDLOG_INSTALL)
     set(project_config_out "${CMAKE_CURRENT_BINARY_DIR}/spdlogConfig.cmake")
     set(config_targets_file "spdlogConfigTargets.cmake")
     set(version_config_file "${CMAKE_CURRENT_BINARY_DIR}/spdlogConfigVersion.cmake")
-    set(export_dest_dir "${CMAKE_INSTALL_LIBDIR}/spdlog/cmake")
+    set(export_dest_dir "${CMAKE_INSTALL_DATAROOTDIR}/spdlog/cmake")
 
     #---------------------------------------------------------------------------------------
     # Include files
     #---------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


orbea:
I believe the travis failures are unrelated to anything I did here.

gabime:
> spdlog does not install any architecture independent binary files?

It actually installs  libspdlog.a file which is architecture dependent

orbea:
Would you mind elaborating on how to build that? It does not seem to build by default.

Regardless there seems to be no reference to the library in the pkgconfig file or cmake files.
```
prefix=/usr
includedir=${prefix}/include

Name: spdlog
Description: Super fast C++ logging library. 
Version: 1.3.1
```
I suspect they should either be changed to reflect that the library exists or this PR is still correct. You would likely know which is more appropriate. :)

gabime:
It is build in v1.4.0 - not yet released, but you can try checking out from latest master

orbea:
Yes, I see it in the master. My apologies for missing that and this PR won't make sense for 1.4 and newer.

However I see a new problem in the master, the install target seems to be missing the pkgconfig file altogether?

gabime:
yes. didnt have time to add it

PR #1216, opened 2019-09-08
Transformed the registry in a normal class with a global instance

caiazza:
As I told you in the issue I opened, there is no reason I can see for the registry to be a singleton. It can still have its global instance but also allow for a local instance to be defined so that you can use spdlog as a local service of the user code

I also moved the create template function you embedded in the synchronous_factory class in the registry itself so that one can use it with the local registry as well as with the global one. 

I tested with the latest version of visual studio and with gcc 7.4.0 on Ubuntu. All changes should be completely backward-compatible


gabime:
> I also moved the create template function you embedded in the synchronous_factory class in the registry itself so that one can use it with the local registry as well as with the global one.

What about the async factory?

caiazza:
I didn't find the async factory yesterday. I noted afterwards that it is in a separate header. In any caseI wouldn't have changed it because I wanted your opinion about this minimal change first as those would be a bit more substantial and you should decide what policy to adopt about it. If you don't want to add the async functions directly to the registry header, one can add a function which takes a registry as argument as overload of the current async function and let the current function call that using the global instance as argument. This would be the most minimal change.
Otherwise, because you have the thread pool embedded in the registry you could have this async functions as templated methods of the registry itself, something like create_async. One advantage of this second design is that you could get rid of all the accessors to the thread pool and tp_mutex which currently seemed to be used only by the async functions. Imo this would make the design much cleaner and robust.
If you look at it the actual factory IS already the registry, and you should use it like that.

gabime:
Seems it is not so simple as we thought and I don’t want to make big changes right now(before official release), so I will just leave this open for a while until I have clear idea on this.

Thanks anyway!

caiazza:
As you wish, that's why I limited myself to those small changes which basically change nothing in the structure of the code. You simply have the factory function directly embedded in the registry while for the async you have anyway a separate header that doesn't need to be changed. The other changes are a bit deeper but they still would mostly be cosmetic, rather than logical.

From a personal point of view I would suggest to merge those couple of commits just so I have my fork in sync with yours and I can easily merge your next changes and test them in my project.

Best

caiazza:
Hi Gabi,
As I 'm not very experienced in the collaborative development, can I ask you why my proposal was unsatisfactory?

Best

gabime:
@caiazza Like i said, it seems not very trivial and incomplete (async factory) and I don't want to perform significant changes in this area right now. Thanks for your efforts!

caiazza:
I didn't add the async factory because I was waiting for your feedback after the official release. I didn't add the async factory at the beginning because I forgot that it was there but it's just as trivial as with the sync factory: just copy the code and refer to the local instance of the registry instead of the global and then in the existing global factory function you can simply call the local factory function of the global instance.
I will add that part as well and push you a new pull request.
I'm currently using it with the changed I pushed you so that I can instantiate independent logging services for different components of my application and it's working very well.

PR #1221, opened 2019-09-10
Add message-count formatter "%j"

josefschroettle:
Every logger gets its own message count which is incremented atomically
Now still wrapped in '#if defined()' so this could be an option in tweakme.h


gabime:
Thanks. However the feature was existed already and recently removed.

PR #1228, opened 2019-09-17
Added ARCH_INDEPENDENT option

### LINKED ISSUE #722: CMake config enforces 32/64 bitness
spdlog is enforcing 32 or 64 bit mode in the CMake config, which I don't think makes sense for a header-only library.

Package spdlog 0.17.0 using [the instructions](https://github.com/gabime/spdlog/wiki/9.-CMake):

```
[spdlog]> cmake -H. -B_builds -DCMAKE_INSTALL_PREFIX=/path/to/install -DCMAKE_BUILD_TYPE=Release
[spdlog]> cmake --build _builds --target install
```

On Windows 7 with the Visual Studio 2015 generator, but I believe the principle is generic, if you use the 32-bit compiler then `spdlogConfigVersion.cmake` enforces that:

```cmake
# if the installed or the using project don't have CMAKE_SIZEOF_VOID_P set, ignore it:
if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "" OR "4" STREQUAL "")
   return()
endif()

# check that the installed version has the same 32/64bit-ness as the one which is currently searching:
if(NOT CMAKE_SIZEOF_VOID_P STREQUAL "4")
  math(EXPR installedBits "4 * 8")
  set(PACKAGE_VERSION "${PACKAGE_VERSION} (${installedBits}bit)")
  set(PACKAGE_VERSION_UNSUITABLE TRUE)
endif()
```

And vice-versa if building for 64-bit.

So this can fail:

```cmake
find_package(spdlog 0.17.0 CONFIG REQUIRED)
```

With:

```
CMake Error at CMakeLists.txt:14 (find_package):
  Could not find a configuration file for package "spdlog" that is compatible
  with requested version "0.17.0".

    C:/tools/spdlog-0.17.0/lib/cmake/spdlog/spdlogConfig.cmake, version: 0.17.0 (32bit)
```

Even though the right version is present.

scatter-dev:
Fixes #722 as CMake has added an option to disable enforcing 32/64-bitness parity for header-only libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,7 +197,11 @@ if (SPDLOG_INSTALL)
 
     include(CMakePackageConfigHelpers)
     configure_file("${project_config_in}" "${project_config_out}" @ONLY)
-    write_basic_package_version_file("${version_config_file}" COMPATIBILITY SameMajorVersion ARCH_INDEPENDENT)
+    if(${CMAKE_VERSION} VERSION_GREATER 3.13)
+        write_basic_package_version_file("${version_config_file}" COMPATIBILITY SameMajorVersion ARCH_INDEPENDENT)
+    else()
+        write_basic_package_version_file("${version_config_file}" COMPATIBILITY SameMajorVersion)
+    endif()
     install(FILES
             "${project_config_out}"
             "${version_config_file}" DESTINATION "${export_dest_dir}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Copyright(c) 2019 spdlog authors
 # Distributed under the MIT License (http://opensource.org/licenses/MIT)
 
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.14)
 
 #---------------------------------------------------------------------------------------
 # Start spdlog project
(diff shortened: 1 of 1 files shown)


scatter-dev:
It appears the CI is failing as its CMake version is out of date.

gabime:
3.14 will invalidate lots of existing installations. instead of forcing 3.14, can it  done conditionally after checking the version is right?

scatter-dev:
Unless I'm misunderstanding something about how users tend to install, this won't break existing installations. This doesn't force 3.14 for using spdlog, only for building it. [3.14 is when the ARCH_INDEPENDENT keyword was added](https://gitlab.kitware.com/cmake/cmake/issues/16184), which is what allows the generated files to ignore 32/64-bitness.

gabime:
> This doesn't force 3.14 for using spdlog, only for building it

Yes this is the problem. I don't wont to mandate such recent CMake version. What's the problem to add simple condition ?

scatter-dev:
No problem, I just misunderstood you.

gabime:
Thanks :+1:  Merged

gabime:
@scatter-dev the CMake docs state that 

> ARCH_INDEPENDENT is intended for header-only libraries or similar packages with no binaries.

So the non-header-only will be affected ? it is also in the same package

PR #1234, opened 2019-09-20
Unbreak on more BSDs

jbeich:
Upstreaming https://github.com/DragonFlyBSD/DeltaPorts/blob/c9622be6d398/ports/devel/spdlog/dragonfly/patch-include_spdlog_details_os.h

Instead of `lwp_gettid` let's use `pthread_getthreadid_np` which is available on [AIX](https://www.ibm.com/support/knowledgecenter/en/ssw_ibm_i_74/apis/users_22.htm), [DragonFly](https://www.dragonflybsd.org/cgi/web-man?command=pthread_getthreadid_np&section=3), [FreeBSD](https://man.freebsd.org/pthread_getthreadid_np/3)

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -50,8 +50,17 @@
 #ifdef __linux__
 #include <sys/syscall.h> //Use gettid() syscall under linux to get thread id
 
-#elif defined(__FreeBSD__)
-#include <sys/thr.h> //Use thr_self() syscall under FreeBSD to get thread id
+#elif defined(_AIX)
+#include <pthread.h>     // for pthread_getthreadid_np
+
+#elif defined(__DragonFly__) || defined(__FreeBSD__)
+#include <pthread_np.h>  // for pthread_getthreadid_np
+
+#elif defined(__NetBSD__)
+#include <lwp.h>         // for _lwp_self
+
+#elif defined(__sun)
+#include <thread.h>      // for thr_self
 #endif
 
 #endif // unix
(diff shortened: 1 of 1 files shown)


gabime:
Are those changes were actually tested to work on all those platforms? I don’t have the setup to try those.

jbeich:
Both issues are common:
- `fstat64` never existed on any BSD as `off_t` became 64-bit **decades** ago
- `pthread_getthreadid_np` and `_lwp_self` copypasta was tested by  Firefox users for years
- `getthrid` is recent but OpenBSD doesn't support old releases; confirmed to work as expected in a VM

I don't have access to AIX or Solaris.

gabime:
Merged. Thanks.

gabime:
btw why did you remove the  “!(__CYGWIN__)”?

jbeich:
> btw why did you remove the `!(CYGWIN)`?

Blacklist was replaced with whitelist i.e., list only broken platforms. Cygwin doesn't define `__linux__`.
https://github.com/llvm/llvm-project/blob/llvmorg-8.0.1/clang/lib/Basic/Targets/X86.h#L513-L523

According to [Glibc docs](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html) one can define `-D_FILE_OFFSET_BITS=64` to make `fstat` behave as `fstat64` but  `os-inl.h` is part of spdlog API. As spdlog headers may not be the first headers in a consumer by the time `_FILE_OFFSET_BITS` is defined system headers may already be processed. So, we can't get rid of `fstat64` conditional yet.

gabime:
> Blacklist was replaced with whitelist i.e., list only broken platforms.

Looks better. Thanks

Issue #1237, opened 2019-09-22
Please add a pkgconfig file.


orbea:
In the update from `1.3.1` to `1.4.0` spdlog lost its pkgconfig file and this breaks the configure process for some programs that use spdlog as a dependency. It would be great if it could get added back.

For example with `libixion-0.15.0`.
```
checking for SPDLOG... no
configure: error: Package requirements (spdlog >= 0.16.0) were not met:

Package 'spdlog', required by 'virtual:world', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables SPDLOG_CFLAGS
and SPDLOG_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```

gabime:
I just didn't have time to deal with this. PRs are welcome

orbea:
@gabime I'll give it a try, but what is the correct linking flag for the spdlog library?

PR #1238, opened 2019-09-23
cmake: Add a new spdlog.pc pkgconfig file.

orbea:
Fixes https://github.com/gabime/spdlog/issues/1237.

Some things to note.

* I only tested this with the shared library, I am not sure if anything more needs to be done with the static library version or the header only version.
* The pkgconfig file expects `CMAKE_INSTALL_LIBDIR` to be a relative path, I am not sure there is a better way to handle this. It is important to retain the `{exec_prefix}/` and only append the relative library directory (i.e. `lib64`) for cross compiling.
* Its important to add `-DSPDLOG_FMT_EXTERNAL` to the `CFlags` when using an external fmt so that other projects that use `spdlog` find the correct headers when including `fmt.h`.
* This is enough to fix the `libixion` build, but currently they use the `spdlog.pc` file, the headers and not the library.

This is the generated pkgconfig file on my system.
```
prefix=/usr
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib64

Name: libspdlog
Description: Fast C++ logging library.
URL: https://github.com/gabime/spdlog
Version: 1.4.0
CFlags: -I${includedir}/spdlog -DSPDLOG_FMT_EXTERNAL
Libs: -L${libdir}/spdlog -lspdlog
```
Please review in case I missed anything, thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,7 @@ set(SPDLOG_SRCS
         src/file_sinks.cpp
         src/async.cpp)
 
+set(SPDLOG_CFLAGS "${PROJECT_NAME}")
 
 if (SPDLOG_BUILD_SHARED)
     if(WIN32)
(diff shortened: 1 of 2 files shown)


orbea:
For reference here is the pkgconfig file with default options.
```
prefix=/usr/local
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib

Name: libspdlog
Description: Fast C++ logging library.
URL: https://github.com/gabime/spdlog
Version: 1.4.0
CFlags: -I${includedir}/spdlog
Libs: -L${libdir}/spdlog -lspdlog
```

gabime:
Thanks!

PR #1243, opened 2019-09-24
Add option to selection MSVC runtime when compiling spdlog as a library

cmorrison31:
When compiling spdlog to use as a library, there is currently no way to select what MSVC runtime to use. A project that links against the MSVC static runtime will be forced to use spdlog as a header only lib. This pull request creates a new CMake variable, SPDLOG_MSVC_RUNTIME_LIBRARY, which doesn't change default behavior.

Unfortunately, CMake has no built-in support for the runtime library flag until version 3.15.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,7 +104,8 @@ endif()
 add_library(spdlog::spdlog ALIAS spdlog)
 
 target_compile_definitions(spdlog PUBLIC SPDLOG_COMPILED_LIB)
-target_compile_options(spdlog PRIVATE /${SPDLOG_MSVC_RUNTIME_LIBRARY}$<$<CONFIG:Debug>:d>)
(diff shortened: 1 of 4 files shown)


tt4g:
Some CMake generators contain `/MD` and `/MT` in `CMAKE_CXX_FLAGS`. It is dangerous to just add the compiler flags `/MD` and `/MT`.

see: [Compile with /MT instead of /MD using CMake \- Stack Overflow](https://stackoverflow.com/questions/14172856/compile-with-mt-instead-of-md-using-cmake)

gabime:
> Some CMake generators contain /MD and /MT in CMAKE_CXX_FLAGS. It is dangerous to just add the compiler flags /MD and /MT.

Yes, seems problematic. If needed, one should just upgrade to CMake3.15 and pass the 'MSVC_RUNTIME_LIBRARY' flag.

PR #1245, opened 2019-09-27
Update example.cpp

alongL:
if there is no this file ["logs/multisink.txt"] , true will result in file open  exception.


gabime:
This should not throw. What does the exception say?

alongL:
> This should not throw. What does the exception say?

This should be auto create, doesn't it?

PR #1246, opened 2019-09-27
Set additional CPACK variables for RPM generator

tcraigtyler:
Sets the following when querying the generated RPM's info:

Group       : System Environment/Libraries
License     : MIT
URL         : https://github.com/gabime/spdlog
Summary     : Fast C++ logging library
Description :
Very fast, header-only/compiled, C++ logging library.


gabime:
Thanks

Issue #1251, opened 2019-09-30
Compile error on Windows (Visual Studio 2017) with in pattern_formatter-inl.h (with std::chrono)


bsergean:
I started getting that error after upgrading to latest master (as of today or tomorrow), and compiling the lib **by using cmake include_directory** instead of using a single header file.

```
c:\projects\ixwebsocket\third_party\spdlog\include\spdlog\details\pattern_formatter-inl.h(881): error C2589: '(': illegal token on right side of '::'
c:\projects\ixwebsocket\third_party\spdlog\include\spdlog\details\pattern_formatter-inl.h(880): note: while compiling class template member function 'void spdlog::details::elapsed_formatter<Padder,std::chrono::seconds>::format(const spdlog::details::log_msg &,const tm &,spdlog::memory_buf_t &)'
        with
        [
            Padder=spdlog::details::null_scoped_padder
        ]
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\type_traits(616): note: see reference to class template instantiation 'spdlog::details::elapsed_formatter<Padder,std::chrono::seconds>' being compiled
        with
        [
            Padder=spdlog::details::null_scoped_padder
        ]
c:\projects\ixwebsocket\third_party\spdlog\include\spdlog\details\pattern_formatter-inl.h(1216): note: see reference to class template instantiation 'std::is_convertible<spdlog::details::elapsed_formatter<Padder,std::chrono::seconds> *,_Ty *>' being compiled
        with
        [
            Padder=spdlog::details::null_scoped_padder,
            _Ty=spdlog::details::flag_formatter
        ]
c:\projects\ixwebsocket\third_party\spdlog\include\spdlog\details\pattern_formatter-inl.h(881): note: see reference to variable template 'const bool conjunction_v<std::negation<std::is_array<spdlog::details::elapsed_formatter<spdlog::details::null_scoped_padder,std::chrono::duration<__int64,std::ratio<1,1> > > > >,std::is_convertible<spdlog::details::elapsed_formatter<spdlog::details::null_scoped_padder,std::chrono::duration<__int64,std::ratio<1,1> > > *,spdlog::details::flag_formatter *>,std::is_convertible<std::default_delete<spdlog::details::elapsed_formatter<spdlog::details::null_scoped_padder,std::chrono::duration<__int64,std::ratio<1,1> > > >,std::default_delete<spdlog::details::flag_formatter> > >' being compiled
[...]

PR #1252, opened 2019-09-30
Fix windows compile error where std::max is overriden by a macro (#1251)

bsergean:
See https://github.com/gabime/spdlog/issues/1251

Stack Overflow says that putting parenthesis around std::max is a way to fix it, and I've seen this pattern used twice in the code base (actually in fmt library).

```
include/spdlog/fmt/bundled/printf.h
47:    return (std::max)(static_cast<int>(value), 0);

include/spdlog/fmt/bundled/format.h
1041:    int num_zeros = (std::max)(params.num_digits - full_exp, 1);
```

The type of the objects wasn't super clear so I went with the parenthesis option.

### DIFF
--- a/include/spdlog/details/pattern_formatter-inl.h
+++ b/include/spdlog/details/pattern_formatter-inl.h
@@ -878,7 +878,7 @@ public:
 
     void format(const details::log_msg &msg, const std::tm &, memory_buf_t &dest) override
     {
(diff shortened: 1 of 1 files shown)


bsergean:
https://stackoverflow.com/questions/2789481/problem-calling-stdmax

gabime:
Thanks!

bsergean:
Sure thing.

PR #1253, opened 2019-10-01
CMakeLists.txt: removed spdlog subdir from library install

DasRoteSkelett:
The library gets installed into ${CMAKE_INSTALL_LIBDIR}/spdlog, which
is for unix / linux a rather strange place, hence, put it where the linker
is more likely to find it.

Signed-off-by: Matthias Schoepfer <matthias.schoepfer@ithinx.io>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,7 +196,7 @@ if (SPDLOG_INSTALL)
     # Include files
     #---------------------------------------------------------------------------------------
     install(DIRECTORY include/ DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}" PATTERN "fmt/bundled" EXCLUDE)
-    install(TARGETS spdlog spdlog_header_only EXPORT spdlog DESTINATION "${CMAKE_INSTALL_LIBDIR}/spdlog")
+    install(TARGETS spdlog spdlog_header_only EXPORT spdlog DESTINATION "${CMAKE_INSTALL_LIBDIR}")
 
     if(NOT SPDLOG_FMT_EXTERNAL)
         install(DIRECTORY include/${PROJECT_NAME}/fmt/bundled/


gabime:
Merged. Thanks @DasRoteSkelett

orbea:
@DasRoteSkelett and @gabime Installing it to `spdlog/` was not strange, installing libraries to their own directory and installing them directly to the library directory are both common practices.

DasRoteSkelett:
@orbea can you give some reference? If I read 
https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s06.html
Correctly, it only states that *APPLICATIONS* may have a subdirectory. spdlog is a library though.

orbea:
@DasRoteSkelett No, but thanks for the reference, I haven't seen it before. I spoke from experience where other programs have done similar, but maybe its just a common mistake?

orbea:
@DasRoteSkelett How about the cmake files?

Here these are installed.
```
usr/lib64/spdlog/cmake/spdlogConfig.cmake
usr/lib64/spdlog/cmake/spdlogConfigTargets-release.cmake
usr/lib64/spdlog/cmake/spdlogConfigTargets.cmake
usr/lib64/spdlog/cmake/spdlogConfigVersion.cmake
```
But if we compare with fmt.
```
usr/lib64/cmake/fmt/fmt-config-version.cmake
usr/lib64/cmake/fmt/fmt-config.cmake
usr/lib64/cmake/fmt/fmt-targets-release.cmake
usr/lib64/cmake/fmt/fmt-targets.cmake
```
Would it be better if they were installed as `cmake/spdlog` instead of `spdlog/cmake`?

DasRoteSkelett:
The default in Linux is /usr/lib/<packagename>/cmake, but windows seems to prefer <LIBDIR>/cmake/<packagename>. This will usually be picked up by cmake, but other tools (i.e. openembedded) put those files in the wrong package. That is why we still prefer the first one, if you want to be interoperable with Windows, use the fmt format one... :-/

orbea:
Thanks for the details, that was informative! I suppose its fine the way it is now, but I don't use windows.

PR #1257, opened 2019-10-07
Uses GenerateExportHeaders from cmake to export only symbols needed e…

DasRoteSkelett:
…xternally

This should only affect the library part, and to be more specific, the shared
library part. Cmake allows to generate a special header file through
GenerateExportHeader()[1]. This defines - amongst other - a SPDLOG_EXPORT
Symbol, that does the right thing, depending on compiler / OS. On Windows,
default is, if I recall correctly, that all symbols are not exported while
Linux does it the other way around. So this patch sets default to *not* export
symbols except those marked SPDLOG_EXPORT. Behaviour with Windows and Linux
should be the same. Also, I removed SPDLOG_BUILD_SHARED in favour of the
standard BUILD_SHARED_LIBS, which is a global symbol. This *might* give some
issues when used as add_subdirectory(), but I am not sure.
Also I made Threads::Threads private, so a depending cmake package does not
need to have find_package(Threads) on.

Why all that? On x86_64, gcc 8.3.0 .so code size reduction is:
811320 bytes original vs. 532536 now (~65% of the original). And, I *guess*,
this will resolve the issue that it is not possible to use shared libs
with windows. But I cannot test it, I only have Linux.

[1] https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html

Signed-off-by: Matthias Schoepfer <matthias.schoepfer@ithinx.io>


gabime:
Thanks.
I tries using shared lib but got many link errors:
```console
 cmake .. -DBUILD_SHARED_LIBS=1
make -j
```

And got 
```
...
usr/bin/ld: example.cpp:(.text._ZN3fmt2v616visit_format_argINS0_13arg_formatterINS0_8internal12buffer_rangeIcEEEENS0_20basic_format_contextISt20back_insert_iteratorINS3_6bufferIcEEEcEEEEDTclfp_Li0EEEOT_RKNS0_16basic_format_argIT0_EE[_ZN3fmt2v616visit_format_argINS0_13arg_formatterINS0_8internal12buffer_rangeIcEEEENS0_20basic_format_contextISt20back_insert_iteratorINS3_6bufferIcEEEcEEEEDTclfp_Li0EEEOT_RKNS0_16basic_format_argIT0_EE]+0x11e2): undefined reference to `fmt::v6::internal::basic_data<void>::zero_or_powers_of_10_64'
/usr/bin/ld: example.cpp:(.text._ZN3fmt2v616visit_format_argINS0_13arg_formatterINS0_8internal12buffer_rangeIcEEEENS0_20basic_format_contextISt20back_insert_iteratorINS3_6bufferIcEEEcEEEEDTclfp_Li0EEEOT_RKNS0_16basic_format_argIT0_EE[_ZN3fmt2v616visit_format_argINS0_13arg_formatterINS0_8internal12buffer_rangeIcEEEENS0_20basic_format_contextISt20back_insert_iteratorINS3_6bufferIcEEEcEEEEDTclfp_Li0EEEOT_RKNS0_16basic_format_argIT0_EE]+0x13b5): undefined reference to `char fmt::v6::internal::thousands_sep_impl<char>(fmt::v6::internal::locale_ref)'
/usr/bin/ld: example.cpp:(.text._ZN3fmt2v616visit_format_argINS0_13arg_formatterINS0_8internal12buffer_rangeIcEEEENS0_20basic_format_contextISt20back_insert_iteratorINS3_6bufferIcEEEcEEEEDTclfp_Li0EEEOT_RKNS0_16basic_format_argIT0_EE[_ZN3fmt2v616visit_format_argINS0_13arg_formatterINS0_8internal12buffer_rangeIcEEEENS0_20basic_format_contextISt20back_insert_iteratorINS3_6bufferIcEEEcEEEEDTclfp_Li0EEEOT_RKNS0_16basic_format_argIT0_EE]+0x13ed): undefined reference to `fmt::v6::internal::basic_data<void>::zero_or_powers_of_10_32'
/usr/bin/ld: example.cpp:(.text._ZN3fmt2v616visit_format_argINS0_13arg_formatterINS0_8internal12buffer_rangeIcEEEENS0_20basic_format_contextISt20back_insert_iteratorINS3_6bufferIcEEEcEEEEDTclfp_Li0EEEOT_RKNS0_16basic_format_argIT0_EE[_ZN3fmt2v616visit_format_argINS0_13arg_formatterINS0_8internal12buffer_rangeIcEEEENS0_20basic_format_contextISt20back_insert_iteratorINS3_6bufferIcEEEcEEEEDTclfp_Li0EEEOT_RKNS0_16basic_format_argIT0_EE]+0x15a5): undefined reference to `fmt::v6::internal::basic_data<void>::zero_or_powers_of_10_
...```

DasRoteSkelett:
You are not using external fmt, right? I will check...

DasRoteSkelett:
Hi @gabime , 
Thanks for the feedback, I think, I addressed the issues with another commit. See if you still think, it is okay for you. 

Regards,

   Matthias

gabime:
Thanks again!
format.h and core.h are part of the bundled  fmt and better not touched. Otherwise this won’t work with external or different fmt version.

Another issue is that I still want spdlog to work without CMake(e.g when using the header only version or other build like meson), but depending on the generated spdlog_export.h makes this impossible.
Maybe the approach would be put our own version spdlog_export.h into the repo and make it work with proper ifdefs on both windows and unix?

DasRoteSkelett:
Hi @gabime ,
regarding fmt: have you considered including it with add_subdirectory()? Then, you do not need to touch fmt at all! To me, that sounds like the more propper approach.

I am looking into the header_only version. I do not know if meson has a similar approach, i do not know meson at all. I would strongly suggest to stick to one build system and also *not* to re-engineer generate_export_header. It is there for a reason. It covers basically all supported compilers on all supported platforms.

gabime:
> regarding fmt: have you considered including it with add_subdirectory()? Then, you do not need to touch fmt at all! To me, that sounds like the more propper approach.

Yes, maybe in future versions

> I would strongly suggest to stick to one build system and also not to re-engineer generate_export_header. 

In any case, header-only without cmake should work

DasRoteSkelett:
> Yes, maybe in future versions

If you would consider it, I would change my PR in that direction??

> In any case, header-only without cmake should work

Okay, I will think of a solution. It would be nice to get this to work, since the code size reduction is significant and (have you tested?) you can probably now have shared libs on windows.

DasRoteSkelett:
Hi @gabime ,

I reworked my PR. Now, it should be less intrusive to meson and header only use cases.It would be very nice, if you could have a look at it. As stated, this reduces the binary size significantly and should solve the problem not to be able to have shared libs with windows (though, I have not tested this).

gabime:
Thanks. What's the purpose ```SPDLOG_USE_EXPORT_HEADER``` ? shouldn't it always be enabled?

DasRoteSkelett:
Hi @gabime 

> Thanks. What's the purpose SPDLOG_USE_EXPORT_HEADER ? shouldn't it always be enabled?

It serves as a Guard, it gets defined iff generate_export_header() from the cmake is used. If not, the more or less "default" or original behaviour is selected. 

Regards,

  Matthias

gabime:
ok, but why is there an option not to generate it? it should always be used

DasRoteSkelett:
Hi @gabime 

> In any case, header-only without cmake should work

That is, why a guard is needed here. If you are not building with cmake, the spdlog_export.h will *not* get generated. With this guard, it does not break meson build nor header-only usage. I was under the impression, that thats exactly what you wanted ?!

gabime:
I still think that  having own version of spdlog_export.h without relying on cmake is a preffered approach. It will avoid all those ```#ifdef SPDLOG_USE_EXPORT_HEADER``` all over the place.

DasRoteSkelett:
Hi @gabime 

> I still think that having own version of spdlog_export.h without relying on cmake is a preffered approach. It will avoid all those #ifdef SPDLOG_USE_EXPORT_HEADER all over the place.

Well, and I am still thinking, you are not right in this. If you want, I can hide this #ifdef guard stuff in a separate header. But having a headerfile that does it all will not scale. The reason to use cmake ("cross-platform make") is, to make your library compile over a large variety of OS and Compiler Combinations in almost arbitrary combinations, and that is what is does really well. So I would suggest to use it on this topic, instead of writing a *really* long and ugly header file with lots of #ifdefs  that checks for OS (Windows, Linux, MacOS, Android, ...) *and* Compiler (MSVC, GCC, clang, cygwin, ...). Quite frankly, I think, some of your code could probably be reduced by relying on cmake even more. But that is another topic. 

You can always decide to decline this merge request, but since it seems it fixes dll builds for windows *and* reduces binary size significantly (and therefore is a performance boost), it really adds to spdlog. 

Regards,
   Matthias

DasRoteSkelett:
Hi @gabime ,

I rebased the changes, so they once again can be cleanly applied. Please consider this for merging, as it reduces the resulting binary size quite a bit for the cmake build and allows to build a proper dll with windows. 

If you do not want this to get merged, then please close the pull request. I spend some considerable time with this, and if you are not planning to merge this, it would be fair to let me know that I do not waste even more time with this. If you have any requests, that I can comply (i.e. *not* the own version of spdlog_exports.h), please let me know. 

Regards,

   Matthias

gabime:
I was hoping to at least use this PR as a reference for any future work on this.
But if you prefer I can close it - I don't know when I will have time to tackle this.

PR #1270, opened 2019-10-18
Fix build failure on clang 7 with libc++

jktjkt:
Unlike the GNU C++ STL, there's no implicit include for <array> in this one, apparently.


gabime:
Thanks @jktjkt !

Issue #1278, opened 2019-10-22
Allow using ansicolor sinks on Windows


tambry:
[Windows 10 natively supports ANSI escape sequences](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) (enabled by `ENABLE_VIRTUAL_TERMINAL_PROCESSING`).

Thus it may be desirable to use an `ansicolor_sink` instead of a `wincolor_sink`. Unfortunately the sinks compiled [are conditional on `_WIN32` being defined](https://github.com/gabime/spdlog/blob/dae4f9fef60b92a9d1ada158f8f1a970eae7141c/src/color_sinks.cpp#L15), making it impossible to use the ansicolor sinks.

Would it be possible to provide an additional flag `SPDLOG_FORCE_ANSICOLOR` in CMake that would force the ansicolor sinks to be built?

gabime:
What is the advantage ansi colors over the existing windows color sink?

tambry:
Enables using VT100 sequences for formatting in log messages. E.g. `spdlog::info("\u001b[31mRed text\u001b[0mBack to normal");`.

I haven't measured, but probably also slightly better performance due to not having to make system calls for colouring.

Cyphall:
On top of that, VT100 also allows to use a full 2x24-bit color range (foreground & background) instead of the 2x4-bit predefined colors:
https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#extended-colors

PR #1279, opened 2019-10-22
#1278: Allow forcing ansicolor sinks on Windows

tambry:
See #1278 for rationale.


gabime:
Thanks. However, I am not sure there is enough demand for this to be added. I prefer to wait and see.

PR #1283, opened 2019-10-24
Minor cmake code duplication improvement

bencsikandrei:
Create a macro to add tests for both static/header only.

The only differences between these two are the libraries they link
with and the target names. 
There might be a reason to keep both, but it seemed like a nice opportunity 
for a refactor.

Created the simple macro:

_spdlog_prepare_test(<target> <spdlog_lib>)

which does the work. 

Signed-off-by: Andrei-Florin BENCSIK <andrei.bencsik@gmail.com>


gabime:
Thanks. what’s the advantage of macro here instead of a regular function?

bencsikandrei:
Hello,

In this case I think there isn't much of a difference since it's not setting anything inside the macro/function nor does it return() anything.
My understanding is that macros don't create a new scope, so all variables are string replaced (much like in C++) when the macro is called - also anything created inside is visible outside the scope, unlike functions which create a new scope. The return() statement inside a function returns from that particular function whilst from a macro it returns from the parent scope (as in C++, kind of).

I don't think there's much of an advantage one way or the other, they should work the same, but if you prefer a function I'm on board with that and I'll happily modify it.

gabime:
Since the rest of spdlog’s cmake scripts are already using functions, lets use a function here as well  for the sake of consistency.

bencsikandrei:
Done!
I made another commit, but I believe it should be squashed from github (with the commit message of your choice).
Otherwise I can do it myself and rephrase the messages as well.

gabime:
It doesn't pass the tests unfortunately.

bencsikandrei:
Because VIM ate my 'end' from 'endfunction' :(

Issue #1290, opened 2019-10-29
backtrace support produces confusing results with multiple loggers


jktjkt:
This might be an A/B question. Here's the A-part:

I'm currently using several loggers, all of them sharing one sink, to provide something like "categories". I have one category for `io`, one for `interface-to-an-external-lib`, another one for `processing`, etc. Parts of my code which perform IO log to the `io` logger, etc.

I would like to use the backtrace subsystem to show additional details when an error has happened. During the course of normal operation, nobody cares about a hardcopy of the actual IO (it can be about 200 kB/s for me), but when the code logs an error about a malformed communication, then the IO details become important. The problem with backtraces is that they work per-logger. When I dump the messages, I get all IO messages together, followed by the "processing" messages, etc. I would like to have all messages across all loggers in order. I think that this could be solved by a "proxy logger" which just feeds all messages to some other logger, and the other logger preserves the original logger's name.

Here's the B-part:

What I'm really after is not necessarily multiple loggers. I just want different parts of my code to "tag" the messages as they are produced. If there's a better way then multiple loggers, I'll be happy to do that. I think I still need to set the log levels "per component", though, just because of the volume of the trace messages...

gabime:
>  I think that this could be solved by a "proxy logger" which just feeds all messages to some other logger, and the other logger preserves the original logger's name.

In current implementation,I don't see a clear solution for this (perhaps define a dedicated custom “backtrace” sink that forwards all its messages to a dedicated backtrace logger?).

> What I'm really after is not necessarily multiple loggers. I just want different parts of my code to "tag" the messages as they are produced. If there's a better way..

That’s the recommended way - To have a logger per component/subsytem (they all can share the same set of sinks of course).

PR #1292, opened 2019-10-30
improve systemd journald support

### LINKED ISSUE #1289: systemd-journald sink does not log logger name
My application uses several loggers. When I use an `spdlog::sinks::ansicolor_stderr_sink_mt`, the output contains (among other things) the logger's name. When I switch to using `spdlog::sinks::systemd_sink_mt`, the logger name is not available anymore.

I think that this is because the systemd/journald sink does not use `spdlog::base_sink::formatter_` at all even though it inherits from `base_sink`. I suspect that this might have been due to the fact that the default formatter outputs stuff such as date/time/priority which is available via other means with journald. The logger name, however, is effectively lost this way.

jktjkt:
These three patches which make it easier to use `systemd`/`journald`'s structured logging. When everything is in place, one can, for example, watch the log stream via this user-friendly :) command:

```
journlctl -f -u my-app.service \
  PRIORITY=1 PRIORITY=2 PRIORITY=3 PRIORITY=4 \
    SYSLOG_IDENTIFIER=only-important-messages-are-shown-now \
 + \
 PRIORITY=1 PRIORITY=2 PRIORITY=3 PRIORITY=4 PRIORITY=5 PRIORITY=6 PRIORITY=7 \
    SYSLOG_IDENTIFIER=I-want-details-from-this
```
That command will follow journal stream from the `my-app.service`. The logger whose name is `only-important-messages-are-shown-now` is rather heavily filtered, but everything from a logger called `I-want-details-from-this` is shown. These options are [documented](https://www.freedesktop.org/software/systemd/man/journalctl.html) in `journalctl`'s manpage. Sadly, there is neither range-checking nor [negative matching](https://github.com/systemd/systemd/issues/2720) in there.

I'm doing this mostly for use cases described in #1290 ...

##  ~~journald: wider range of log severities~~

~~The previous code made it impossible to distinguish between the `debug` and `trace` severities. So instead of trying hard to preserve the equivalence of `spdlog::level::debug` with `LOG_DEBUG`, which is -- in the end -- just an arbitrary name, make sure that functionality is not lost.~~

[Removed per explicit request]( https://github.com/gabime/spdlog/pull/1292#discussion_r340777258); here's how to implement this:
```c++
#include <spdlog/sinks/systemd_sink.h>

template<typename Mutex>
class journald_sink : public spdlog::sinks::systemd_sink<Mutex> {
public:
    journald_sink()
    {
        this->syslog_levels_ = {/* spdlog::level::trace      */ LOG_DEBUG,
              /* spdlog::level::debug      */ LOG_INFO,
              /* spdlog::level::info       */ LOG_NOTICE,
              /* spdlog::level::warn       */ LOG_WARNING,
              /* spdlog::level::err        */ LOG_ERR,
              /* spdlog::level::critical   */ LOG_CRIT,
              /* spdlog::level::off        */ LOG_ALERT};
    }
};
```

##  journald: structured output for logger's name

Previously, the logger name was effectively lost. There were two choices on how to add it:

- Via a formatter, which would mean that `journalctl` would not be able to filter against that. That would be suboptimal.

- As a "syslog identifier". This means that `journalctl` will, by default, stop showing the daemon's executable name and replace that via the logger name. The PID is still shown, and if one would like to go back to the previous behavior, it is still possible via `journalctl -o with-unit`.

I think that the second option is strictly better than the first one.

fixes #1289

##  journald: fix source file location

This is what my manpage says, and what the [original blog post](http://0pointer.de/blog/projects/journal-submit.html) says as well.

### DIFF
--- a/include/spdlog/sinks/systemd_sink.h
+++ b/include/spdlog/sinks/systemd_sink.h
@@ -8,6 +8,9 @@
 #include "spdlog/details/synchronous_factory.h"
 
 #include <array>
+#ifndef SD_JOURNAL_SUPPRESS_LOCATION
+#define SD_JOURNAL_SUPPRESS_LOCATION
+#endif
 #include <systemd/sd-journal.h>
 
 namespace spdlog {
@@ -63,7 +66,7 @@ protected:
         else
         {
             err = (sd_journal_send)("MESSAGE=%.*s", static_cast<int>(length), msg.payload.data(), "PRIORITY=%d", syslog_level(msg.level),
-                "SOURCE_FILE=%s", msg.source.filename, "SOURCE_LINE=%d", msg.source.line, "SOURCE_FUNC=%s", msg.source.funcname, nullptr);
+                "CODE_FILE=%s", msg.source.filename, "CODE_LINE=%d", msg.source.line, "CODE_FUNC=%s", msg.source.funcname, nullptr);
         }
 
         if (err)
--- a/include/spdlog/sinks/systemd_sink.h
+++ b/include/spdlog/sinks/systemd_sink.h
@@ -61,11 +61,13 @@ protected:
         {
             // Note: function call inside '()' to avoid macro expansion
             err = (sd_journal_send)(
-                "MESSAGE=%.*s", static_cast<int>(length), msg.payload.data(), "PRIORITY=%d", syslog_level(msg.level), nullptr);
+                "MESSAGE=%.*s", static_cast<int>(length), msg.payload.data(), "PRIORITY=%d", syslog_level(msg.level),
+                "SYSLOG_IDENTIFIER=%.*s", static_cast<int>(msg.logger_name.size()), msg.logger_name.data(), nullptr);
         }
         else
         {
             err = (sd_journal_send)("MESSAGE=%.*s", static_cast<int>(length), msg.payload.data(), "PRIORITY=%d", syslog_level(msg.level),
+                "SYSLOG_IDENTIFIER=%.*s", static_cast<int>(msg.logger_name.size()), msg.logger_name.data(),
                 "CODE_FILE=%s", msg.source.filename, "CODE_LINE=%d", msg.source.line, "CODE_FUNC=%s", msg.source.funcname, nullptr);
         }
 


include/spdlog/sinks/systemd_sink.h
```diff
@@ -58,12 +58,14 @@ class systemd_sink : public base_sink<Mutex>
...
-                "MESSAGE=%.*s", static_cast<int>(length), msg.payload.data(), "PRIORITY=%d", syslog_level(msg.level), nullptr);
+                "MESSAGE=%.*s", static_cast<int>(length), msg.payload.data(), "PRIORITY=%d", syslog_level(msg.level),
+                "SYSLOG_IDENTIFIER=%.*s", static_cast<int>(msg.logger_name.size()), msg.logger_name.data(), nullptr);
         }
         else
         {
             err = (sd_journal_send)("MESSAGE=%.*s", static_cast<int>(length), msg.payload.data(), "PRIORITY=%d", syslog_level(msg.level),
-                "SOURCE_FILE=%s", msg.source.filename, "SOURCE_LINE=%d", msg.source.line, "SOURCE_FUNC=%s", msg.source.funcname, nullptr);
+                "SYSLOG_IDENTIFIER=%.*s", static_cast<int>(msg.logger_name.size()), msg.logger_name.data(),
+                "CODE_FILE=%s", msg.source.filename, "CODE_LINE=%d", msg.source.line, "CODE_FUNC=%s", msg.source.funcname, nullptr);
```
gabime: The systemd man says that the  CODE_xxx  fields are added  automatically.. so why add them again?

include/spdlog/sinks/systemd_sink.h
```diff
@@ -25,12 +25,12 @@ class systemd_sink : public base_sink<Mutex>
     //
     systemd_sink()
         : syslog_levels_{{/* spdlog::level::trace      */ LOG_DEBUG,
-              /* spdlog::level::debug      */ LOG_DEBUG,
-              /* spdlog::level::info       */ LOG_INFO,
+              /* spdlog::level::debug      */ LOG_INFO,
+              /* spdlog::level::info       */ LOG_NOTICE,
```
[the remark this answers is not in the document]
gabime: In opinion distinguishing between trace and debug is less important than confusing debug with info..

include/spdlog/sinks/systemd_sink.h
gabime: It probably depends on the exact version of the lib

include/spdlog/sinks/systemd_sink.h
```diff
@@ -8,6 +8,7 @@
 #include "spdlog/details/synchronous_factory.h"
 
 #include <array>
+#define SD_JOURNAL_SUPPRESS_LOCATION
```
gabime: #ifndef SD_JOURNAL_SUPPRESS_LOCATION
#define SD_JOURNAL_SUPPRESS_LOCATION
#endif

gabime: otherwise we can get ugly warnings..

jktjkt: done

include/spdlog/sinks/systemd_sink.h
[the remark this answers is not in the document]
jktjkt: OK; I disagree with this, but it's your project. For the record, and if someone gets here looking for how to fix this, it's pretty trivial to override this in a custom sink (thanks for making that mapping `protected`!):

```c++
#include <spdlog/sinks/systemd_sink.h>

template<typename Mutex>
class journald_sink : public spdlog::sinks::systemd_sink<Mutex> {
public:
    journald_sink()
    {
        this->syslog_levels_ = {/* spdlog::level::trace      */ LOG_DEBUG,
              /* spdlog::level::debug      */ LOG_INFO,
              /* spdlog::level::info       */ LOG_NOTICE,
              /* spdlog::level::warn       */ LOG_WARNING,
              /* spdlog::level::err        */ LOG_ERR,
              /* spdlog::level::critical   */ LOG_CRIT,
              /* spdlog::level::off        */ LOG_ALERT};
    }
};
```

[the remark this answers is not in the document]
jktjkt: (And I forgot to clearly state that I removed this commit from the PR so that the rest can be merged.)

gabime:
Thanks @jktjkt . Merged.

PR #1293, opened 2019-10-30
Add new define to choose a less verbose version of SPDLOG_LOGGER_CALL

macedot:
Creates a SPDLOG_LOGGER_CALL_LITE define to switch  to a LITE version of SPDLOG_LOGGER_CALL, without file, line and function parameters on its call.

### DIFF
--- a/include/spdlog/spdlog.h
+++ b/include/spdlog/spdlog.h
@@ -285,12 +285,21 @@ inline void critical(wstring_view_t fmt, const Args &... args)
 // SPDLOG_LEVEL_OFF
 //
 
+#ifdef SPDLOG_LOGGER_CALL_LITE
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. Whats the purpose? I don't think performance would be (almost) any better.

macedot:
Main porpoise was to replicate standard calls behavior, since those don't include the mentioned parameters. Besides/thus, its way less verbose. 🙂

gabime:
This macro is primary for use by the other macros and not directly, so verbosity is ok here.

macedot:
Indeed. My goal was to reduce verbosity on the other macros. Although one could call this "low-level" macro directly, I was aiming to the "high-level" ones. Cheers any way.

PR #1294, opened 2019-11-02
Improve log macros

pck:
Hello. I think the log macros should be changed to compile with "&my_logger" expressions. Thanks.


gabime:
Thanks @pck Merged.

PR #1296, opened 2019-11-04
Correctly guard SetHandleInformation API call

sylveon:
`__cplusplus_winrt` only detected C++/CX (which can be used without compiling for UWP, SetHandleInformation would be available in those cases), and did not detect native UWP C++. This patch fixes that by using the WINAPI_FAMILY_PARTITION macro in the Windows SDK headers in the same way those headers remove SetHandleInformation in UWP builds.

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -130,7 +130,7 @@ SPDLOG_INLINE void prevent_child_fd(FILE *f)
 {
 
 #ifdef _WIN32
-#if !defined(__cplusplus_winrt)
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)
     auto file_handle = reinterpret_cast<HANDLE>(_get_osfhandle(::_fileno(f)));
     if (!::SetHandleInformation(file_handle, HANDLE_FLAG_INHERIT, 0))
         SPDLOG_THROW(spdlog_ex("SetHandleInformation failed", errno));


sylveon:
How would you feel about adding a `[[maybe_unused]]` to silence warnings on `/W4` with MSVC?

gabime:
> How would you feel about adding a [[maybe_unused]] to silence warnings on /W4 with MSVC?

Which warnings specifically? The problem is spdlog is c++11

gabime:
On second thought.What if windows sdk is not installed? will this break?

sylveon:
Without the windows sdk, sethandleinformation or any windows header isn't
available at all. The warning is C4100, and maybe_unused could be
conditionally enabled using the preprocessor (or you could manually disable
the warning using pragmas)

gabime:
> could manually disable the warning using pragmas

Before which line?

gabime:
> Without the windows sdk, sethandleinformation or any windows header isn't
available at all.

Are you sure? I dont see any mention of this [here](https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-sethandleinformation)

sylveon:
Around the method implementation's body since the warning is related to a
parameter not being used.

gabime:
You mean FILE *f not being used if ```WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM``` is not satisfied ?

sylveon:
Correct, hence maybe_unused

gabime:
The real problem is that this function would silently..

gabime:
[Here](57085c892f9cc4cd36f9738109864a78ec7d7e3f) is my shot at it. If ```SPDLOG_PREVENT_CHILD_FD``` is not defined(default) - then the entrire function will not be defined.

If ```SPDLOG_PREVENT_CHILD_FD``` is defined under winrt - then the lib won't compile - which is  way better than failing silently.

sylveon:
Sounds like a valid option.

cngege:
- 经过测试，spdlog的文件日志部分是支持在UWP程序中运行的，如果你的程序不能如期工作，请检查，路径部分是否正确，并存在，检查 `spdlog::flush_on(spdlog::level::info);` 是否正确设置，祝你好运
- After testing, it is confirmed that the file logging component of spdlog supports running in UWP programs. If your program does not work as expected, please check if the path is correct and exists, and verify if `spdlog::flush_on(spdlog::level::info);` is set correctly. Good luck!

PR #1304, opened 2019-11-06
"#include" <spdlog/.*> instead of "spdlog/.*"

RedDwarf69:
The meaning of using quotes to #include is implementation defined, so it
may or not may be what we want. At least POSIX
(https://pubs.opengroup.org/onlinepubs/9699919799/utilities/c99.html)
says: "headers whose names are enclosed in double-quotes ( "" ) shall be
searched for first in the directory of the file with the #include line",
so not what we want since "spdlog" ends up twice in the path.

For example, in my Fedora 31 system:
```
$ cat test.cpp
#include <spdlog/spdlog.h>
$ strace -f -e file g++ -c -o /dev/null test.cpp -Ilocal/include/ 2>&1 | grep 'spdlog/common.h'
[pid 14210] openat(AT_FDCWD, "local/include/spdlog/spdlog/common.h", O_RDONLY|O_NOCTTY) = -1 ENOENT (No such file or directory)
[pid 14210] openat(AT_FDCWD, "local/include/spdlog/common.h", O_RDONLY|O_NOCTTY) = 4
[pid 14210] openat(AT_FDCWD, "local/include/spdlog/details/spdlog/common.h", O_RDONLY|O_NOCTTY) = -1 ENOENT (No such file or directory)
[pid 14210] openat(AT_FDCWD, "local/include/spdlog/sinks/spdlog/common.h", O_RDONLY|O_NOCTTY) = -1 ENOENT (No such file or directory)
$
```

### DIFF
--- a/include/spdlog/async.h
+++ b/include/spdlog/async.h
@@ -14,9 +14,9 @@
 // This is because each message in the queue holds a shared_ptr to the
 // originating logger.
 
-#include "spdlog/async_logger.h"
-#include "spdlog/details/registry.h"
-#include "spdlog/details/thread_pool.h"
+#include <spdlog/async_logger.h>
+#include <spdlog/details/registry.h>
+#include <spdlog/details/thread_pool.h>
 
 #include <memory>
 #include <mutex>
--- a/include/spdlog/async_logger-inl.h
+++ b/include/spdlog/async_logger-inl.h
@@ -4,11 +4,11 @@
 #pragma once
 
 #ifndef SPDLOG_HEADER_ONLY
-#include "spdlog/async_logger.h"
+#include <spdlog/async_logger.h>
 #endif
 
-#include "spdlog/sinks/sink.h"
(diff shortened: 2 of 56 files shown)


gabime:
How about using relative includes e.g. include “../details/os.h” ?

Seems even more accurate to me

RedDwarf69:
Personally I would say

Pros
---
- In POSIX/gcc it's safer since there is no risk of a different header being included by error.

But it's only slighty safer. All the headers are "namespaced" with `spdlog/` anyway, the risk is minimum.

Cons
----
- It's still implementation defined. God knows what will happen in non POSIX/gcc systems
- While probably supported in most places, I don't think the meaning of `..` as "parent directory" is standardized.
- In POSIX/gcc, being so safe, it doesn't allow people to override the headers if they want to (mocking?)

So cons wins.

gabime:
Merged. Thanks @RedDwarf69

PR #1305, opened 2019-11-07
prevent clang-tidy warnings while build

ClausKlein:
note: each use of assert() produce one!

i.e.:
tests/test_pattern_formatter.cpp:14:95: warning: do not implicitly decay an array into a pointer;
consider using gsl::array_view or an explicit cast instead [cppcoreguidelines-pro-bounds-array-to-pointer-decay]
    oss_logger.set_formatter(std::unique_ptr<spdlog::formatter>(new spdlog::pattern_formatter(args...)));
                                                                                              ^

### DIFF
--- a/include/spdlog/common-inl.h
+++ b/include/spdlog/common-inl.h
@@ -15,12 +15,12 @@ static const char *short_level_names[] SPDLOG_SHORT_LEVEL_NAMES;
 
 SPDLOG_INLINE string_view_t &to_string_view(spdlog::level::level_enum l) SPDLOG_NOEXCEPT
 {
-    return level_string_views[l];
+    return level_string_views[l];   // NOLINT
 }
 
 SPDLOG_INLINE const char *to_short_c_str(spdlog::level::level_enum l) SPDLOG_NOEXCEPT
 {
-    return short_level_names[l];
+    return short_level_names[l];    // NOLINT
 }
 
 SPDLOG_INLINE spdlog::level::level_enum from_str(const std::string &name) SPDLOG_NOEXCEPT
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -62,7 +62,7 @@
 #endif
 
 #ifndef SPDLOG_FUNCTION
-#define SPDLOG_FUNCTION __FUNCTION__
+#define SPDLOG_FUNCTION static_cast<const char *>(__FUNCTION__) // NOLINT
 #endif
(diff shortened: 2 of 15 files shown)


gabime:
Thanks. However I prefer fixing those warnings instead of suppressing. What is the config of the tidy? Can you post the config file?

ClausKlein:
##################################################
# begin of config part
# see https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/
# and https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py
#
### checkAllHeader:='include/spdlog/[acdlstv].*'
checkAllHeader?='include/spdlog/[^f].*'
checkAllHeader?='$(CURDIR)/.*'

# NOTE: there are many errors with boost::test, doctest, catch test framework! CK
### CHECKS:='-cppcoreguidelines-avoid-c-arrays,-modernize-avoid-c-arrays'
CHECKS?='-*,cppcoreguidelines-*,-cppcoreguidelines-pro-type-vararg,-cppcoreguidelines-special-member-functions,-cppcoreguidelines-avoid-c-arrays'
CHECKS?='-*,portability-*,readability-*'
CHECKS?='-*,misc-*,boost-*,cert-*,misc-unused-parameters'

`run-clang-tidy.py -header-filter='include/spdlog/[^f].*' -checks='-*,cppcoreguidelines-*,-cppcoreguidelines-pro-type-vararg,-cppcoreguidelines-special-member-functions,-cppcoreguidelines-avoid-c-arrays' | tee run-clang-tidy.log 2>&1
`

ClausKlein:
You may try this version:

https://github.com/ClausKlein/spdlog/blob/develop/.clang-tidy

ClausKlein:
> Thanks. However I prefer fixing those warnings instead of suppressing. 

It would be possible to fix some of the warnings with static_cast<>.
But many of them are problems about pointer Arithmetik und buffer handling!

It's advisable to first prevent the warnings related to the usage of spdlog:

`#define SPDLOG_ACTIVE_LEVEL 0 // NOLINT`
`#define SPDLOG_FUNCTION static_cast<const char *>(__func__) // NOLINT`

see too: https://github.com/ClausKlein/Serial-TUN/blob/develop/tun-driver.h

gabime:
I fixed some of the warnings. Thanks.

PR #1306, opened 2019-11-08
struct tm doesn't have tm_gmtoff in ISO C

lancesun:
The field is a BSD and GNU extension (https://www.gnu.org/software/libc/manual/html_node/Broken_002ddown-Time.html)


gabime:
Thanks. I fixed it little bit different here: ff3e6c7248a0bfb1d44e72fd0f822320178263bc

PR #1307, opened 2019-11-08
Add ringbuffer sink

eudoxos:
Ringbuffer sink keeps user-given number of most recent log messages in
memory and returns them upon request (using the ringbuffer_sink::last
method). This can be useful for e.g. remote debugging of a running app.

### DIFF
--- a/include/spdlog/details/circular_q.h
+++ b/include/spdlog/details/circular_q.h
@@ -75,7 +75,14 @@ public:
     // Return number of elements actually stored
     size_t size() const
     {
-        return (tail_ - head_) % max_items_;
+        if (tail_ > head_)
+        {
+           return (tail_ - head_) % max_items;
+        }
+        else
+        {
+            return max_items - (head_ - tail_ ) % max_items;
+        }
     }
 
     // Return const reference to item by index.
--- a/include/spdlog/details/circular_q.h
+++ b/include/spdlog/details/circular_q.h
@@ -75,7 +75,7 @@ public:
     // Return number of elements actually stored
     size_t size() const
     {
-        if (tail_ > head_)
+        if (tail_ >= head_)
         {
            return tail_ - head_;
         }
--- a/include/spdlog/details/circular_q.h
+++ b/include/spdlog/details/circular_q.h
@@ -77,11 +77,11 @@ public:
     {
         if (tail_ > head_)
         {
-           return (tail_ - head_) % max_items_;
+           return tail_ - head_;
         }
         else
         {
-            return max_items_ - (head_ - tail_ ) % max_items_;
+            return max_items_ - (head_ - tail_ );
         }
     }
 
--- a/include/spdlog/details/circular_q.h
+++ b/include/spdlog/details/circular_q.h
@@ -77,11 +77,11 @@ public:
     {
         if (tail_ > head_)
         {
-           return (tail_ - head_) % max_items;
+           return (tail_ - head_) % max_items_;
         }
         else
         {
-            return max_items - (head_ - tail_ ) % max_items;
+            return max_items_ - (head_ - tail_ ) % max_items_;
         }
     }
 
--- a/include/spdlog/details/circular_q.h
+++ b/include/spdlog/details/circular_q.h
@@ -72,6 +72,20 @@ public:
         return v_[head_];
     }
 
+    // Return number of elements actually stored
+    size_t size() const
+    {
+        return (tail_ - head_) % max_items_;
+    }
+
+    // Return const reference to item by index.
+    // If index is out of range 0…size()-1, the behavior is undefined.
+    const T &at(size_t i) const
+    {
+        assert(i < size());
+        return v_[(head_+ i) % max_items_];
+    }
+
     // Pop item from front.
     // If there are no elements in the container, the behavior is undefined.
     void pop_front()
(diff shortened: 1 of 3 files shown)


gabime:
Thanks. Seems it can already be implemented with the [backtrace](https://github.com/gabime/spdlog#backtrace-support) feature (maybe need to add ```backtrace::last(size_t lim)``` method to it).

eudoxos:
> Thanks.
> spdlog avoids 3rd party dependencies, so please use [spdlog::details::circualr_q]
Done. I had to add `circular_q::size()` and `circular_q::at(size_t)` for random access without modifying the data.

include/spdlog/details/circular_q.h
```diff
@@ -72,6 +72,20 @@ class circular_q
         return v_[head_];
     }
 
+    // Return number of elements actually stored
+    size_t size() const
+    {
+        return (tail_ - head_) % max_items_;
```
gabime: I think it breaks when head_ > tail_ (can happen)

include/spdlog/sinks/ringbuffer_sink-inl.h
```diff
@@ -0,0 +1,46 @@
...
+#include "spdlog/common.h"
+#include "spdlog/details/os.h"
+
+namespace spdlog {
+namespace sinks {
+
+template<typename Mutex>
+SPDLOG_INLINE ringbuffer_sink<Mutex>::ringbuffer_sink(size_t buf_size)
+{
+    buf=details::circular_q<std::string>(buf_size);
```
gabime: Using circular_q<log_msg_buffer> instead of atring will improve performance:
1: Avoids allocation of string in each message
2. format() can be avoided until really needed (in the last(size_t lim) method)
 see https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/details/backtracer.h as an example

eudoxos: I tried first with `log_msg` but deferred formatting was garbage as presumably some of the data were out of scope already. That's why I went the `std::string` way with perf impact. If I use `log_msg_buffer`, this issue will not happen? If so, good to hear and I will be happy to change that.

include/spdlog/details/circular_q.h
eudoxos: Sorry, a stupid mistake. Should be solved now.

gabime: no need to ``` % max_items``` and will give wrong result if if tail_ == head_(i.e. if empty it will give max_items_) 

Here is a fixed version:
```c++
size_t size() const
    {
        if (tail_ >= head_)
        {
           return (tail_ - head_);
        }
        else
        {
            return max_items_ - (head_ - tail_ );
        }
    }
```

include/spdlog/sinks/ringbuffer_sink-inl.h
gabime: > If I use log_msg_buffer, this issue will not happen?

No, it won't happen. That is the sole purpose of ```log_msg_buffer```. To deep copy log_messages off the stack into it's internal buffer.

include/spdlog/details/circular_q.h
```diff
@@ -72,6 +72,27 @@ class circular_q
...
+    {
+        if (tail_ > head_)
+        {
+           return (tail_ - head_) % max_items_;
+        }
+        else
+        {
+            return max_items_ - (head_ - tail_ ) % max_items_;
+        }
+    }
```
gabime: no need to % max_items and will give wrong result if if tail_ == head_(i.e. if empty it will give max_items_)

Here is a fixed version:

size_t size() const
    {
        if (tail_ >= head_)
        {
           return (tail_ - head_);
        }
        else
        {
            return max_items_ - (head_ - tail_ );
        }
    }

eudoxos: Thanks for patience. Fixed.

include/spdlog/details/circular_q.h
```diff
@@ -72,6 +72,27 @@ class circular_q
         return v_[head_];
     }
 
+    // Return number of elements actually stored
+    size_t size() const
+    {
+        if (tail_ > head_)
```
gabime: should be “if(tail_ >= head_)” to cover the empty queue case

include/spdlog/sinks/ringbuffer_sink-inl.h
```diff
@@ -0,0 +1,46 @@
...
+
+template<typename Mutex>
+SPDLOG_INLINE void ringbuffer_sink<Mutex>::sink_it_(const details::log_msg &msg)
+{
+    buf_.push_back(details::log_msg_buffer{msg});
+}
+
+template<typename Mutex>
+SPDLOG_INLINE std::vector<std::string> ringbuffer_sink<Mutex>::last(size_t lim)
+{
```
gabime: Note this will return in a confusing order (the head is not at position 0). I think what we need here is way to iterate over the items in the actual order.

eudoxos: As far as I see, you are mistaken... `circular_q::at(i)` will return item at `(head_+i)%max_items_` in the storage vector.

gabime: Right. I didn't notice correctly.

PR #1311, opened 2019-11-11
Remove redundant semicolon

inhzus:
Remove redundant semicon of macro SPDLOG_LOGGER_CALL which causes clang diagnoses empty statement.


gabime:
Thanks @inhzus

Issue #1297, opened 2019-11-04
Please add clipping alignment


eudoxos:
Alignment specifiers in patterns allow one to align left, right or centered. However, if the string being aligned is longer than alignment, it will stick out. Would it be possible to request clipping of the resulting string, e.g. via `%!12` or some modifier other than `-` and `=` already in use? This would be useful to e.g show only first 12 characters of base filename or logger name (so that messages stay aligned). Up to consideration whether the clipped variant should also have left/centered/right versions, I'd be fine with left-aligned (right-padded or right-clipped) only.

gabime:
sounds good

eudoxos:
Current idea for the mini-syntax:
* `[nothing]`, `-`. `=` retain their function;
* `>` for clip-right (or pad-right, if shorter), `<` for clip-left (or pad-left, if shorter);
* `|` for ellipsize-middle (insert `…` or `_` – configurable globally?), `/` for ellipsize-right (like clip-right, only put ellipsis instead of the last char if longer), `\` for ellipsize-left (like clip-left, with ellipsis instead of the first char)

gabime:
To much syntax for such small feature :)
I suggest to keep it simple with a single flag that  truncates the right side if longer than the specified size.
 This would cover 99% of the needs.

gabime:
@eudoxos Implemeted in cee35f7d246f7afb5fa6cc9d1509bacbea9201e7.

Use the optional ```!``` flag. It would truncate the right side if needed (see also the [updated wiki](https://github.com/gabime/spdlog/wiki/3.-Custom-formatting#aligning)) :
```
set_pattern("[%8!] %v");
set_pattern("[%-8!] %v");
set_pattern("[%=8!] %v");
```

Feedback is welcome

FreeYourSoul:
This feature is awesome, just saying

gabime:
@FreeYourSoul Thanks!

amurzeau:
Hi,
Thanks for this feature too :) 

Truncating on the left would have been intersting when truncating function name, as when using C++, namespace and class names are less discriminating than the method name:

For example with `%30!!` with method `spdlog::details::scoped_padder::count_digits`:
- Truncating on the right (as done currently): `spdlog::details::scoped_padder`
- Truncating on the left: `s::scoped_padder::count_digits`

This is the same for filenames if we get absolute path, the beginning of the path is less likely interesting the the end.

[...]

PR #1314, opened 2019-11-12
Preliminary attempt at right_clip alignment

eudoxos:
This is a lame attempt at #1297, doing right-clip only. It does not fit the current architecture very well, and left-clipping would be impossible. It also relies on conservative resizing of `memory_buf_t` (as there is no other way to remove chars from the end of the buffer).

Perhaps a better approach would be to derive from (or extend) `aggregate_formatter` which would take the entire user-string and decide what to do: insert padding left/right, clip left/right, or clip left/middle/right with ellipsis marker.

This is from the example:
image

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -36,8 +36,9 @@ int main(int, char *[])
     spdlog::debug("This message should be displayed..");
 
     // Customize msg format for all loggers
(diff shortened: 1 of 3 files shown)


gabime:
Thanks.
I already started developing something similar few days ago.  Will commit soon.

eudoxos:
I am sure yours will be better, with the knowledge of internals you posess :) Cheers!

PR #1322, opened 2019-11-19
Allowing multiple registries (removing singleton from the registry)

caiazza:
As I mentioned in the other pull request now also the async factory is a template function in the registry.
The current versions of the factory function simply call into the factory function of the global instance, without any semantic change. Additionally a user is able to create a registry and create the factory functions of that registry to create new loggers in that specific registry.
You are still able to use all the global functionalities just as before but with the additional possibility to use the registry independently.

The one thing that can be changed, depending on your taste is the following:
In the current version you have 2 factory functions, create_async and create_async_nb which simply call a specialization of a templated struct where the template parameter which is specialized is just the value of an enum. You could achieve the same result passing the OverflowPolicy as a function argument to the static function of the async_factory_impl struct. This all seems a bit convoluted and I don't understand the ratio of it.

In my change, to keep it minimal, the registry as a create_async function that just mirrors the behaviour of that static function but rather than using a template argument to make any distinction it uses the OverflowPolicy as a simple function argument.
The alternative is to have 2 functions in the registry, called creat_async and create_async_nb which mirror the actual syntax of the 2 free functions you currently have.

What do you prefer?


gabime:
As I see it, you there a 2 distinct goals in this PR: 
1. Make the registry ctor/dtor public.
2. Be able to use the factories using given registry object instead of the global instance. 
For this I think a simple solution is just add overload to the factories to accept reference to registry:
```c++
struct synchronous_factory
{

    template<typename Sink, typename... SinkArgs>
    static std::shared_ptr<spdlog::logger> create(std::string logger_name, details::registry &registry, SinkArgs &&... args)
    {
        auto sink = std::make_shared<Sink>(std::forward<SinkArgs>(args)...);
        auto new_logger = std::make_shared<spdlog::logger>(std::move(logger_name), std::move(sink));
        registry.initialize_logger(new_logger);
        return new_logger;
    }

    // default registry
    template<typename Sink, typename... SinkArgs>
    static std::shared_ptr<spdlog::logger> create(std::string logger_name, SinkArgs &&... args)
    {
        return create<Sink>(std::move(logger_name), details::registry::instance(), std::forward<SinkArgs>(args)...);
    }
};
```

Same goes for the async factory:
```c++
template<async_overflow_policy OverflowPolicy = async_overflow_policy::block>
struct async_factory_impl
{
    template<typename Sink, typename... SinkArgs>
    static std::shared_ptr<async_logger> create(std::string logger_name, details::registry& registry_inst, SinkArgs &&... args)
    {
        // create global thread pool if not already exists..
        auto &mutex = registry_inst.tp_mutex();
        std::lock_guard<std::recursive_mutex> tp_lock(mutex);
        auto tp = registry_inst.get_tp();
        if (tp == nullptr)
        {
            tp = std::make_shared<details::thread_pool>(details::default_async_q_size, 1);
            registry_inst.set_tp(tp);
        }

        auto sink = std::make_shared<Sink>(std::forward<SinkArgs>(args)...);
        auto new_logger = std::make_shared<async_logger>(std::move(logger_name), std::move(sink), std::move(tp), OverflowPolicy);
        registry_inst.initialize_logger(new_logger);
        return new_logger;
    }

    template<typename Sink, typename... SinkArgs>
    static std::shared_ptr<async_logger> create(std::string logger_name, SinkArgs &&... args)
    {
        return async_factory_impl::create<Sink>(std::move(logger_name),  details::registry::instance() ,std::forward<SinkArgs>(args)...);
    }
};
```

In any case, There is no demand for such feature, so I will close this PR for now. If the need arise in the  future, I will add the needed changes.

caiazza:
Yes, you could make the factory like that as well, if you prefer. It doesn't change anything from the conceptual point of view.

Anyway I don't know what is your threshold for a demand, you can say I'm demanding it ;)
I'm actually already using it in that way in my project, to avoid global instances. I was hoping you could merge it to avoid maintaining a separate fork just for this trivial issue, but be as you wish.

PR #1370, opened 2019-12-23
Fix invalid meson option

segfault-magnet:
The option is: 'enable_tests_ho' as stated in meson_options.txt


gabime:
thanks @segfault-magnet

PR #1377, opened 2019-12-30
check thread-local storage support for clang

tangshi:
thread-local storage is not supported on iOS 8.0 and earlier

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -56,7 +56,7 @@
 
 // disable thread local on msvc 2013
 #ifndef SPDLOG_NO_TLS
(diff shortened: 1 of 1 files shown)


gabime:
Are you sure that  cxx_thread_local is always defined for all compiler versions if tls is enabled? this is risky

tangshi:
> Are you sure that cxx_thread_local is always defined for all compiler versions if tls is enabled? this is risky

No, its a clang specific macro, so I check it before used 
`defined(__has_feature) && !__has_feature(cxx_thread_local)`

gabime:
Thanks, but closing. The build fails, and spdlog doesn't  officially supports ios anyway.

PR #1394, opened 2020-01-19
add max files for rotating days

ruoshui1314:


ruoshui1314:
It does not try to check files that exists in disk when initialization.
The patch will try to search files when service is started or restarted and also export max_files attribute into interface.

include/spdlog/sinks/daily_file_sink.h
```diff
@@ -104,6 +103,28 @@ class daily_file_sink final : public base_sink<Mutex>
     }
 
 private:
+    void init_filenames_q()
+    {
+        using details::os::path_exists;
+
+        filenames_q_ = details::circular_q<filename_t>(static_cast<size_t>(max_files_));
+        std::vector<filename_t> filenames;
```
gabime: This vector can be avoided. Just pushback to the q  instead of accumulating in filenames.

include/spdlog/sinks/daily_file_sink.h
```diff
@@ -104,6 +103,28 @@ class daily_file_sink final : public base_sink<Mutex>
     }
 
 private:
+    void init_filenames_q()
```
gabime: please add underscore since it is private

include/spdlog/sinks/daily_file_sink.h
ruoshui1314: I think it will remove the latest log file when rotating if push back to the q directly.
I am confused whether It leads to an mistake that could not get the previews log files when service or process is restarted.
In this way it will never remove any previews files.
@gabime

gabime: right

PR #1407, opened 2020-01-30
CMakeLists.txt: put GNUInstallDirs after set project language

vitlav:
See also https://gitlab.kitware.com/cmake/cmake/merge_requests/4300

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,11 +6,15 @@ cmake_minimum_required(VERSION 3.2)
 #---------------------------------------------------------------------------------------
 # Start spdlog project
 #---------------------------------------------------------------------------------------
-include(GNUInstallDirs)
 include(cmake/utils.cmake)
 include(cmake/ide.cmake)
 
 spdlog_extract_version()
+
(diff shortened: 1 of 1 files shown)


gabime:
@vitlav What would be the effect on the installed spdlog location after this fix?

vitlav:
> @vitlav What would be the effect on the installed spdlog location after this fix?
.so library will placed to appropriate libdir (/usr/lib64 on biarch system, /usr/lib/x86_64-linux-gnu/ on deb based systems).
libdir in autotools depends on arch.

gabime:
what is biarch system?

vitlav:
> what is biarch system?
Well, a 64 bit Linux system with biarch support does not use /usr/lib for 64 bit libs.
Please check these links:
https://bugzilla.mozilla.org/show_bug.cgi?id=496708
https://lists.gnu.org/archive/html/coreutils/2011-10/msg00057.html

In any case, warning against incorrect using GNUInstallDirs will appear since next cmake major release (3.17).

gabime:
Thx. Merged

PR #1409, opened 2020-02-02
Make syslog_sink constructor compatible to the syslog_sink constructor before…

msc42:
… commit e68cf1c9ed9aa64d23488bfd970a850a32666733


gabime:
It was changed for a reason... You use the factory methods if you don't want to pass params

PR #1414, opened 2020-02-05
Make tests support empty SPDLOG_EOL

### LINKED ISSUE #1413: Changing tweakme.h breaks tests, should this be improved upon?
First, thanks for the awesome work! I like spdlog a lot.

For our needs, we do not need most of the formatting, so we disable a number of features before building spdlog. We enable in `tweakme.h` the following:
 * `SPDLOG_CLOCK_COARSE`
 * `SPDLOG_NO_THREAD_ID`
 * `SPDLOG_NO_NAME`
 * `SPDLOG_NO_ATOMIC_LEVELS`
 * `SPDLOG_EOL`

When I run the tests afterwards, I get:
```
test cases: 118 | 106 passed | 12 failed
assertions: 421 | 409 passed | 12 failed
The following tests FAILED:
 	  1 - spdlog-utests (Failed)
```
This is pretty impressive, and I guess it means that I can trust the resulting library. But for my good nights sleep I want to look into the tests and see exactly which ones failed and why. And for our CI it would be great if I could get a green light from the test, so a 100% success.

Now my question: If I can provide a merge request that adapts some of the tests to support also the defines from `tweakme.h`, would that be acceptable for merging? I guess I can not cover all possible cases, but hopefully a number of relevant ones. For example, supporting an empty `SPDLOG_EOL` would be a relatively straightforward case in `TEST_CASE("rotating_file_logger1", "[rotating_logger]]")` where the number of expected lines would be `1` instead of `10`:
```
...
    for (int i = 0; i < 10; ++i)
    {
        logger->info("Test message {}", i);
    }

    logger->flush();
    auto filename = basename;
    if (strlen(spdlog::details::os::default_eol)) == 0) {
        REQUIRE(count_lines(filename) == 1);
    } else {
        REQUIRE(count_lines(filename) == 10);
    }
}
```

emmenlau:
This PR is currently a demo only. It is created against tag v1.5.0. It should show-case how one could add support for empty `SPDLOG_EOL` in the tests. This seems a "relatively" common request amongst users so it may be nice if the tests can support it.

Closes #1413 .

### DIFF
--- a/tests/test_file_logging.cpp
+++ b/tests/test_file_logging.cpp
@@ -15,8 +15,10 @@ TEST_CASE("simple_file_logger", "[simple_logger]]")
     logger->info("Test message {}", 2);
 
     logger->flush();
-    REQUIRE(file_contents(filename) == std::string("Test message 1\nTest message 2\n"));
-    REQUIRE(count_lines(filename) == 2);
+    require_message_count(filename, 2);
+    using spdlog::details::os::default_eol;
+    REQUIRE(file_contents(filename) == fmt::format("Test message 1{}Test message 2{}",
+        default_eol, default_eol));
 }
 
 TEST_CASE("flush_on", "[flush_on]]")
@@ -34,8 +36,10 @@ TEST_CASE("flush_on", "[flush_on]]")
     logger->info("Test message {}", 1);
     logger->info("Test message {}", 2);
 
-    REQUIRE(file_contents(filename) == std::string("Should not be flushed\nTest message 1\nTest message 2\n"));
-    REQUIRE(count_lines(filename) == 3);
+    require_message_count(filename, 3);
+    using spdlog::details::os::default_eol;
+    REQUIRE(file_contents(filename) == fmt::format("Should not be flushed{}Test message 1{}Test message 2{}",
+        default_eol, default_eol, default_eol));
 }
 
 TEST_CASE("rotating_file_logger1", "[rotating_logger]]")
@@ -52,7 +56,7 @@ TEST_CASE("rotating_file_logger1", "[rotating_logger]]")
 
     logger->flush();
     auto filename = basename;
-    REQUIRE(count_lines(filename) == 10);
+    require_message_count(filename, 10);
 }
 
 TEST_CASE("rotating_file_logger2", "[rotating_logger]]")
@@ -81,7 +85,8 @@ TEST_CASE("rotating_file_logger2", "[rotating_logger]]")
 
     logger->flush();
     auto filename = basename;
-    REQUIRE(count_lines(filename) == 10);
+    require_message_count(filename, 10);
+
     for (int i = 0; i < 1000; i++)
     {
 
--- a/tests/test_async.cpp
+++ b/tests/test_async.cpp
@@ -169,9 +169,10 @@ TEST_CASE("to_file", "[async]")
         }
     }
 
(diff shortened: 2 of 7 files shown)


gabime:
I think it could be cleaner if a new “count_messages()” function was introduced with all the logic in it..

emmenlau:
> I think it could be cleaner if a new “count_messages()” function was introduced with all the logic in it..

Yes, you are right. I'll take a look at it. Do you have a preferred source file where this method would go, since its used in multiple tests?

gabime:
tests/utils.cpp is the most suitable for this

emmenlau:
Dear @gabime thanks for the constructive feedback. I've adapted the tests and indeed the code comes out much better. Do you think this is reasonable to add?

gabime:
sure, but they fail

emmenlau:
The travis build works except an installation problem with `g++7`. The Appveyor build is something else, that may be related with newlines on MSVC. That brings up a question. It fails in places where I changed
```
"Test message 2\n"
```
to
```
"Test message 2" + std::string(spdlog::details::os::default_eol)
```

Its actually interesting how this worked before, because AFAIK on MSVC the `SPDLOG_EOL` was defined as `\r\n`. So why was the test expecting only `\n`? Or am I on the wrong track?

gabime:
That'ss because ```file_contents(filename) ``` open the file in text mode so the \r\n gotconverted to \n

tests/test_file_logging.cpp
```diff
@@ -34,8 +34,8 @@ TEST_CASE("flush_on", "[flush_on]]")
     logger->info("Test message {}", 1);
     logger->info("Test message {}", 2);
 
-    REQUIRE(file_contents(filename) == std::string("Should not be flushed\nTest message 1\nTest message 2\n"));
-    REQUIRE(count_lines(filename) == 3);
+    require_message_count(filename, 3);
+    REQUIRE(file_contents(filename) == std::string("Should not be flushed") + spdlog::details::os::default_eol + std::string("Test message 1") + spdlog::details::os::default_eol + std::string("Test message 2") + spdlog::details::os::default_eol);
```
gabime: please use fmt::format for a cleaner version:

```c++
using spdlog::details::os::default_eol;
REQUIRE(file_contents(filename) == fmt::format("Should not be flushed{}Test message 1{}Test Message 2{}",
    default_eol, default_eol, default_eol);
```

emmenlau: I see that this is a cleaner notation. But the style I'm currently using is also used in several other places. I won't gt time to make this improvement, is it ok to leave this out?

emmenlau:
> That's because `file_contents(filename) ` open the file in text mode so the `\r\n` got converted to `\n`

How should I best solve this? Or can you help? I personally would change the behavior of `file_contents()` to open in binary mode, but will that change many other tests?

gabime:
Open in binary mode is better, I agree. I am not sure how it would affect the windows tests, if any. You can try and see

tests/test_file_logging.cpp
gabime: There is no rush, please take your time. I will wait with the merge.

emmenlau:
Seems I was lucky with this change, the tests now all work? Could you consider this for merging?

gabime:
Sure, but I would appreciate if you first added the changes I requested.

gabime:
Thanks @emmenlau . Merged.

Issue #989, opened 2019-02-06
Windows Event Log Support


sinainnet:
Hi,

I'm curious to know if there is any implementation of sink in SPDLOG based on Windows event log. I will be grateful if anyone let me know.

Thanks

howar6hill:
@gabime Is this feature still needed? If yes, I want to help on this.

gabime:
Its nice to have. Thanks !

TheQwertiest:
Shouldn't this be fixed by #1418 ?

gabime:
Right. Closing..

gabime:
Fixed in #1418

PR #1418, opened 2020-02-07
Add eventlog_sink for logging to Windows Event Log (local only)

ban-dana:
adressing issue https://github.com/gabime/spdlog/issues/989


include/spdlog/sinks/eventlog_sink.h
```diff
@@ -0,0 +1,100 @@
+// Copyright(c) 2015-present, Gabi Melman & spdlog contributors.
```
gabime: No need for this abstraction sink. We need just a single "win_eventlog_sink.h" and users just would use the "win_eventlog_sink" in it directly

include/spdlog/sinks/eventlog_sink.h
```diff
@@ -0,0 +1,100 @@
+// Copyright(c) 2015-present, Gabi Melman & spdlog contributors.
+// Distributed under the MIT License (http://opensource.org/licenses/MIT)
+
+// Requires the following registry entries to be present, with the following modifications:
```
gabime: I saw that the ```add_registry_info``` already creates such entry? why is it "Requires the following registry entries to be present" ?

include/spdlog/sinks/eventlog_sink.h
```diff
@@ -0,0 +1,100 @@
...
+ * Windows Event Log sink
+ */
+template<typename Mutex>
+class eventlog_sink : public base_sink<Mutex>
+    {
+    public:
+        explicit eventlog_sink(
+            std::string const& source,
+            std::string const& log = "Application",
+            std::string const& message_file_path = "%windir%\\System32\\mscoree.dll");
```
gabime: why this dll in particualr ?

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
+// Copyright(c) 2020 bandana2004@gmail.com
+// Distributed under the MIT License (http://opensource.org/licenses/MIT)
+
+namespace eventlog_sink_win32 {
```
gabime: This should be the only file/sink (no need for extra abstraction)

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    win32_error(std::string const& func_name, DWORD error = GetLastError())
+        : spdlog_ex(format(func_name, error), error)
+    {}
+};
+
+struct sid
+{
+    static SID* duplicate(SID* p_src_sid)
```
gabime: Please add some comments It is really hard to understand what is the purpose of those function

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+        if (!CopySid(sid_size, p_dest_sid, p_src_sid))
+        {
+            sid::free(&p_dest_sid);
+            SPDLOG_THROW(win32_error("CopySid"));
+        }
+
+        return p_dest_sid;
+    }
+
+    static void free(SID** ppsid)
```
gabime: better call it something else than "free" (free_sid)

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+            if (!HeapFree(GetProcessHeap(), 0, (LPVOID) *ppsid))
+            {
+                SPDLOG_THROW(win32_error("HeapFree"));
+            }
+        }
+
+        *ppsid = nullptr;
+    }
+
+    static SID* get_current_user_sid()
```
gabime: Whats the purpose ? Please add comments

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+
+        return sid::duplicate((SID *) ((TOKEN_USER*) buffer.data())->User.Sid);
+    }
+};
+
+struct eventlog
+{
+    static WORD get_event_type(details::log_msg const& msg)
+    {
+        WORD type = EVENTLOG_SUCCESS;
```
gabime: I think simple switch is much clearer here

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+                }
+            }
+        }
+        return type;
+    }
+
+    static WORD get_event_category(details::log_msg const& msg)
+    {
+        WORD category = 1;
+        if (msg.level >= level::debug)
```
gabime: think simple switch is much clearer here

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+        long stat = RegCreateKeyEx(HKEY_LOCAL_MACHINE, subkey.c_str(), 0, NULL,
+                                   REG_OPTION_NON_VOLATILE, KEY_SET_VALUE, NULL,
+                                   &hkey, &disposition);
+        if (stat == ERROR_SUCCESS)
+        {
+            if (disposition == REG_CREATED_NEW_KEY && !message_file_path.empty())
+            {
+                auto const expanded_message_file_path_length = ExpandEnvironmentStrings(message_file_path.c_str(), (LPSTR) &disposition, 0);
+                std::vector<char>expanded_message_file_path(expanded_message_file_path_length + 1);
+                ExpandEnvironmentStrings(message_file_path.c_str(), expanded_message_file_path.data(), expanded_message_file_path_length);
```
gabime: should be the return value here (can fail with rv 0)

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+            SPDLOG_THROW(win32_error("ReportEvent"));
+    }
+
+    void flush_() override {}
+
+public:
+    sink(
+        std::string const& source, 
+        std::string const& log, 
+        std::string const& message_file_path)
```
gabime: Is this "message_file_path" param really needed? Most users will just be happy with default dll

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+            ~process_token_t()
+            {
+                if (hasToken_ && !CloseHandle(hToken_))
+                    win32_error::report(win32_error::format("CloseHandle"));
+            }
+        } current_process_token(GetCurrentProcess()); // GetCurrentProcess returns pseudohandle, no leak here!
+
+        // Get the required size
+        DWORD tusize = 0;
+        GetTokenInformation(current_process_token.hToken_, TokenUser, NULL, 0, &tusize);
```
gabime: Should be checked (can fail with zero)

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+    {
+        if (ppsid && *ppsid)
+        {
+            if (!HeapFree(GetProcessHeap(), 0, (LPVOID) *ppsid))
+            {
+                SPDLOG_THROW(win32_error("HeapFree"));
+            }
+        }
+
+        *ppsid = nullptr;
```
gabime: will crash if ppsid is null

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    static SID* get_current_user_sid()
+    {
+        struct process_token_t
```
gabime: Please add comments about all those window functions - its very hard to understand the purpose

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+        DWORD tusize = 0;
+        GetTokenInformation(current_process_token.hToken_, TokenUser, NULL, 0, &tusize);
+        if (GetLastError() != ERROR_INSUFFICIENT_BUFFER)
+            SPDLOG_THROW(win32_error("GetTokenInformation"));
+
+        std::vector<unsigned char> buffer(tusize);
+        if (!GetTokenInformation(current_process_token.hToken_, TokenUser, (LPVOID) buffer.data(), tusize, &tusize))
+            SPDLOG_THROW(win32_error("GetTokenInformation"));
+
+        return sid::duplicate((SID *) ((TOKEN_USER*) buffer.data())->User.Sid);
```
gabime: Do we really need all those complicated sid functions? The "ReportEventA" docs state it can be NULL.

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+    HANDLE hEventLog_;
+    SID* current_user_sid_;
+    std::string source_;
+    std::string log_;
+
+    void add_registry_info(std::string const& message_file_path)
+    {
+        std::string subkey("SYSTEM\\CurrentControlSet\\Services\\EventLog\\");
+        subkey += log_ + "\\" + source_;
+
```
gabime: Can be replaced with the cleaner
```c++
std::string subkey = fmt::format("SYSTEM\\CurrentControlSet\\Services\\EventLog\\{}\\{}", log_, source_);
```

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+                                   &hkey, &disposition);
+        if (stat == ERROR_SUCCESS)
+        {
+            if (disposition == REG_CREATED_NEW_KEY && !message_file_path.empty())
+            {
+                auto const expanded_message_file_path_length = ExpandEnvironmentStrings(message_file_path.c_str(), (LPSTR) &disposition, 0);
+                std::vector<char>expanded_message_file_path(expanded_message_file_path_length + 1);
+                ExpandEnvironmentStrings(message_file_path.c_str(), expanded_message_file_path.data(), expanded_message_file_path_length);
+
+                RegSetValueEx(hkey, "EventMessageFile", 0, REG_SZ, (LPBYTE) expanded_message_file_path.data(), expanded_message_file_path_length);
```
gabime: Should be checked for success (ERROR_SUCCESS)

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+                DWORD typesSupported = 7;
+                RegSetValueEx(hkey, "TypesSupported", 0, REG_DWORD, (LPBYTE) &typesSupported, sizeof(DWORD));
```
gabime: Should be checked for success (ERROR_SUCCESS)

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+        DWORD disposition {};
+        long stat = RegCreateKeyEx(HKEY_LOCAL_MACHINE, subkey.c_str(), 0, NULL,
+                                   REG_OPTION_NON_VOLATILE, KEY_SET_VALUE, NULL,
+                                   &hkey, &disposition);
+        if (stat == ERROR_SUCCESS)
+        {
+            if (disposition == REG_CREATED_NEW_KEY && !message_file_path.empty())
+            {
+                auto const expanded_message_file_path_length = ExpandEnvironmentStrings(message_file_path.c_str(), (LPSTR) &disposition, 0);
+                std::vector<char>expanded_message_file_path(expanded_message_file_path_length + 1);
```
gabime: I get warning here: Severity	Code	Description	Project	File	Line
Warning	C26451	Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow

include/spdlog/sinks/eventlog_sink_win32.h
ban-dana: Right, but then the user information won't be available in the event viewer, which may sometimes be useful as the event log is a global facility of multi-user environment.

include/spdlog/sinks/eventlog_sink.h
ban-dana: Registering a source is something that requires elevation (aka "running as administrator") so it's usually done during the setup. However I wanted to give the user an option to do it in code.

include/spdlog/sinks/eventlog_sink.h
ban-dana: It's a part of Microsoft .NET Framework that is installed on pretty much every Windows system out there, and this dll happens to contain the needed resource. I don't think this is going to change any time soon, but since it is still an undocumented hack, I also give a way to do that with a custom built dll if so desired or if Microsoft changes their old dll or whatever.

include/spdlog/sinks/win_eventlog_sink.h
```diff
@@ -0,0 +1,357 @@
...
+namespace sinks {
+
+namespace win_eventlog {
+
+namespace internal
+{
+
+struct utils
+{
+    /** Reports a message to stderr */
```
gabime: not needed. just throw spdlog_ex

include/spdlog/sinks/win_eventlog_sink.h
```diff
@@ -0,0 +1,357 @@
...
+    static void report(char const* message) SPDLOG_NOEXCEPT
+    {
+        fprintf(stderr, "%s", message);
+        fflush(stderr);
+    }
+
+    /** Reports a message to stderr */
+    static void report(std::string const& message) SPDLOG_NOEXCEPT
```
gabime: not needed. just throw spdlog_ex

include/spdlog/sinks/win_eventlog_sink.h
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /** Runs a func and reports an exception to stderr if any */
+    static void report_exceptions(std::function<void()> func)
+    {
+        report_exceptions<int>([&func] () { func(); return 0; }, 0);
+    }
+};
+
+/** Windows error */
```
gabime: not needed. tooo complicated. just throw spdlog_ex with general description.

include/spdlog/sinks/win_eventlog_sink.h
```diff
@@ -0,0 +1,357 @@
...
+                if (stat != ERROR_SUCCESS)
+                    SPDLOG_THROW(win32_error("RegSetValueEx", stat));
+            }
+        }
+        else
+        {
+            SPDLOG_THROW(win32_error("RegCreateKeyEx", stat));
+        }
+
+        RegCloseKey(hkey);
```
gabime: need to closed even if exception happened above

include/spdlog/sinks/eventlog_sink_win32.h
```diff
@@ -0,0 +1,246 @@
...
+    {
+        std::string subkey("SYSTEM\\CurrentControlSet\\Services\\EventLog\\");
+        subkey += log_ + "\\" + source_;
+
+        ::HKEY hkey {};
+        DWORD disposition {};
+        long stat = RegCreateKeyEx(HKEY_LOCAL_MACHINE, subkey.c_str(), 0, NULL,
+                                   REG_OPTION_NON_VOLATILE, KEY_SET_VALUE, NULL,
+                                   &hkey, &disposition);
+        if (stat == ERROR_SUCCESS)
```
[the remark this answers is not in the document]
ban-dana: Correct, on modern Windows non-admin users are not allowed to modify HKEY_LOCAL_MACHINE by default.

include/spdlog/sinks/win_eventlog_sink.h
ban-dana: This is used in two cases: one is when getting the SID for the current user fails for some reason, it still allows moving forward without the user id which is better than nothing.
Another thing is when something bad happens in the destructor, "just throw spdlog_ex" may end up with an exception during stack unwind and crash. So all the destructors just report any exceptions to stderr.

gabime: suddenly printing to stderr might not be desired by users and even lead to bad performance. If the code need to move forward(like in the dtor), than just let it move forward, if not just throw..

include/spdlog/sinks/eventlog_sink.h
gabime: Thus approach still seems bit confusing to me. If we expect the user to set those reg entries beforehand using regedit, then there is no real benefit for all this registry code, and it can be simplified dramatically. 
I suggest removing the registry code  it completely and let the user take care of it as written in your top comment (use regedit).

gabime:
@ban-dana Thanks. Merged.
I also added optimization to avoid generating ```std::string``` on each call [here](https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/sinks/win_eventlog_sink.h#L225-L226).

Issue #1419, opened 2020-02-07
1.5.0: errors while running the tests using cmake


mr-c:
This happened while testing a new Debian package for spdlog

```
cd obj-x86_64-linux-gnu && cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DCMAKE_INSTALL_RUNSTATEDIR=/run "-GUnix Makefiles" -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DSPDLOG_BUILD_EXAMPLE=ON -DSPDLOG_BUILD_EXAMPLE_HO=ON -DSPDLOG_BUILD_BENCH=OFF -DSPDLOG_BUILD_TESTS=ON -DSPDLOG_BUILD_TESTS_HO=ON -DSPDLOG_BUILD_SHARED=ON -DSPDLOG_FMT_EXTERNAL=ON ..

# ...

	cd obj-x86_64-linux-gnu && make -j8 test ARGS\+=-j8
make[1]: Entering directory '/build/spdlog-1.5.0/obj-x86_64-linux-gnu'
Running tests...
/usr/bin/ctest --force-new-ctest-process -j8
Test project /build/spdlog-1.5.0/obj-x86_64-linux-gnu
    Start 1: spdlog-utests
    Start 2: spdlog-utests-ho
1/2 Test #1: spdlog-utests ....................***Failed    3.46 sec

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
spdlog-utests is a Catch v2.8.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
file_helper_size
-------------------------------------------------------------------------------
/build/spdlog-1.5.0/tests/test_file_helper.cpp:28
...............................................................................

/build/spdlog-1.5.0/tests/test_file_helper.cpp:36: FAILED:
  REQUIRE( static_cast<size_t>(helper.size()) == expected_size )
with expansion:
[...]

PR #1421, opened 2020-02-07
Don't run spdlog-utests and spdlog-utests-ho in parallel

tt4g:
Ref #1419 

Change ctest not to execute `spdlog-utests` and `spdlog-utests-ho` in parallel.


tt4g:
Since `meason.build` does not seem to be running in parallel, `ctest` also suppressed parallel execution.

https://github.com/gabime/spdlog/blob/5378f352395c0a03bb5be1c1411523475d95b145/tests/meson.build#L42

https://github.com/gabime/spdlog/blob/5378f352395c0a03bb5be1c1411523475d95b145/tests/meson.build#L47

gabime:
Thanks @tt4g , Merged.

PR #1423, opened 2020-02-07
Properly handling SPDLOG_PREVENT_CHILD_FD

dominicpoeschko:
Using the SPDLOG_PREVENT_CHILD_FD option there where still a race when
a other thread was using fork and exec in between the call to fopen and fcntl.

Using open and O_CLOEXEC when possible prevents this race.

I have no idea if this problem  exists on Windows.


dominicpoeschko:
O_CLOEXEC was not defined in earlier posix versions

gabime:
well in this case it is better to get a compiler error instead of defaulting to unexpected behavior

PR #1424, opened 2020-02-08
adding additional log overload to prevent unnecessary fmt::format

dominicpoeschko:
calling log with a string_view as msg called
``` c++
 template<typename... Args>
    void log(source_loc loc, level::level_enum lvl, string_view_t fmt, const Args &... args)
```

instead of
``` c++
template<class T, typename std::enable_if<std::is_convertible<const T &, spdlog::string_view_t>::value, T>::type * = nullptr>
    void log(source_loc loc, level::level_enum lvl, const T &msg)
```

which lead to an unnecessary call to fmt::format


gabime:
Merged. Thanks.

PR #1427, opened 2020-02-10
Handling SPDLOG_PREVENT_CHILD_FD in tcp_sink

dominicpoeschko:
Adding SOCK_CLOEXEC to socket

Fixing bug in sink_it_ (bytes_sent not added to buffer)

If requested I would also add some code to better handle the connect.
When behind a proxy and the tcp packed get swallowed the default connect timeout in
most Implementations is around 2 minutes.
I would add some code to reduce this timeout, maybe also as parameter to constructor.

### DIFF
--- a/include/spdlog/sinks/tcp_sink.h
+++ b/include/spdlog/sinks/tcp_sink.h
@@ -47,7 +47,7 @@ protected:
         size_t bytes_sent = 0;
         while (bytes_sent < formatted.size())
         {
-            auto write_result = ::write(sock_, formatted.data(), formatted.size() - bytes_sent);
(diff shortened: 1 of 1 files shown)


gabime:
> I would add some code to reduce this timeout, maybe also as parameter to constructor.

Sounds good. Thanks.

dominicpoeschko:
Another useful feature would be some tag to prepend the message with.
This could be useful when one Server is handling multiple clients so he knows which "application"
send the message

gabime:
Once could set for log pattern to achieve this. No need to do this in the sink itself..

PR #1428, opened 2020-02-10
Add timeout to tcp_sink connect

dominicpoeschko:
somewhat more code then I thought.
The code comes form my personal codebase and is tested since some time,
that is the reason for all the edge cases handled.

### DIFF
--- a/include/spdlog/sinks/tcp_sink.h
+++ b/include/spdlog/sinks/tcp_sink.h
@@ -3,20 +3,20 @@
 
 #pragma once
 
-#include <arpa/inet.h>
-#include <fcntl.h>
-#include <memory>
-#include <mutex>
-#include <netdb.h>
-#include <poll.h>
 #include <spdlog/common.h>
 #include <spdlog/details/null_mutex.h>
 #include <spdlog/sinks/base_sink.h>
-#include <string>
+
+#include <arpa/inet.h>
+#include <fcntl.h>
+#include <netdb.h>
+#include <poll.h>
 #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


dominicpoeschko:
sorry for my inability to use your tools...

gabime:
Thanks. Thats a lot of code indeed. Seems battle proven :).
There are many ways that this sink can delay the program, and even crash it (sigpipe if the connection breaks) , so if a user is sensitive about delays he should use this sink under spdlog::async_logger anyway.

I suggest we leave this code as reference and  not merge and decide in the future if it is really needed and worth the extra complexity.

btw, there are some features I started work on: reconnect on error, lazy connect(not to connect on ctor), option to ignore sigpipe, etc.

dominicpoeschko:
reconnect on error was a feature I would have suggested next :)

for sure merge the code whenever you need it :)

When needed I can contribute some features IPV6 support, UDP, SSL.

gabime:
UDP would be definitely needed, and probably should be a new dedicated simple sink.
SSL is interesting but probably out of the scope of spdlog. One can use ssl tunnel if needed.

dominicpoeschko:
when not merging you should fix the bug hiding here
```c++
            rv = ::connect(socket_rv, rp->ai_addr, rp->ai_addrlen);
            if (rv == 0)
            {
                break;
            }
            else
            {
                socket_rv = -1;
                last_errno = errno;
                ::close(socket_rv);
            }
```

setting socket_rv to -1 after close not befor

gabime:
sure. good catch

PR #1430, opened 2020-02-12
Add critical section to filename function of  each file sink

tt4g:
`filename()` member function of file sink has no critical section but `spdlog::details::file_helper::filename_` member variable may be changed by calling `sink_it_()`.

This causes a multi-threaded data race.
Added a critical section to avoid data race.

No critical section has been added to  `spdlog::sinks::basic_file_sink::filename()` because `spdlog::sinks::basic_file_sink` doesn't  rotate, the file name doesn't change.

### DIFF
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -70,8 +70,9 @@ public:
         }
     }
 
-    const filename_t &filename() const
+    const filename_t &filename()
     {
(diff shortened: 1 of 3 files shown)


gabime:
Thanks @tt4g  Merged. 
Looks like the return by reference is bad here as well..Need to return by value

tt4g:
> Need to return by value

Oops! I missed that.

PR #1434, opened 2020-02-12
Add a script to test the build performance

ClausKlein:
The build_performance_measure.py may be used to compare
the build time needed with cmake od meson based builds.

To reduce the build time, ccache is used.
Note: The tests needs to run at least 2 times to fill the ccache!

Prevent cmakelint warnings too.
Enable all tests and examples to have as much as possible build targets.


gabime:
Thanks for the pr. However I don't think it should in the scope of this repo.

Issue #1437, opened 2020-02-14
Prevent clang-tidy warnings


ClausKlein:
I use clang-tidy to check the example code:

```shell
bash-5.0$ ninja -vn check
[1/1] cd /Users/clausklein/Workspace/cpp/spdlog/example && /Users/clausklein/scripts/run-clang-tidy.py -p /Users/clausklein/Workspace/cpp/spdlog/.build-spdlog-Debug example
bash-5.0$ 
```
[run-clang-tidy-example.log](https://github.com/gabime/spdlog/files/4203345/run-clang-tidy-example.log)

ClausKlein:
Some may be prevented with this simple patch:

[mypatch.txt](https://github.com/gabime/spdlog/files/4203354/mypatch.txt)

ClausKlein:
P.S. you should fix your default config file and move it to top level source tree root dir

```shell
bash-5.0$ pwd
/Users/clausklein/Workspace/cpp/spdlog
bash-5.0$ cat .clang-tidy 
---
Checks: "-*,\
cppcoreguidelines-*,\
performance-*,\
-performance-unnecessary-value-param,\
modernize-*,\
-modernize-use-trailing-return-type,\
google-*,\
-google-runtime-references,\
misc-*,\
-misc-non-private-member-variables-in-classes,\
cert-*,\
readability-*,\
clang-analyzer-*\
"

WarningsAsErrors: ''
HeaderFilterRegex: 'include/spdlog/[^f].*'
AnalyzeTemporaryDtors: false
FormatStyle:     none

CheckOptions:
  - key:             google-readability-braces-around-statements.ShortStatementLines
    value:           '1'
  - key:             google-readability-function-size.StatementThreshold
    value:           '800'
  - key:             google-readability-namespace-comments.ShortNamespaceLines
    value:           '10'
  - key:             google-readability-namespace-comments.SpacesBeforeComments
    value:           '2'
  - key:             modernize-loop-convert.MaxCopySize
    value:           '16'
  - key:             modernize-loop-convert.MinConfidence
    value:           reasonable
  - key:             modernize-loop-convert.NamingStyle
    value:           CamelCase
  - key:             modernize-pass-by-value.IncludeStyle
    value:           llvm
  - key:             modernize-replace-auto-ptr.IncludeStyle
    value:           llvm
  - key:             modernize-use-nullptr.NullMacros
    value:           'NULL'

bash-5.0$ 
```

PR #1438, opened 2020-02-14
Feature/spdlog fmt external

ClausKlein:
this may be used to see  all my changes to improve code quality (too of the cmake based build system)

this fix too https://github.com/gabime/spdlog/issues/1437


gabime:
Not every tidy warning must be fixed. Using ‘magic_24’ instead of just 24 is funny.
Sometimes numbers are ok :)

PR #1440, opened 2020-02-15
Feature/add clang tidy target

ClausKlein:
Fix clang-tidy config file
Now you are able to check the rules configured within this file

Note: better to install and use run-clang-tidy.py  from clang git repo.
This script run in parallel clang-tidy processes as ninja do.
Please read this why it is better to use:
https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py
see too
https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/
https://clang.llvm.org/extra/clang-tidy/index.html


cmake/clang-tidy.cmake
```diff
@@ -0,0 +1,36 @@
+option(ENABLE_CLANG_TIDY "Add run-clang-tidy automatically to builds" OFF)
```
gabime: Please rename the option to SPDLOG_ENABLE_TIDY

cmake/clang-tidy.cmake
```diff
@@ -0,0 +1,36 @@
...
+find_program(CLANG_TIDY_EXE
```
gabime: this blck untill the end should be executed only if tidy option in ON

cmake/clang-tidy.cmake
```diff
@@ -0,0 +1,36 @@
...
+        NAMES run-clang-tidy.py run-clang-tidy run-clang-tidy-7
+        HINTS ${CMAKE_CURRENT_SOURCE_DIR}/scripts
+        DOC "Path to clang-tidy executable")
+if(CLANG_TIDY_EXE)
```
gabime: Probably should be renamed to TIDY_RUNNER instead OF CLANG_TIDY_EXE

cmake/clang-tidy.cmake
```diff
@@ -0,0 +1,36 @@
...
+    message(STATUS "clang-tidy found: ${CLANG_TIDY_EXE}")
+    set(DCMAKE_EXPORT_COMPILE_COMMANDS ON)
+    set(CLANG_TIDY_CMD ${CLANG_TIDY_EXE})
+    message(STATUS "cmake source dir: ${CMAKE_CURRENT_SOURCE_DIR}")
+
+    if(ENABLE_CLANG_TIDY)
+        # NOTE: the project config file .clang-tidy is not found if the
```
gabime: what do you mean?

cmake/clang-tidy.cmake
ClausKlein: OK

gabime:
Whats the benefit of using run-clang-tidy.py instead of the current script that is used in spdlog?

cmake/clang-tidy.cmake
ClausKlein: cmake may create a pre/ or post build target with clang-tidy.
this is anabled if  CMAKE_LANG_CLANG_TIDY is set. see https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_CLANG_TIDY.html

Note: the build will be than very slow! So it is disabled default

cmake/clang-tidy.cmake
ClausKlein: OK

cmake/clang-tidy.cmake
ClausKlein: wait a moment ...

There are tree way to work:

as pre compile step:
```shell
[1/26] /opt/local/bin/cmake -E __run_co_compile --launcher=ccache --tidy=/opt/local/bin/clang-tidy --source=/Users/clausklein/Workspace/cpp/spdlog/src/spdlog.cpp -- /opt/local/bin/g++  -DSPDLOG_COMPILED_LIB -I/Users/clausklein/Workspace/cpp/spdlog/include -g   -Wall -Wextra -Wconversion -pedantic -Wfatal-errors -std=c++11 -MD -MT CMakeFiles/spdlog.dir/src/spdlog.cpp.o -MF CMakeFiles/spdlog.dir/src/spdlog.cpp.o.d -o CMakeFiles/spdlog.dir/src/spdlog.cpp.o -c /Users/clausklein/Workspace/cpp/spdlog/src/spdlog.cpp
[2/26] /opt/local/bin/cmake -E __run_co_compile --launcher=ccache --tidy=/opt/local/bin/clang-tidy --source=/Users/clausklein/Workspace/cpp/spdlog/src/stdout_sinks.cpp -- /opt/local/bin/g++  -DSPDLOG_COMPILED_LIB -I/Users/clausklein/Workspace/cpp/spdlog/include -g   -Wall -Wextra -Wconversion -pedantic -Wfatal-errors -std=c++11 -MD -MT CMakeFiles/spdlog.dir/src/stdout_sinks.cpp.o -MF CMakeFiles/spdlog.dir/src/stdout_sinks.cpp.o.d -o CMakeFiles/spdlog.dir/src/stdout_sinks.cpp.o -c /Users/clausklein/Workspace/cpp/spdlog/src/stdout_sinks.cpp
```
as target check:

```shell
cmake --build ../.build-spdlog-Debug --target check | tee run-clang-tidy.log 2>&1
```

and on console with run-clang-tidy.py which needs the compile_commands.json too
```shell
run-clang-tidy.py example/*.cpp
```

cmake/clang-tidy.cmake
ClausKlein:
```cmake
option(SPDLOG_ENABLE_TIDY "Add clang-tidy automatically as prebuild step" OFF)

find_program(CLANG_TIDY_EXE
        NAMES clang-tidy clang-tidy-7
        DOC "Path to clang-tidy executable")
find_program(CLANG_TIDY_RUNNER
        NAMES run-clang-tidy run-clang-tidy-7 run-clang-tidy.py
        HINTS ${CMAKE_CURRENT_SOURCE_DIR}/scripts)
if(CLANG_TIDY_EXE)
    message(STATUS "clang-tidy found: ${CLANG_TIDY_EXE}")
    set(DCMAKE_EXPORT_COMPILE_COMMANDS ON)
    set(CLANG_TIDY_CMD ${CLANG_TIDY_EXE})
    message(STATUS "cmake source dir: ${CMAKE_CURRENT_SOURCE_DIR}")

    if(SPDLOG_ENABLE_TIDY)
        # NOTE: the project config file .clang-tidy is not found if the
        # binary tree is not part of the source tree! CK
        set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY_CMD} CACHE STRING "" FORCE)
    else()
        set(CMAKE_CXX_CLANG_TIDY "" CACHE STRING "" FORCE) # delete it
    endif()

    if(CLANG_TIDY_RUNNER)
        if(NOT TARGET check)
            add_custom_target(check)
            message(STATUS "check target added")
            set_target_properties(check PROPERTIES EXCLUDE_FROM_ALL TRUE)
        endif()

        add_custom_command(TARGET check PRE_BUILD
            # -p BUILD_PATH Path used to read a compile command database (compile_commands.json).
            # NOTE: we use defaults checks from .clang-tidy and examples only!
            COMMAND ${CLANG_TIDY_RUNNER} -p ${CMAKE_CURRENT_BINARY_DIR} ### ${CMAKE_CURRENT_SOURCE_DIR}/example
            WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"   # location of compile_commands.json
            COMMENT "Running check on targets at ${CMAKE_CURRENT_SOURCE_DIR} ..."
            VERBATIM
        )
    endif()
else()
    message(AUTHOR_WARNING "clang-tidy not found!")
endif()
```

ClausKlein:
you may run in on CI server, in IDE, in editor, ...

ClausKlein:
It may be forgotten to used and tests only one file: example/example.cpp

The check target can always check all files of the project on CI / travis and from IDE’s

„Es ist nicht genug, zu wissen, man muß auch anwenden; 
     es ist nicht genug, zu wollen, man muß auch tun.“ 
Johann Wolfgang von Goethe <http://www.zitate-online.de/autor/goethe-johann-wolfgang-von/> (Werk: Wilhelm Meisters Wanderjahre)

gabime:
I meanthe .clang-tidy file itself, instead of the default ome

ClausKlein:
If I understand you right, you don't like the file .clang-tidy file itself?

This is not an options.

```shell
clang-tidy --help
...
clang-tidy options:

  --checks=<string>              - 
                                   Comma-separated list of globs with optional '-'
                                   prefix. Globs are processed in order of
                                   appearance in the list. Globs without '-'
                                   prefix add checks with matching names to the
                                   set, globs with the '-' prefix remove checks
                                   with matching names from the set of enabled
                                   checks. This option's value is appended to the
                                   value of the 'Checks' option in .clang-tidy
                                   file, if any.
  --config=<string>              - 
                                   Specifies a configuration in YAML/JSON format:
                                     -config="{Checks: '*',
                                               CheckOptions: [{key: x,
                                                               value: y}]}"
                                   When the value is empty, clang-tidy will
                                   attempt to find a file named .clang-tidy for
                                   each source file in its parent directories.
```

PR #1441, opened 2020-02-15
Prepare a test to find an installed libspdlog

ClausKlein:
The example should be able to build as a standalone project too if
external libfmt was used.

Require the minimum fmt version with find_package().

Require the minimum spdlog version with find_package() at
standalone example.


gabime:
I don't want to require 6.1.2. It can be used with older versions too

ClausKlein:
No minimum Version?

gabime:
Not sure what it is

ClausKlein:
Build spdlod with  extern fmt v5.0.0?

PR #1448, opened 2020-02-19
Add option for dll-local registry on Linux

Schneegans:
Currently, when using spdlog in shared libraries, it behaves quite differently between Windows and Linux. On Linux, the `static registry s_instance` is visible to other shared libraries and therefore shared across them. Hence we have **only one global** default logger. On Windows however, there is a `static registry s_instance`  for each shared object, hence we have **multiple default loggers**, one per shared object.

In our use case, we actually like the Windows-behaviour of having one default logger per shared object. This allows us to print log messages from multiple plug-ins of our application and the log messages will be automatically prepended with the plugin's name.

This PR adds an option to hide the `registry` class from shared objects on Linux and thus making spdlog behave more similar on Linux and Windows.

You can consider this PR as a suggestion - you have a much better understanding of the code - I am sure that this can be implemented much cleaner (are there other statics which behave similar?). And maybe there are some side-effects of this approach I am not aware of.

For us however, this is a very valuable tweak and others may find it useful as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ endif ()
 # build shared option
 if(NOT WIN32)
     option(SPDLOG_BUILD_SHARED "Build shared library" OFF)
+    option(SPDLOG_DLL_LOCAL_REGISTRY "If enabled, the logger registry will not be shared between shared objects" OFF)
 endif()
 
 # example options
@@ -120,6 +121,10 @@ if (SPDLOG_BUILD_SHARED)
     add_library(spdlog SHARED ${SPDLOG_SRCS} ${SPDLOG_ALL_HEADERS})
 else()
     add_library(spdlog STATIC ${SPDLOG_SRCS} ${SPDLOG_ALL_HEADERS})
+endif()
+
+if (SPDLOG_DLL_LOCAL_REGISTRY)
+    target_compile_definitions(spdlog PUBLIC SPDLOG_DLL_LOCAL_REGISTRY)
 endif()
 
 add_library(spdlog::spdlog ALIAS spdlog)
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -54,6 +54,12 @@
 #define SPDLOG_DEPRECATED
 #endif
 
+#if defined(SPDLOG_DLL_LOCAL_REGISTRY) && (defined(__GNUC__) || defined(__clang__))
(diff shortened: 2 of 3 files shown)


gabime:
> This allows us to print log messages from multiple plug-ins of our application and the log messages will be automatically prepended with the plugin's name.

How do you achieve this?

Schneegans:
The setup is like this: Our application and all of it's plug-ins link against a certain core library which contains a function like this:

```cpp
std::shared_ptr<spdlog::logger> createLogger(std::string const& name) {
  static auto sink = std::make_shared<spdlog::sinks::basic_file_sink_mt>("app.log");
  auto logger = std::make_shared<spdlog::logger>(name, sink);
  logger->set_pattern("%^[%L][%n]%$%v");
  return logger;
}
```

All plug-ins and other libraries of the application then call at some point
```cpp
spdlog::set_default_logger(core::createLogger("awesome-plugin"));
```

Then, when we call
```cpp
spdlog::info("Loading Done.");
```

in our plug-in, `[I][awesome-plugin] Loading Done.` will be printed to the log. Basically all libraries and all plug-ins have their own default logger, all of which use the same sink. I think this is currently impossible on Linux without the proposed `__attribute__((visibility("hidden")))`.

gabime:
Do you have multiple sink objects writing to same log file? seems problematic. better pass a single sink instance around..

Schneegans:
No it's just one. Note the `static` in the code above.

gabime:
Right. I missed the static.. 

Anyway, I am not sure if this need to be merged.. It solves a little a bit too niche problem.. 
I think in your case it would be better not to use the registry at all to avoid confusions..

kanli201904:
hello,  if we can' t use the theadpool in the shared library XXX.so,  only use threadpool in the application process (which call the XXX.so) ,can that solve the problem ?
 eg: in the shared library 
{
auto async_file = spdlog::rotating_logger_mt(logID, logFullname, 1024 * 1024 * 1024, fileNum);
		async_file->set_level(m_Loglevel);
		async_file->flush_on(spdlog::level::info);
		return async_file;
}
in the application :
{...
std::shared_ptr<spdlog::logger> logger = 
			std::make_shared<spdlog::async_logger>(logID, m_Sinks.begin(), m_Sinks.end(),
									spdlog::thread_pool(), spdlog::async_overflow_policy::block);
....
}

PR #1449, opened 2020-02-19
Add preserve_parent_name in async/logger::clone method

szobov:
In order to support logical hierarchy of loggers.

Imagine you have something like this architecture:
````
Application / Controller / First
                         / Second
````
And you want to have the log messages and understand from witch part it come.
After changes you can do:

```
first_controller_logger = controller_logger.clone("first", true);
```
and you will get in log output something similar to:

```
[2014-10-31 23:46:59.678] [controller.first] [info] Some message
```

What is quite convenient.

Actually now you can do it by wrapping clone method, but with my changes it
won't require any user's code.

Thanks.

### DIFF
--- a/include/spdlog/async_logger-inl.h
+++ b/include/spdlog/async_logger-inl.h
@@ -84,9 +84,14 @@ SPDLOG_INLINE void spdlog::async_logger::backend_flush_()
     }
 }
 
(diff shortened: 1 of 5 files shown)


gabime:
Thanks. spdlog doesn't have the notion of logger hierarchy, and as you noted this can achieved easily by passing the full name to clone.

PR #1459, opened 2020-03-02
replace bool with atomic<bool>

GuyKroizman:
fix usage of bool without sync between threads.
https://stackoverflow.com/a/9201088/193016


gabime:
All usages of to this bool is protected by mutex already...

GuyKroizman:
Right. Sorry for wasting you time.

PR #1472, opened 2020-03-15
Enhance to_hex like the hexdump command

ngugcx:
Support:
1. Show ASCII.
`spdlog::debug("{:a}", spdlog::to_hex(data.begin(), data.end()));`
image
2. Split output with different size per line.
`spdlog::debug("{:a}", spdlog::to_hex(data.begin(), data.end(), 16));`
image
`spdlog::debug("{:a}", spdlog::to_hex(data.begin(), data.end(), 8));`
image

### DIFF
--- a/include/spdlog/fmt/bin_to_hex.h
+++ b/include/spdlog/fmt/bin_to_hex.h
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include <cctype>
+
 //
 // Support for logging binary data as hex
 // format flags:
@@ -12,6 +14,7 @@
 // {:s} - don't separate each byte with space.
 // {:p} - don't print the position on each line start.
 // {:n} - don't split the output to lines.
+// {:a} - show ASCII if :n is not set 
 
 //
 // Examples:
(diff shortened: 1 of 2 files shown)


ngugcx:
The build failed with apt-get errors. Is there anything wrong with the build system? @gabime .
```
$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install clang-7
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 clang-7 : Depends: libstdc++-4.9-dev but it is not installable
           Depends: libgcc-4.9-dev but it is not installable
           Depends: libobjc-4.9-dev but it is not installable
E: Unable to correct problems, you have held broken packages.
apt-get.diagnostics
apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install clang-7" failed and exited with 100 during .
```

gabime:
Thanks. Cleaned a little and merged.

Very cool !

ngugcx:
Thanks!
I also used the delimiter variable at the beginning, and then I found that the delimiter may be other characters, but the padding should be blank.

PR #1475, opened 2020-03-18
Fix Windows setenv check

Crunkle:
Use the generic Windows macro here to fix compilation on non-MSVC toolchains. The `setenv` is not guaranteed available under MinGW and should also use `putenv`.


gabime:
Thanks. Merged.

PR #1478, opened 2020-03-18
Adding CIFuzz

Leo-Neat:
OSS-Fuzz was wondering if spdlog would use CIFuzz, a new feature being rolled out in OSS-Fuzz. It allows for pull requests to be fuzzed in CI using fuzz targets enabled in OSS-Fuzz. Documentation for this feature can be found at https://google.github.io/oss-fuzz/getting-started/continuous-integration/ Let us know if you have any questions!


gabime:
Thanks. Might do it in the future when I have more time

PR #1495, opened 2020-03-27
add /WX mscv compiler option for only mscv compiler

trondhe:
clang on windows support both gcc and mscv style options. Clang.exe on windows
defaults to gcc style, which will result in /WX unknown compiler command.
This will set /WX if and only if the compiler is MSVC and greater than version 1900.

Could perhaps be done in a prettier way, but it solves the issue nicely.

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -28,9 +28,9 @@ endfunction()
 
 # Turn on warnings on the given target
 function(spdlog_enable_warnings target_name)
-	set(MSVC_OPTIONS "/W3")
+	list(APPEND MSVC_OPTIONS "/W3")
 	if(MSVC_VERSION GREATER  1900)  #Allow non fatal security wanrnings for msvc 2015
-		set(MSVC_OPTIONS "${MSVC_OPTIONS} /WX")
+		list(APPEND MSVC_OPTIONS "/WX")
 	endif()
     target_compile_options(${target_name} PRIVATE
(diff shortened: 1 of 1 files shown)


gabime:
The msvc 2015 fail..

trondhe:
Ah the joys of CMake string handling. Using the list api is probably more sound regardless.
I'm not sure what the travil ci issue is though.

cmake/utils.cmake
```diff
@@ -28,13 +28,15 @@ endfunction()
 
 # Turn on warnings on the given target
 function(spdlog_enable_warnings target_name)
```
gabime: Need to check if wee under msvc before creating this list

trondhe: Made some changes. You think this is good?

gabime: why all the whitespace changes all of the sudden?

trondhe: Ahh, sorry.
The code in the spdlog_enable_warnings function uses 4 spaces as indentation, but when I added the if statements for msvc, my vscode indented with tabs.
But it seems it did it that way to due mixed usage, didnt notice, so I just did a convert tabs to spaces inside the utils file.

I should probably revert the space/tabs thing and it should probably be done as a single pull request just changing tabs to 4 spaces or the other way around depending on your preference.

gabime:
Merged. Thanks.

PR #1521, opened 2020-04-18
add single logger method and log_msg constructor and tests/test_time_…

ron003:
…point.cpp
The test tests both the sink log method (log_msg) arg and logger log method.
Thanks,
Ron

### DIFF
--- a/include/spdlog/logger.h
+++ b/include/spdlog/logger.h
@@ -147,6 +147,19 @@ public:
         log(loc, lvl, string_view_t{msg});
     }
 
+    void log(log_clock::time_point &log_time, source_loc loc, level::level_enum lvl, string_view_t msg)
+    {
+        bool log_enabled = should_log(lvl);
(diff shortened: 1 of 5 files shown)


include/spdlog/logger.h
```diff
@@ -147,6 +147,19 @@ class SPDLOG_API logger
         log(loc, lvl, string_view_t{msg});
     }
 
+    void log(log_clock::time_point &log_time, source_loc loc, level::level_enum lvl, string_view_t msg)
```
gabime, line 150: I think log_time should passed by value here

ron003: I agree. That's a mistake. I know I had it once as a ref, but I thought I had changed it everywhere.
I missed this one. Sorry about this.
I pushed a change to ron003:single_log_method_with_time_point
Is that the proper way to rectify?
Thanks

gabime:
Merged. Thanks

PR #1522, opened 2020-04-20
Update to include funcname

matthesoundman:
When logging with ```SPDLOG_LOGGER_DEBUG``` macro the ```funcname``` is missing. 

- Updated ```full_formatter``` to include the ```funcname``` if source location is present.


gabime:
Thanks. Function name can be added using %! flag. It does not appear in the default format by design

matthesoundman:
Ok. It just looks strange with ```[file:line]``` enclosed in brackets with no way to add to brackets without a complete custom formatter.

PR #1525, opened 2020-04-20
Set minimum version of fmt to 5.0.0

Montellese:
The used `fmt::fmt` target has been introduced with `fmt 5.0.0`.

Without this `spdlog` also picks up older versions of `fmt` like `4.0.0` on Ubuntu 18.04 through `libfmt-dev` or even older versions on Ubuntu 16.04 through `libfmt3-dev`. Compilation fails but the error message is not straight forward at all.

I'm not sure if the version should be higher depending on what features of `fmt` are used within `spdlog`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,7 +188,7 @@ target_link_libraries(spdlog_header_only INTERFACE Threads::Threads)
 #---------------------------------------------------------------------------------------
 if(SPDLOG_FMT_EXTERNAL OR SPDLOG_FMT_EXTERNAL_HO)
     if (NOT TARGET fmt::fmt)
-        find_package(fmt REQUIRED)
+        find_package(fmt 5.3.0 REQUIRED)
     endif ()
     target_compile_definitions(spdlog PUBLIC SPDLOG_FMT_EXTERNAL)
     target_compile_definitions(spdlog_header_only INTERFACE SPDLOG_FMT_EXTERNAL)


gabime:
I don't think this would work for fmt 6.x.x versions (incompatible with 5 according to cmake)

e-i-n-s:
@gabime 

I can confirm that it will work. I was confused, too :-)

I used CMake to generate the CMake target of `fmt`:
```
cmake -G "Visual Studio 15 2017" <path to fmt repo> -DCMAKE_INSTALL_PREFIX=install
cmake --build . --target install
```

The generated `<cmake install path>\lib\cmake\fmt\fmt-config-version.cmake` contain the version:
```
set(PACKAGE_VERSION "6.2.1")
```

Now I'm able to consume this CMake target (Windows):
```
SET fmt_DIR=<cmake install path >\lib\cmake\fmt
cmake -G "Visual Studio 15 2017" <path to spdlog repo>  -DSPDLOG_FMT_EXTERNAL=On ..
```

Works fine with `find_package(fmt 5.0.0 REQUIRED)`

As a test I switched to `find_package(fmt 7.0.0 REQUIRED)`. It gave a correct error message:
```
-- Build spdlog: 1.6.0
-- Build type: Release
CMake Error at CMakeLists.txt:191 (find_package) in the spdlog repository:
  Could not find a configuration file for package "fmt" that is compatible
  with requested version "7.0.0".

    <path to spdlog repo>/lib/cmake/fmt/fmt-config.cmake, version: 6.2.1
```

gabime:
This is indeed confusing. This is what cmake docs say:

> When the [version] argument is given Config mode will only find a version of the package that claims compatibility with the requested version (format is major[.minor[.patch[.tweak]]]). If the EXACT option is given only a version of the package claiming an exact match of the requested version may be found. CMake does not establish any convention for the meaning of version numbers. Package version numbers are checked by “version” files provided by the packages themselves. For a candidate package configuration file <config-file>.cmake the corresponding version file is located next to it and named either <config-file>-version.cmake or <config-file>Version.cmake. If no such version file is available then the configuration file is assumed to not be compatible with any requested version. A basic version file containing generic version matching code can be created using the CMakePackageConfigHelpers module. When a version file is found it is loaded to check the requested version number. 

So I still not sure what this means :(

Montellese:
@gabime Yeah I didn't find any understandable description of how CMake handles these versions and how to apply rules like `>=` or `<=`.
In Kodi we are using `fmt 6.1.2` and it works fine there.

According to one of our users `spdlog` is not compatible with `fmt 5.1.0` so maybe we need a different version here. Do you have any idea?

wsnipex:
snip from fmt-config-version.cmake

`set(PACKAGE_VERSION "6.1.2")

if(PACKAGE_VERSION VERSION_LESS PACKAGE_FIND_VERSION)
  set(PACKAGE_VERSION_COMPATIBLE FALSE)
else()
  set(PACKAGE_VERSION_COMPATIBLE TRUE)
  if(PACKAGE_FIND_VERSION STREQUAL PACKAGE_VERSION)
    set(PACKAGE_VERSION_EXACT TRUE)
  endif()
endif()
`

gabime:
@wsnipex whatever that means :) I have no idea

gabime:
@Montellese Not sure  but 5.2.0 might be ok

Montellese:
I just went through the `fmt` releases starting with `5.1.0` and built `spdlog 1.5.0` against it with the following results:
* `5.1.0` doesn't compile
* `5.2.0` doesn't compile
* `5.2.1` doesn't compile
* `5.3.0` compiles and all unit tests pass.
* `6.0.0` compiles and all unit tests pass.
* `6.1.0` compiles and all unit tests pass.

So I'm assuming that the minimum version is `5.3.0`. I will change it to that.

gabime:
Good to know.
Thanks for  investigating!

PR #1563, opened 2020-05-20
Fixed ringbuffer_sink last_raw and last_formatted

vekkuli:
Fixed ringbuffer_sink last_raw and last_formatted giving the first lim number of elements instead of last.


gabime:
please see https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/details/circular_q.h#L94

the at() function it already takes care if this offsetting the given index properly

gabime:
Merged. Thanks @vekkuli

Issue #1566, opened 2020-05-22
Old files not deleted when max reached and app restarts


IIFE:
Hi.

This is sort of a follow up to #1531 

I have daily file sink setup so that it creates a new file every day (rotation hour 0) and keep a max of 7 files.

I tried to test the behaviour by moving the PC date forward one day. A new files gets created but no old files are deleted when the count reaches 8 or above. **Note:** I restart my app that uses spdlog after changing the date. Does the daily sink need the app to keep running in order to delete old files, or should it work when app is restarted?

tt4g:
@IIFE AFAIK, daily sink delete old filed when calling `spdlog::sinks::daily_file_sink::dink_it_()`.
The constructor of `spdlog::sinks::daily_file_sink` doesn't seem to do any rotate deletion.

https://github.com/gabime/spdlog/blob/83b9149930f392d7797b54fe97a66ab3f2120671/include/spdlog/sinks/daily_file_sink.h#L80-L99

In order to avoid throwing an exception by deleting the file in the daily_sink constructor, it is designed to delete the file only when the log is output (Note that I'm writing this as a guess since I didn't add this feature).

https://github.com/gabime/spdlog/blob/83b9149930f392d7797b54fe97a66ab3f2120671/include/spdlog/sinks/daily_file_sink.h#L151-L153

Maybe your application didn't output any logs during testing.

IIFE:
@tt4g Thanks. My app is definitely logging messages on start up. I have debugged the code and the sinkt_it function gets called, however the should_rotate flag is always false.

I don't know if the daily file sink expects the app to continue running in order for old files to be deleted, and it's not handling the case where app is restarted. It's not very clear from the code.

tt4g:
Oh, I'm sorry I missed something.
`should_rotate` is `true` if the time of the message indicates a time in the future from the next rotation time.
Only at this time, the old rotated files will be deleted.
In other words, the application must meet the rotation time as it continues to run.

IIFE:
@tt4g Indeed, however the ctor when setting the next rotation time (next_rotation_tp_) for some reason adds 24 hours when the rotation time is after the current time, i.e.:

```
    log_clock::time_point next_rotation_tp_()
    {
        auto now = log_clock::now();
        tm date = now_tm(now);
        date.tm_hour = rotation_h_;
        date.tm_min = rotation_m_;
        date.tm_sec = 0;
        auto rotation_time = log_clock::from_time_t(std::mktime(&date));
[...]

PR #1567, opened 2020-05-24
Delete old files on init

IIFE:
- Current init of file names queue does not cater for case where the logger is initialised several days after the last log (all it does is go back day by day and stop if the file does not exist). Changed this to ensure we cater for all files in the log directory.
- Extract the YYYY-MM-DD component of the file name to allow lexicographical ordering by inserting key/value pair of date to file name in a map.

Solves issue #1566

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -98,10 +98,13 @@ using win32_find_data = WIN32_FIND_DATAW;
 #define SPDLOG_FILENAME_T(s) L##s
 #else
 using filename_t = std::string;
+#define SPDLOG_FILENAME_T(s) s
+#endif
+
+#if defined(_WIN32) && !defined(SPDLOG_WCHAR_FILENAMES)
 using win32_find_data = WIN32_FIND_DATAA;
 #define find_first_file  FindFirstFileA
 #define find_next_file  FindNextFileA
-#define SPDLOG_FILENAME_T(s) s
 #endif
 
 using log_clock = std::chrono::system_clock;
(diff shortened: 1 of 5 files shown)


IIFE:
@tt4g Thanks for the suggestions. All applied now.

P.S. There was no need to include `<utility>`

IIFE:
@tt4g Hi. Was just wondering is there any timeline on when this can be merged please so that I can take the latest spdlog version? Thanks.

tt4g:
@IIFE I don't have the right to approve it.
Let's wait for @gabime review.

IIFE:
@tt4g @gabime Just pushed a new change as I forgot to use the template name for FileNameCalc.

gabime:
@IIFE Thanks for the pr. After thinking about it for a while I came to the conclusion  that the  additional complexity just not worth it. This sink is supposed to be very simple and it is only in charge of cleaning files created in the current object (hence circular memory q). If a process ends and get restarted by the operator, then the operator can easily write cleaning script if needed.

IIFE:
@gabime OK thanks no problem, I'll just implement a sink with the desired behaviour.

One thing I would say is maybe worth updating the documentation to explicitly mention that the daily file sink does not clean up on app restart, as I don't think it's very obvious without reading the source code (hence why I used it with the assumption that it did).

gabime:
sure. good idea

themightyoarfish:
Is the documentation updated? I would have expected the hourly and daily sinks to clear old files. After all, why have a `max_files` parameter, when the number of retained files can exceed it by simply restarting the program?
I could not find any documentation on this behaviour.

gabime:
Feel free to document it in the wiki (it is free for all), or open PR to document it elsewhere in the codebase

PR #1571, opened 2020-05-26
fix deadlock on filename() call

### LINKED ISSUE #1570: daily_file_sink has deadlock bug
The `sink_it` function of `daily_file_sink` runs under a lock, however it can call `delete_old_`, which internally calls `filename()`. `filename()` tries to acquire the lock, which is already held open. This leads to deadlock.

Happy to fix.

IIFE:
Fixes #1570

- rotating-file-sink was not actually using filename() at all anyway, so remove.
- daily-file-sink can just use file_helper_.filename()

### DIFF
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -70,12 +70,6 @@ public:
         }
     }
 
-    filename_t filename()
-    {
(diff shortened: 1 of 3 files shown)


IIFE:
Didn't realise they were public. Put them back.

Issue #1369, opened 2019-12-23
Compile time formating checking(FMT_STRING)?


nanoric:
fmt library support compile-time format error checking using FMT_STRING.
can spdlog enable this?

gabime:
Maybe in one of the future versions.

rabbull:
Seems current version of fmt embedded does not support compile-time format checking, so it's necessary to upgrade fmt to the latest version and test for compatibility first. If needed I'd like to prepaire a pr for it.

gabime:
> Seems current version of fmt embedded does not support compile-time format checking, so it's necessary to upgrade fmt to the latest version

No sure I understand. spdlog already  uses the latest version of fmt - 6.1.2

rabbull:
> > Seems current version of fmt embedded does not support compile-time format checking, so it's necessary to upgrade fmt to the latest version
> 
> No sure I understand. spdlog already uses the latest version of fmt - 6.1.2

Hmm that's awkward. Sry for the mistake. So is there anyone working on this feature? May I take this on?

gabime:
Sure. help is welcome.

A possible  approach would to add a new macro in common.h that defaults to nothing or set to FMT_STRING by the user if desired. Then the fmt formatting code in logger.h would wrap the format string with this new macro.

rabbull:
Bad news. I noticed all strings are finally converted into `string_view_t` before formatting, but the `FMT_STRING` macro does not support `string_view_t` (aka `fmt::basic_string_view<char>`) variable as its param.

Example:
```C++
#include <iostream>
#include <spdlog/spdlog.h>
int main() {
    using string_view_t = fmt::basic_string_view<char>;
    std::cout << fmt::format(FMT_STRING(string_view_t{"{}"}), 150) << std::endl;
}
```

While being compiled MinGW-w64 gives (haven't tried on the other compilers)
```
...
error: could not convert '{fmt::v6::basic_string_view<char>(((const char*)"{}")), ((sizeof (fmt::v6::basic_string_view<char>(((const char*)"{}"))) / sizeof (main(int, char**)::<lambda()>::str::char_type)) - 1)}' from '<brace-enclosed initializer list>' to 'fmt::v6::basic_string_view<fmt::v6::basic_string_view<char> >'
...
```

[...]

PR #1574, opened 2020-05-31
Add support for FMT_STRING compile time checking.

Tridacnid:
Additionally, add negative compilation unit test for compilers and c++ standard that support relaxed constexpr.

Hello, I discovered that the FMT_STRING compile time checking wasn't supported and figured I wanted it enough to add it. After the fact I found #1369, hopefully this pull request addresses all concerns by the maintainers. If not, please let me know what I can do to improve the design to make the pull more appealing.

I'm a Linux developer but I did my best to check the Windows support by booting to Windows10 and running the unit tests in Visual Studio with `-DSPDLOG_BUILD_TESTS=ON -DSPDLOG_BUILD_WARNINGS=ON -DSPDLOG_WCHAR_SUPPORT=ON`


Tridacnid:
I changed quite a few things to perfect forwarding.
```
    template<typename T>
    void trace(T &&msg)
    {
        log(level::trace, std::forward<T>(msg));
    }
```

If this is not desirable, let me know and I can switch them back.

include/spdlog/logger.h
```diff
@@ -73,78 +73,74 @@ class SPDLOG_API logger
...
+    void log(level::level_enum lvl, T &&msg)
     {
-        log(source_loc{}, lvl, msg);
+        log(source_loc{}, lvl, std::forward<T>(msg));
     }
 
     // T can be statically converted to string_view
-    template<class T, typename std::enable_if<std::is_convertible<const T &, spdlog::string_view_t>::value, T>::type * = nullptr>
-    void log(source_loc loc, level::level_enum lvl, const T &msg)
+    template<class T, typename std::enable_if<std::is_convertible<T &&, spdlog::string_view_t>::value, T>::type * = nullptr>
```
Tridacnid, line 140: SFINAE + universal reference?

include/spdlog/logger.h
```diff
@@ -179,48 +175,48 @@ class SPDLOG_API logger
     }
 
     // T cannot be statically converted to string_view or wstring_view
-    template<class T, typename std::enable_if<!std::is_convertible<const T &, spdlog::string_view_t>::value &&
-                                                  !is_convertible_to_wstring_view<const T &>::value,
-                          T>::type * = nullptr>
-    void log(source_loc loc, level::level_enum lvl, const T &msg)
+    template<class T,
+        typename std::enable_if<!std::is_convertible<T &&, spdlog::string_view_t>::value && !is_convertible_to_wstring_view<T &&>::value,
```
Tridacnid, line 179: SFINAE + universal reference?

include/spdlog/logger.h
```diff
@@ -251,51 +247,9 @@ class SPDLOG_API logger
...
-    template<typename... Args>
-    void critical(wstring_view_t fmt, const Args &... args)
-    {
-        log(level::critical, fmt, args...);
-    }
-
     // T can be statically converted to wstring_view
-    template<class T, typename std::enable_if<is_convertible_to_wstring_view<const T &>::value, T>::type * = nullptr>
-    void log(source_loc loc, level::level_enum lvl, const T &msg)
+    template<class T, typename std::enable_if<is_convertible_to_wstring_view<T &&>::value, T>::type * = nullptr>
```
Tridacnid, line 251: SFINAE + universal reference?

Tridacnid:
I have marked a couple spots I'm unsure about, I've never combined universal references and SFINAE so I'm not sure my change is 100% correct in these places.

gabime:
> If this is not desirable, let me know and I can switch them back.

Probably not needed. Lets concentrate on the FMT_STRING and remove other non related changes. universal references are very confusing and i prefer to avoid them if not really needed.

Tridacnid:
> > If this is not desirable, let me know and I can switch them back.
> 
> Probably not needed. Lets concentrate on the FMT_STRING and remove other non related changes. universal references are very confusing and i prefer to avoid them if not really needed.

In hindsight,
```
    template<typename T>
    void trace(T &&msg)
    {
        log(level::trace, std::forward<T>(msg));
    }
```
was a bad example, because we'd want to catch 
```
spdlog::trace(FMT_STRING("I have too many expected params {}!"));
```

Tridacnid:
> > If this is not desirable, let me know and I can switch them back.
> 
> Probably not needed. Lets concentrate on the FMT_STRING and remove other non related changes. universal references are very confusing and i prefer to avoid them if not really needed.

Can do. If avoiding universal references is a goal I can probably remove them everywhere, including the templated FormatString type I added.

Tridacnid:
I'd like to update the CI to run the negative compilation test when possible (clang10, gcc7). Is something like this acceptable?

```
    - env: GCC_VERSION=7 BUILD_TYPE=Release CPP=14
      os: linux
      addons: *gcc7

# further down, in script
script:
    ...
    -DSPDLOG_BUILD_FAILING_TESTS=ON \ #Activates negative test when C++ version >11 and compiler supports relaxed constexpr
    ...
```

gabime:
spdlog is c++11 so lets not. Also, those tests are probably more relevant to fmt than to spdlog.

Tridacnid:
> spdlog is c++11 so lets not. Also, those tests are probably more relevant to fmt than to spdlog.

If that's the direction you'd like to take I'll defer to you. I added it so that a regression in the API could be detected when using FMT_STRING.

Tridacnid:
Do you even want the FMT_STRING compilation error test? If it isn't going to run with CI I worry it would fall out of maintenance and become a burden.

gabime:
Please remove the test. why c++14 btw? This feature wont work in c++11?

Tridacnid:
FMT_STRING needs relaxed constexpr to work correctly. In c++11 it falls back to being a runtime check.

gabime:
Ok. so after removing the universal refs and tests we this would be much smaller pr with minimal changes 2 fo 2 files only: logger.h  spdlog.h. Waiting for your fixes

Tridacnid:
Sounds like a plan, I'll try to get the updates pushed in the next day or two.

Tridacnid:
Updated.

gabime:
Thanks. Why the universal ref for FormatString?

Tridacnid:
I debated with myself whether you wanted me to eliminate those as well, sounds like I should have.

No reason other than that's how I originally implemented it, I'll run it to ground this evening.

Tridacnid:
Updated.

gabime:
Merged. Thanks @Tridacnid

PR #1580, opened 2020-06-04
Add support for custom filename calculator in rotating_file_sink.

TamasFlorin:
Hello,

I have recently started replacing an old implementation of our logger for one of the projects and so far it's being going great.
However, I do have one use-cases that is not supported out of the box, specifically, setting a custom filename for the "rotating_file_sink" (eg: we want to have "output1.log" instead of "output.1.log").

Please note that this was implemented following the model from "daily_file_sink".
The code was compiled with `-DSPDLOG_BUILD_TESTS=ON -SPDLOG_BUILD_WARNINGS=ON`.


gabime:
Thanks.  I think `#include <spdlog/details/os.h>` is needed.

TamasFlorin:
> Thanks. I think `#include <spdlog/details/os.h>` is needed.

Good catch! Thanks

gabime:
Merged. Thanks @TamasFlorin

gabime:
@TamasFlorin Unfortunately this broke the ABI compatibility with 1.6.0 so I had to revert.

Seems that this will have to wait to version 2.x 

Sorry.

TamasFlorin:
> @TamasFlorin Unfortunately this broke the ABI compatibility with the 1.x branch so I had to revert.
> 
> 
> 
> Seems that this will have to wait to version 2.x
> 
> 
> 
> Sorry.

Thanks for letting me know @gabime 
I there anything that I have to do on my side?

gabime:
You can reproduce this using the following steps:

1. Compile the example with version 1.6.1 using -DSPDLOG_BUILD_SHARED=ON to build `libspdlog.1.6.1.so`
2. Checkout your version and compile again with -DSPDLOG_BUILD_SHARED to make new version of .so lib.
3. Replace the .so from 1 with the .so from 2 and try to execute the example **without**  recompiling it

PR #1591, opened 2020-06-11
Cleanup options-related lines in CMakeLists.txt

Pospelove:


gabime:
Thanks. Why the name change of SPDLOG_WCHAR_TO_UTF8_SUPPORT?

Pospelove:
Only for consistency purposes. Other definitions have the same name as the CMake options that enables them. Changing `SPDLOG_WCHAR_TO_UTF8_SUPPORT` to `SPDLOG_WCHAR_SUPPORT` allows to do this:
```cmake
if(${OPTION})
    target_compile_definitions(spdlog PUBLIC ${OPTION})
    target_compile_definitions(spdlog_header_only INTERFACE ${OPTION})
endif()
```

However, I can change it back and handle this option another way if it's important.

CMakeLists.txt
```diff
@@ -199,49 +199,24 @@ endif()
...
-endif()
-
-if(SPDLOG_NO_EXCEPTIONS)
-    target_compile_definitions(spdlog PUBLIC SPDLOG_NO_EXCEPTIONS)
-
-    target_compile_definitions(spdlog_header_only INTERFACE SPDLOG_NO_EXCEPTIONS)
-
-    if(NOT MSVC)
-        target_compile_options(spdlog PRIVATE -fno-exceptions)
+foreach(OPTION
```
tt4g, line 202: Suggestion: The variable name `OPTION` has a name conflict with CMake command `option()`, so I recommend not to use it.

gabime:
I agree with @tt4g . lets replace `OPTION` with different word to avoid collisions. 
As for   `SPDLOG_WCHAR_TO_UTF8_SUPPORT`, lets not replace it because it can break project builds that depend on it.

Pospelove:
Done

PR #1605, opened 2020-07-06
Define FMT_USE_INTERNAL for fmt 7.0.0

candrews:
fmt 7.0.0 renamed the internal namespace to detail. The former is still provided as an alias if the FMT_USE_INTERNAL macro is defined.

See: https://github.com/fmtlib/fmt/issues/1538


candrews:
https://github.com/gabime/spdlog/pull/1606 is a better approach; closing this PR in favor of that one.

PR #1616, opened 2020-07-10
The ability to specify multiple color range formatting regions when setting the logger's pattern.

virtualthoughts:
One additional commit to allow for constructing a logger with a std::vector of sinks.


gabime:
Adding std::vector for each message would kill performance (dynamic allocation per log).
So I tend to keep it simple and allow only one color range

PR #1621, opened 2020-07-14
Allow multiple ranges... take two

virtualthoughts:
I modified my code to use simple arrays.  I understand the desire to avoid allocations as I come from an embedded background.  I added an entry to tweakme.h for "MAX_RANGES" which I defaulted to 3.  I updated the formatter benchmark with a few more complex patterns that included 1, 2, and 3 color ranges to see how it affected speed.  I ran the new bench mark on your original code, my vector code (the slowest), and the new array based code (comparable to your original).  As is the nature with benchmarking the results vary from run to run, but are within about 0-8ns of your original code.  I also noticed that when I build it seems like the windows colored output file doesn't seem to get compiled as I'm running linux.  I don't have a windows machine handy to test that on, so those changes are a best effort on my part.  Oh I also added a test case as it looked like if I didn't specify any range then nothing would get colored which seemed wrong, so now that is addressed as well.

Hopefully these changes mitigate your performance concerns with my previous attempt.


gabime:
Thanks. However I think this feature doesn't worth the extra complexity. Singke range is good enough for the most part.

PR #1622, opened 2020-07-16
Use cmake to generate version.h

flagarde:
This commit allows to create version.h using the version given in CMake project command.


gabime:
Thanks. Problem it will break projects that use spdlog without cmake(e.g. use as header only)
Please open this PR int the v2.x branch.

flagarde:
Hi, Yes good point !

PR #1626, opened 2020-07-22
Bugfix/shadow warning

Steranoid:
We are trying to include your project in a project of our company (Eaton) but we are using a longer list of warnings in our source files, when we are trying to include your headers, we only have a warning that raise about some variables shadowing some type. This PR add the warning in the spdlog project, fix it and change a little the CMakeLists so when project try to import spdlog, these projects compiler and linter will not consider to raise warning on spdlog headers, considered as system headers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,8 +143,10 @@ endif()
 add_library(spdlog::spdlog ALIAS spdlog)
 
 target_compile_definitions(spdlog PUBLIC SPDLOG_COMPILED_LIB)
-target_include_directories(spdlog PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
-                                         "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
+target_include_directories(spdlog PRIVATE "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+                                          "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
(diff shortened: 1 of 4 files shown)


phlptp:
We were running into a few of these issues as well as we started to integrate spdlog and I think this would address some of the warnings we were seeing.  👍

gabime:
Thanks. The fmt changes belong to the fmt lib project, so please open  the pr there (https://github.com/fmtlib/fmt). About marking spdlov headers as system, not sure! spdlog is not a “system” library after all

saierd:
FIY, fmt has already fixed these warnings for some time, see https://github.com/fmtlib/fmt/commit/c3fa33314045cb4a132eef9a94524fb073a1e696 and https://github.com/fmtlib/fmt/commit/84898b4626a09b2a08daa2523775aa7fa5631373, so I guess it just needs to be updated.

The point of marking the include paths as system headers was to avoid leaking warnings when building this library in another CMake build (using `add_subdirectory`). In that case the problem also occurs when using the compiled version, since the user cannot mark the spdlog include directory as system (he just uses the spdlog target provided by your CMake build).

To make such a use case possible, you should either make sure that the headers are as warning-free as possible (not just with the warning flags you have enabled) or provide a way to mark the headers as system includes. This can be done conditionally by checking whether the CMake file was included by some other project (see e.g. the changes fmt uses in their CMake build [here](https://github.com/fmtlib/fmt/blob/61602a75db980737870e7f5e91de99e49dcd1c9e/CMakeLists.txt#L10)).

Steranoid:
As you can see the warning is raised by headers themselves so you just have to include these headers in your source files to have this warning raised.

Issue #1627, opened 2020-07-22
Have a global level(), and should_log() function


eyalroz:
spdlog has `spdlog::set_level()` - but no `should_log()` and `level()`. It should have those!

... for `level()`, we can't use that particular name since it's taken up by a namespace, so let's make it `get_level()` instead.

PR #1628, opened 2020-07-22
Fix for issue #1627: Global get_level() and should_log()

eyalroz:
* Added: `spdlog::get_level()` API function - like `logger::level()`, except for the name change
* Added: `spdlog::should_log()` API function - like `logger: should_log()`
* Added some code to the example program to use the two new API functions.

### DIFF
--- a/include/spdlog/spdlog.h
+++ b/include/spdlog/spdlog.h
@@ -67,8 +67,14 @@ SPDLOG_API void disable_backtrace();
 // call dump backtrace on default logger
 SPDLOG_API void dump_backtrace();
 
+// Get global logging level
+SPDLOG_API level::level_enum get_level();
+
 // Set global logging level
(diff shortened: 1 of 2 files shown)


gabime:
Thanks. Lets omit the example changes. The example is already way too bi’m g

eyalroz:
@gabime : As per your request, rolled back the changes to the example.

include/spdlog/spdlog.h
```diff
@@ -217,6 +223,17 @@ inline void critical(const T &msg)
     default_logger_raw()->critical(msg);
 }
 
+/*
```
gabime, line 226: Please remove this as well

eyalroz: Oh, sorry :-P

Issue #1639, opened 2020-08-05
[macOS] tcp_sink.h not compileable in macOS


dkruempe:
I just tested the excellent library in linux and macOS. Actually nearly all is working. Except the tcp_sink.h in macOS. Currently I got every time the message that SOCK_CLOEXEC is an undefined identifier. 

As far as I see is the mentioned part in tcp_client.h.

```
int const flags = SOCK_CLOEXEC;
socket_ = ::socket(rp->ai_family, rp->ai_socktype | flags, rp->ai_protocol);
```

Actually the SOCK_CLOEXEC doesn't exists in sys/socket.h. So, it won't compile in macOS currently.

To make the part compilable again I have used the following workaround 

```
int flags = 0;
#ifdef SOCK_CLOEXEC
            flags = SOCK_CLOEXEC;
#endif
socket_ = ::socket(rp->ai_family, rp->ai_socktype | flags, rp->ai_protocol);
```

Actually I am not completely sure if it works 100%.

gabime:
Thanks for reporting. If this works for you a PR would be welcome.

PR #1640, opened 2020-08-06
tcp_client.h bugfix for macOS and add tcp example

dkruempe:
-> fix the issue that the head is not compilable bc. of missing declaration
-> optimize if/else block in connection function. Else not needed bc. of break in if


dkruempe:
Ok, understand so far. I have reset the example file to the original one.

gabime:
Fixes #1639

PR #1106, opened 2019-06-11
Adding Bazel Build rules

cgrinker:
This PR adds [bazel](https://bazel.build/) build rules to spdlog, allowing it to be used as a dependency in bazel projects.

## Usage
Given a correctly installed bazel and msvc/gcc/clang
### WORKSPACE file
```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "spdlog",
    urls = ["https://github.com/cgrinker/spdlog/archive/3bf4a07dc286e10fa32d6503fded647c9ee02d26.zip"],
    strip_prefix = "spdlog-3bf4a07dc286e10fa32d6503fded647c9ee02d26",
    sha256 = "d00ccd202e2abae832441b9121330a4b8b82fdc60564edb315c89506fa3772c3",
)
```

### BUILD file
```
cc_binary(
    name="example",
    srcs=[
      "main.cpp"
    ],
    deps=[
        "@spdlog//:headers",
    ]
)

```

### main.cpp
```cpp
#include <spdlog/spdlog.h>
#include <spdlog/fmt/ostr.h>
#include "spdlog/sinks/stdout_color_sinks.h"

int main(int argc, char** argv)
{
    auto logger = spdlog::stdout_color_mt("example");
    logger->info("Hello With Bazel");
    return 0;
}
```

PR #1645, opened 2020-08-09
Add support for Bazel

Vertexwahn:
This PR helps users of the Bazel build system to integrate `spdlog` in their builds.

In the past it was tried to establish Bazel also as a build system for `spdlog`, but the corresponding [PR](https://github.com/gabime/spdlog/pull/1106) did never went in.

The idea of the [PR](https://github.com/gabime/spdlog/pull/1106) is to make it for users of Bazel as simple as possible to integrate `spdlog`. This would be in my opinion the better solution. I think supporting two build systems (CMake and Bazel) is of course more effort to maintain, but helps to spread `spdlog`.


gabime:
Thanks. I tried before supporting meson in addition with cmake and it proved problematic and very time consuming to support multiple builds. I suggest open a dedicated repo for this instead.

Vertexwahn:
In your case I would only maintain CMake and do not care about Bazel. Neverthless, I would allow PR that dd Bazel files (e.g. `WORKSPACE.bazel`, `BUILD.bazel`) - what do you fear - in the worst case the Bazel build does simply not work - in the best case someone who uses Bazel can benefit from it. Also other project support CMake and Bazel side by side (e.g. [Catch2](https://github.com/catchorg/Catch2))

yesudeep:
I second this. Bazel has sped up my builds several times and reduced the cognitive overload required to understand the build graph (the dependencies are explicit).

gabime:
Guys you are welcome to create a dedicated repo for this. Takes a few seconds.

PR #1646, opened 2020-08-11
Error when include boost

BoilTask:
#define WIN32_LEAN_AND_MEAN


tt4g:
`include/spdlog/fmt/bundled/format-inl.h` is copied from fmt project. so it's a problem with the fmt library.
It is recommended to create pull request in [fmt repo](https://github.com/fmtlib/fmt).

PR #1652, opened 2020-08-20
Fix the MSVC warning C4566

ccwanggl:
referenced from fmtlib


tt4g:
This is fmt library problem and already fixed in fmt 7.0.0 (fmtlib/fmt@36ea32640fecc775d16fd337c447762bede3eac1).
You can use external fmt or wait for the bundled fmt update.

PR #1688, opened 2020-10-01
Replace all header file with one header file (#include<bits/stdc++.h>)

Tk-Rwing:


gabime:
Thanks but it is not stanard.

Issue #1225, opened 2019-09-12
Enhancement suggestion: output of UTF-8 BOM into rotating_file_sink file to make logs natively supporting Unicode viewing


martinjansa:
Hi,

Firstly, thank you for a great library! :-)

We are using it for storing the logging information using the UTF-8 format, but sometimes the tools we are using (on Windows) for viewing the logs don't show the specific characters correctly, when the UTF-8 encoding is not recognized or manually set. It would be very helpfull, if the log file could contain a 3-byte UTF-8 identifier (BOM), that would suggest the tools the encoding of the content.

@gabime, would this be something you could be willing to include into your repo? If so, I would be willing to contribute in getting it done.

So far, I've made a fork of your repo and implemented this functionality into a branch and verified everything works fine. As you can see at https://github.com/martinjansa/spdlog/commit/ff5495ee2bc2d925c4907f951db2472b95c2fc9c, the interface is fully backwards compatible. 

Is it going the right way? If so, how shall we proceed further?

Thank you,

Martin

gabime:
Hi, 
Thanks for the nice feedback :+1: 

There are numerous requests regarding adding something automatically at the start of each file (e.g.  adding speacial header line on start of each new file).

This is why I try to come up with an API or some config struct for all file sinks.

Something like:

```c++
sink->after_open(std::funcion<void (FILE* fd)> callback_fun);

//And maybe
sink->before_close(std::funcion<void (FILE* fd)> callback_fun);

// And even maybe
sink->after_close(std::funcion<void (std::string& filename)> callback_fun);
```

Of course, any help on formalizing such API would be appreciated

martinjansa:
Perfect, sounds great. I'll then wait for the <config> support to come out and use it for this purpose. Please, let me know, if there's anything I could help in that.

m4ce:
@gabime, I think the API spec you put down works really well. The `after_rotate` is exactly what I am looking for in #1181.

gabime:
spdlog now provides an option to get callbacks before/after a log file is created/closed.

```cpp
    // pass the spdlog::file_event_handlers to file sinks for open/close log file notifications
    spdlog::file_event_handlers handlers;
    handlers.before_open = [](spdlog::filename_t filename) { spdlog::info("Before opening {}", filename); };
    handlers.after_open = [](spdlog::filename_t filename, std::FILE *fstream) {
        spdlog::info("After opening {}", filename);
        fputs("After opening\n", fstream);
    };
[...]

Issue #1181, opened 2019-08-09
add compression to rotating file sink


m4ce:
This is not an issue but more of a feature request. It would be really nice if the rotating file sink would optionally compress the rotated log file, let's say to gzip?

log4j2 has a similar feature in its RotatingFile appender: 

```    <RollingFile name="RollingFile" fileName="logs/app.log"
                 filePattern="logs/$${date:yyyy-MM}/app-%d{MM-dd-yyyy}-%i.log.gz">
```

when the `filePattern` ends with `gz`, it will automatically compress it for you.

artemyv:
On windows I'd like to have zip possibility

gabime:
Depending on a zip/gzip/tar/your favourite compression is pretty big. 

A better way would be to provide the option to prodie a function what would be called after each rotation (along with the rotated filenames). 

something like 
```c++
void on_rotated(std::vector<std::string> rotated_filenames);
```

This way, everyone is happy and can use its own favourite cleaning mechanism.

m4ce:
@gabime, that would work well! Would we pass that as a std::function when instantiating the rotating file sink?

gabime:
something like that, yes

m4ce:
@gabime - that would be really useful. Any chance to get this feature in one of the next releases? :)

gabime:
I hope so

m4ce:
@gabime, by the way, I was thinking about this and I was wondering how to go about a possible scenario.

Assume that we rotate the files and then we call `on_rotated`. Also assume that the user will go through the rotated files and gzip them, they will be renamed to `<log>.gz`. When the next rotation runs, the files will be rotated using the old index without taking into consideration the compressed ones.

I guess one way to solve this is to also optionally provide the user with a function to determine the new file name? Open to ideas..

m4ce:
@gabime, in the `rotating_file_sink<Mutex>::rotate_()`, there's a loop that takes care of rotating the files. However, the `calc_filename` will not take into consideration the `.gz` extension when building the `filename.<index>log`..

gabime:
I don't think there is an issue If I understand correctly.

For example:

After the callback only the following files exist: 
```
log.txt (active file, cannot be zipped)
log.1.zip, 
log.2.zip
```

After on the next rotate spdlog will only file "log.txt" and will rotate it to "log1.txt", so the callback will get only "log.1.txt" to zip and after this procedure we get:

```
log.txt,
log1.zip (new)
[...]

PR #1693, opened 2020-10-02
Rotating file sink callback

dilshodm:
### Added callback to rotating_file_sink

It's possible to use callback for manipulating (compressing, copying somewhere, etc) the newly completed and rotated log file (essentially rotating.1.txt in this example). If that callback compresses and appends extension (ex. .gz) it has to be mentioned in the parameters, otherwise keep it empty string. The callback can be any functional which accepts spdlog::filename_t parameter to manipulate it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,22 @@ void rotating_example()
 }
 ```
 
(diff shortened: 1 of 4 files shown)


gabime:
Thanks for the PR. 
Related to #1225 and #1181 that still need to be solved.
Meanwhile I prefer not to merge, as it complicates the (already complex) rotating sink.

PR #1709, opened 2020-10-17
Download googlebenchmark if not found 

kitattyor:
Currently building with `-DSPDLOG_BUILD_ALL=1` fails if googlebench is not installed on the user system (because `bench` depends on googlebench). CMake FetchContent can be used to download and build it (also test it, but I disabled that option) so that the bench build doesn't fail. FetchContent is however only available on CMake 3.11+, hence the hacky version check.

Obviously, the cleaner solution is to bump CMake minimum to 3.11, but I didn't change that because https://github.com/gabime/spdlog/pull/1624

### DIFF
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -9,7 +9,24 @@ if(NOT TARGET spdlog)
 endif()
 
 find_package(Threads REQUIRED)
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. Merged

PR #1711, opened 2020-10-20
Fix typo in comment

ChristofKaufmann:
Replacing a wrong `->` by `.` in a code comment. In the wiki the [same example](https://github.com/gabime/spdlog/wiki/1.1.-Thread-Safety#non-thread-safe-functions) is correct.


gabime:
Thanks.

PR #1712, opened 2020-10-20
Add CPack debian package settings

ChristofKaufmann:
The included CPack file only contained RPM package information. Yet, it was possible to build a debian package using `cpack -G DEB`, but information like description, homepage were incomplete and the filename was different than intended (compared with the RPM). This commit adds `CPACK_DEBIAN_*` settings to the CPack file.

Also `CPACK_RPM_PACKAGE_NAME` was missing as well as `CPACK_DEBIAN_PACKAGE_NAME`.


gabime:
Merged. Thanks @ChristofKaufmann

PR #1740, opened 2020-11-18
Add logger context support

o2gy84:
See examples and unit-tests.

The main idea is we need something that will not share between loggers.
So, we can create something like hierarchy of loggers.

There is something in common with go-style logging (which is almost perfect).


gabime:
Thanks. Interesting approach. Seems not thread safe though..

o2gy84:
By the way, we already use this patch into production. And no issues with it.

o2gy84:
So, can I do something to move this patch into the trunk?  
Or we can just close this PR?

gabime:
Lets close this for now. Thanks anyway !

PR #1742, opened 2020-11-19
Add missing include

jwittbrodt:
Hi,

I noticed a missing `<algorithm>` include in `helpers-inl.h`.  It uses std::transform in l. 26 which I assume it previously (for me) got through some chain of other includes. However this breaks for me using external fmt and gcc-10. Anyway, the include should clearly be there, so I added it.

Cheers,
Jonas


gabime:
Thanks. Merged

PR #1752, opened 2020-11-30
Update null_mutex.h

ZgblKylin:
Using std::exchange for null_atomic_int::exchange


gabime:
Thanks. The problem is it is  c++14 and above, and spdlog is supporting c++11

PR #1768, opened 2020-12-19
Changed is_color_terminal to be more generic.

dominicpoeschko:
I have the issue that the in the alacritty terminal spdlog has no colors,
so I added it the the supported terminals.

To be more generic I added also a check for the COLORTERM variable
which should also indicate if the terminal supports color.


gabime:
Thanks @dominicpoeschko

PR #1771, opened 2020-12-26
ensure stdout color sinks do not leak windows headers with SPDLOG_COMPILED_LIB

### LINKED ISSUE #1770: wincolor_sink pulls in windows.h even with SPDLOG_COMPILED_LIB defined
I don't think it would be too hard to split the dependencies into the inline file. I could make a pr for it if you're okay with that.

Ryan-rsm-McKenzie:
fixes #1770


gabime:
This is excellent. Thanks @Ryan-rsm-McKenzie !

PR #1774, opened 2020-12-27
skip module mode when finding fmt

Ryan-rsm-McKenzie:
`fmt` provides a config file, so there's no need to waste time trying module mode. Also, it would appear the `CONFIG` option was previously passed, [until being removed in what appears to be an accident?](https://github.com/gabime/spdlog/commit/6fe899af104930c381b8bb0498e93b7a09418241#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL71) In any case, there's no explanation for it, so I assume it will be fine to add it back.


gabime:
Thanks @Ryan-rsm-McKenzie

Issue #1759, opened 2020-12-07
Please can basic example show switching to stderr


tonyelewis:
Thanks very much for this library and for having helpful examples in the docs.

I'm trying to figure out the simplest way to switch to `stderr` so that `spdlog::warn()` etc will go to `stderr`. It occurs to me that I think this would be a really useful thing to include in the [Basic-example](https://github.com/gabime/spdlog/wiki/1.-QuickStart#basic-example).

Also, please may I ask: what's the answer? I haven't been able to figure it out.

Thanks very much.

tonyelewis:
Is the following the simplest way to switch to `stderr`?

~~~cpp
spdlog::set_default_logger( spdlog::stderr_color_st( "some_arbitrary_name" ) );
~~~

tt4g:
> Is the following the simplest way to switch to stderr?

Right.

tony-triller:
> Right.

Thanks very much. I think it would be really useful for us to put something like this in the [Basic-example](https://github.com/gabime/spdlog/wiki/1.-QuickStart#basic-example).

But first, one problem with the above code is that it causes `[some_arbitrary_name]` to be inserted into the log messages:

```
[2020-12-16 09:34:50.033] [some_arbitrary_name] [warning] Message
```

&hellip;whereas the original unmodified log has no such field:

```
[2020-12-16 09:35:19.577] [warning] Message
```

`spdlog::stderr_color_st()` doesn't compile and `spdlog::stderr_color_st( "" )` causes a `spdlog::spdlog_ex` to be thrown with a `.what()` message of : `logger with name '' already exists`.

What's the simplest way to switch from `stdout` to `stderr` without changing anything else?

tony-triller:
~~~cpp
spdlog::set_default_logger( spdlog::stderr_color_st( "some_arbitrary_name" ) );
spdlog::set_default_logger( spdlog::stderr_color_st( "" ) );
~~~

&hellip;appears to work. Is there a better way?

tonyelewis:
<sup>(replacing comment mistakenly posted from different account - sorry for the muddle)</sup>

> Right.

Thanks very much. I think it would be really useful for us to put something like this in the [Basic-example](https://github.com/gabime/spdlog/wiki/1.-QuickStart#basic-example).

But first, one problem with the above code is that it causes `[some_arbitrary_name]` to be inserted into the log messages:

```
[2020-12-16 09:34:50.033] [some_arbitrary_name] [warning] Message
```

[...]

PR #1775, opened 2020-12-28
Add README note on switch logging to stderr

tonyelewis:
This is the sort of thing I'm thinking of re #1759. Does this look suitable to you?


gabime:
Thanks @tonyelewis . I think it might be better to put it instead in the wiki in the FAQ section.

PR #1787, opened 2021-01-06
Better support for / separators on Windows, improve wchar filename coverage

sylveon:
The first part of this PR fixes a bug with the default formatting. When not using [/FC](https://docs.microsoft.com/en-us/cpp/build/reference/fc-full-path-of-source-code-file-in-diagnostics) the `__FILE__` macro can take strange values such as `C:\Projects\TranslucentTB\TranslucentTB\tray\../dynamicloader.hpp`. Under the current code, spdlog would print it in the log as `../dynamicloader.hpp`. With this PR, it now correctly prints `dynamicloader.hpp`.

As a bonus, I added an extension point for users: they can change what spdlog considers to be a folder separator, if for example they want to revert to the old behavior.

I've also added an AppVeyor worker which tries building spdlog with wide filenames enabled, and adjusted the unit tests accordingly. If I didn't forget anything, I believe it should pass.


sylveon:
Crud a test broke - it's currently 4:30am here so I will look at it tomorrow.

sylveon:
All green now! I also fixed an issue where win_eventlog_sink would silently discard errors from `ReportEvent` or `MultiByteToWideChar` if WCHAR to UTF-8 was enabled. Now they are correctly handled.

Also fixed a security bug in `utf8_to_wstrbuf` that could result in a potential buffer overflow/stack overrun. The buffer was a narrow char buffer but the size was passed as if it was a wide char buffer, making it seem twice as big than it really is. I fixed the issue by using a proper wide character buffer.

example/example.cpp
```diff
@@ -109,21 +109,21 @@ void stdout_logger_example()
 void basic_example()
 {
     // Create basic file logger (not rotated).
-    auto my_logger = spdlog::basic_logger_mt("file_logger", "logs/basic-log.txt");
+    auto my_logger = spdlog::basic_logger_mt("file_logger", SPDLOG_FILENAME_T("logs/basic-log.txt"));
```
gabime, line 112: While correct, I think the example should be without this SPDLOG_FILENAME_T macro. It is only needed in very special cases (windows + wchar support).

example/example.cpp
```diff
@@ -109,21 +109,21 @@ void stdout_logger_example()
...
 }
 
 #include "spdlog/sinks/rotating_file_sink.h"
 void rotating_example()
 {
     // Create a file rotating logger with 5mb size max and 3 rotated files.
-    auto rotating_logger = spdlog::rotating_logger_mt("some_logger_name", "logs/rotating.txt", 1048576 * 5, 3);
+    auto rotating_logger = spdlog::rotating_logger_mt("some_logger_name", SPDLOG_FILENAME_T("logs/rotating.txt"), 1048576 * 5, 3);
```
gabime, line 119: While correct, I think the example should be without this SPDLOG_FILENAME_T macro. It is only needed in very special cases (windows + wchar support).

example/example.cpp
```diff
@@ -109,21 +109,21 @@ void stdout_logger_example()
...
 }
 
 #include "spdlog/sinks/daily_file_sink.h"
 void daily_example()
 {
     // Create a daily logger - a new file is created every day on 2:30am.
-    auto daily_logger = spdlog::daily_logger_mt("daily_logger", "logs/daily.txt", 2, 30);
+    auto daily_logger = spdlog::daily_logger_mt("daily_logger", SPDLOG_FILENAME_T("logs/daily.txt"), 2, 30);
```
gabime, line 126: While correct, I think the example should be without this SPDLOG_FILENAME_T macro. It is only needed in very special cases (windows + wchar support).

example/example.cpp
```diff
@@ -143,7 +143,7 @@ void async_example()
 {
     // Default thread pool settings can be modified *before* creating the async logger:
     // spdlog::init_thread_pool(32768, 1); // queue with max 32k items 1 backing thread.
-    auto async_file = spdlog::basic_logger_mt<spdlog::async_factory>("async_file_logger", "logs/async_log.txt");
+    auto async_file = spdlog::basic_logger_mt<spdlog::async_factory>("async_file_logger", SPDLOG_FILENAME_T("logs/async_log.txt"));
```
gabime, line 146: While correct, I think the example should be without this SPDLOG_FILENAME_T macro. It is only needed in very special cases (windows + wchar support).

example/example.cpp
```diff
@@ -211,7 +211,7 @@ void multi_sink_example()
     console_sink->set_level(spdlog::level::warn);
     console_sink->set_pattern("[multi_sink_example] [%^%l%$] %v");
 
-    auto file_sink = std::make_shared<spdlog::sinks::basic_file_sink_mt>("logs/multisink.txt", true);
+    auto file_sink = std::make_shared<spdlog::sinks::basic_file_sink_mt>(SPDLOG_FILENAME_T("logs/multisink.txt"), true);
```
gabime, line 214: While correct, I think the example should be without this SPDLOG_FILENAME_T macro. It is only needed in very special cases (windows + wchar support).

include/spdlog/details/file_helper-inl.h
```diff
@@ -118,7 +118,7 @@ SPDLOG_INLINE std::tuple<filename_t, filename_t> file_helper::split_by_extension
     }
 
     // treat cases like "/etc/rc.d/somelogfile or "/abc/.hiddenfile"
-    auto folder_index = fname.rfind(details::os::folder_sep);
+    auto folder_index = fname.find_last_of(details::os::folder_seps_filename);
```
gabime, line 121: Will  break if ```folder_seps_filename``` is more than single  char. Let's use the original ```rfind()```

include/spdlog/details/os-inl.h
```diff
@@ -484,15 +484,15 @@ SPDLOG_INLINE void utf8_to_wstrbuf(string_view_t str, memory_buf_t &target)
     }
 
     int result_size = static_cast<int>(target.capacity());
-    if ((str_size + 1) * 2 < result_size)
+    if (str_size + 1 > result_size)
```
gabime, line 487: Not sure what will happen in str_size is exatcly MAX_INT. 

To be on the safe side, , I think the 

```c++
if (str.size() > static_cast<size_t>((std::numeric_limits<int>::max)()))
    {
        throw_spdlog_ex("UTF-8 string is too big to be converted to UTF-16");
    }
```
should be changed to something like (don't allow sizes bigger than MAX_INT/2):


```c++
if (str.size() > static_cast<size_t>((std::numeric_limits<int>::max/2)()))
    {
        throw_spdlog_ex("UTF-8 string is too big to be converted to UTF-16");
    }
```

include/spdlog/details/os-inl.h
```diff
@@ -484,15 +484,15 @@ SPDLOG_INLINE void utf8_to_wstrbuf(string_view_t str, memory_buf_t &target)
...
     {
         result_size = ::MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, str.data(), str_size, NULL, 0);
     }
 
     if (result_size > 0)
     {
         target.resize(result_size);
-        result_size = ::MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, str.data(), str_size, (LPWSTR)target.data(), result_size);
+        result_size = ::MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, str.data(), str_size, target.data(), result_size);
```
gabime, line 496: Why remove the ```(LPWSTR)``` cast ? Does  ```target.data()``` guaranteed to be (LPWSTR) ?

include/spdlog/details/os-inl.h
```diff
@@ -533,15 +533,10 @@ SPDLOG_INLINE bool create_dir(filename_t path)
...
-#ifdef _WIN32
-    // support forward slash in windows
-    std::replace(path.begin(), path.end(), '/', folder_sep);
-#endif
-
     size_t search_offset = 0;
     do
     {
-        auto token_pos = path.find(folder_sep, search_offset);
+        auto token_pos = path.find_first_of(folder_seps_filename, search_offset);
```
gabime, line 539: Will break if ```folder_seps_filename ```is more than single char. Let's use the original find()

include/spdlog/details/os.h
```diff
@@ -32,11 +32,16 @@ SPDLOG_API std::tm gmtime() SPDLOG_NOEXCEPT;
 SPDLOG_CONSTEXPR static const char *default_eol = SPDLOG_EOL;
 
 // folder separator
+#if !defined(SPDLOG_FOLDER_SEPS)
 #ifdef _WIN32
-static const char folder_sep = '\\';
+#define SPDLOG_FOLDER_SEPS "\\/"
 #else
-SPDLOG_CONSTEXPR static const char folder_sep = '/';
+#define SPDLOG_FOLDER_SEPS "/"
```
gabime, line 39: Seems incorrect. What would happen if a folder name in windows contains the "\" character ? Let's keep the orig code

include/spdlog/pattern_formatter-inl.h
```diff
@@ -814,11 +816,31 @@ class short_filename_formatter final : public flag_formatter
...
+            const std::reverse_iterator<const char*> begin(filename + std::strlen(filename));
+            const std::reverse_iterator<const char*> end(filename);
+
+            const auto it = std::find_first_of(begin, end, std::begin(os::folder_seps), std::end(os::folder_seps) - 1);
+            return it != end ? it.base() : filename;
+        }
     }
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif // _MSC_VER
```
gabime, line 843: Please post a  dedicated pr for this optimization.

gabime (changes requested):
Thanks for this pr. Please see my comments

include/spdlog/pattern_formatter-inl.h
```diff
@@ -814,11 +816,31 @@ class short_filename_formatter final : public flag_formatter
...
+#pragma warning(push)
+#pragma warning(disable : 4127) // consider using 'if constexpr' instead
+#endif // _MSC_VER
     static const char *basename(const char *filename)
     {
-        const char *rv = std::strrchr(filename, os::folder_sep);
-        return rv != nullptr ? rv + 1 : filename;
+        // if the size is 2 (1 character + null terminator) we can use the more efficient strrchr
+        // the branch will be elided by optimizations
+        if (sizeof(os::folder_seps) == 2)
```
gabime, line 827: Please post a dedicated pr for this optimization.

sylveon:
> (e.g. what would happen if a folder path constains the '/' char as part of the path?)

Windows already disallows this and [`CreateFile`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea) (the Win32 API that is ultimately called) interprets / as a valid folder separator. This is easily confirmed by seeing that the tests pass correctly: some tests use / in file paths, and they do not fail
image

example/example.cpp
sylveon: The CI builds the example also, which means that the entry I added to the AppVeyor matrix fails without this.

include/spdlog/details/file_helper-inl.h
sylveon: That's the intention: each character in `folder_seps_filename` is considered as a possible folder separator. `find_last_of` searches for any character in `folder_seps_filename` instead of searching for the entirety of the string.

include/spdlog/details/os-inl.h
sylveon: Yes, because now it uses a `wmemory_buffer`, which as the w prefix says, is always wide.

include/spdlog/details/os.h
sylveon: Folders can't contain this character.

include/spdlog/pattern_formatter-inl.h
sylveon: This change doesn't make sense without the rest of the folder separator changes

include/spdlog/details/os-inl.h
sylveon: Why not add a -1, like this?
```cpp
if (str.size() > static_cast<size_t>(std::numeric_limits<int>::max() - 1))
```

gabime: agreed

example/example.cpp
gabime: Yes, but for such specific case that i must insist. Otherwise users might think this ugly macro is the standard usage

sylveon: Resolved

include/spdlog/details/os-inl.h
sylveon: Resolved

gabime:
Thanks @sylveon . Merged.

Issue #1793, opened 2021-01-09
truncate argument of basic_file_sink might still want "ab" mode


graydon:
This might or might-not be a bug depending on your inclinations, but I figured I'd open an issue to discuss.

In the existing spdlog code, basic_file_sink has a `truncate` argument that, if set, opens the file in `"wb"` mode rather than the default `"ab"` mode.

This does cause truncation! It works as far as that goes. But it also turns _off_ something that's arguably desirable: `"ab"` mode (i.e. the `O_APPEND` flag to `open(2)`) causes file-writes to always go to the current end-of-file, even when the file shrinks between writes. It effectively prefaces every `write(2)` call with an `lseek(2)` call moving the fd's offset to `SEEK_END`.

In `"wb"` mode, if the file shrinks between `write(2)` calls, the space between its end and the file position at the next `write(2)` will be zero-filled. Which is fairly unpleasant! And this happens in a specific, fairly common scenario: the <a href="https://linux.die.net/man/8/logrotate">`logrotate(8)`</a> command will `truncate(2)` the logfile after it makes a copy.

I know spdlog has its own, totally viable _internal_ log rotation system, but if users happen to run something like `logrotate(8)` on their own (say on a cron job) then spdlog will continue writing at the pre-truncation offset, zero-filling the prefix. Thus "rotating" a 10MB log file with `logrotate(8)` will produce a 10MB zero-fill prefix followed by the next log entry.

The solution to this is fairly easy: always open in `"ab"` mode, and treat the `truncate` argument as a request to additionally -- manually -- truncate before opening.

gabime:
Thanks for reporting. 
If this can be fixed easily and make logrotate happy, then I am all for it. PR is welcome.

gabime:
Fixed in pr #1795

PR #1795, opened 2021-01-09
Open files with "ab" mode even if truncating.

graydon:
This is a possible implementation of https://github.com/gabime/spdlog/issues/1793

It passes unit tests, I'm not sure what else needs checking.


gabime:
Thanks @graydon. Merged.

PR #1819, opened 2021-02-05
better numeric_limits<>::max/min fix when using SPDLOG_WCHAR_TO_UTF8_SUPPORT

NukeULater:
NOMINMAX macro doesn't work everytime for some reason


gabime:
What do you mean doesnt work?

NukeULater:
> What do you mean doesnt work?



> What do you mean doesnt work?

Header only compile was causing issues with that,      even with NOMINMAX defined

gabime:
so why does removing the define fix this?

NukeULater:
> so why does removing the define fix this?

It doesn't, the part where I added the parenthesis around std::numeric_limits::max does, and removed the define because it's unnecessary

gabime:
so lets put it back in, otherwise it might break backward compitability for code that relies on it to be already defined.

NukeULater:
if there is not any other code using min/max not sure what backward compatibility it could break, but i re-added it

gabime:
Thanks @NukeULater. Merged.

PR #1835, opened 2021-02-17
Set default value to registry::err_handler_

jneruda:


jneruda:
@gabime This is just a small improvement of the code, it was reported as unitialized variable by Klocwork and since that pointer is tested for null, it would be good to have it initialized.

gabime:
Merged. Thanks @jneruda

Issue #1748, opened 2020-11-25
Level 4 warning when building x86 MSVC


WattsUp:
Building spdlog v1.8.0 in x86 MSVC (19.28.29334.0) with `\Wx \W4` flags fails with:
**Warning C4702**
`details/os-inl.h` on [line 249](https://github.com/gabime/spdlog/blob/4a9ccf7e38e257feecce0c579a782741254eaeef/include/spdlog/details/os-inl.h#L249)
```C++
// Return file size according to open FILE* object
SPDLOG_INLINE size_t filesize(FILE *f)
{
    //...
    throw_spdlog_ex("Failed getting file size from fd", errno);
    return 0; // will not be reached.
}
```

johannes-schmitt:
I had the same issue.
After setting /GL compiler flag, the warning disappeared.

To set the flag in Visual Studio, go to project properties > General and set "Whole Program Optimization" to "Use Link Time Code Generation".

This is not fixing the root cause, though. But it may provide a solution to others.

PR #1837, opened 2021-02-19
remove unreachable code

johannes-schmitt:
Fix for #1748.

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -246,6 +246,13 @@ SPDLOG_INLINE size_t filesize(FILE *f)
 #endif
 #endif
     throw_spdlog_ex("Failed getting file size from fd", errno);
+
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4702) // disable unreachable code warning under msvc
+    return 0; // will not be reached.
+#pragma warning(pop) // disable unreachable code warning under msvc
+#endif // _MSC_VER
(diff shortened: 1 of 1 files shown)


gabime (commented):
I think older compiler would complain since throw_spdlog_ex is a function wrapper to throw, (and also this wrapper can be defined as abort())

johannes-schmitt:
I saw
```
#define SPDLOG_THROW(ex)                                                                                     \\
    do                                                                                                       \\
    {                                                                                                        \\
        printf("spdlog fatal error: %s\n", ex.what());                                                       \\
        std::abort();                                                                                        \\
    } while (0)
```
Would it help if the loop-condition would be changed to `while (1)`? I see that not all compilers support the `[[noreturn]]` attribute on `std::abort`. With this, the compiler would understand that any code afterwards is not reachable.

Another possibility would be to disable the compiler warning at the return statement.

I'm obviously missing the bigger picture here, so any help is appreciated.

gabime:
There is no bigger picture :) 
Since we don't call throw directly, compilers might get confused and complain that there is no return statement after the throw_spdlog_ex() call.

johannes-schmitt:
I this case, ignoring the warning for MSVC sounds like a reasonable solution to make compiler happy and avoid any breaking changes for older compilers. PR has been updated accordingly. What do you think?

gabime:
How about adding new boolean “success” variable, and throw at the end if false?

PR #1847, opened 2021-02-25
add daily sink filename format

fawdlstty:
example:

```cpp
std::shared_ptr<spdlog::logger> _logger = spdlog::daily_logger_format_mt ("module", "%Y%m%d.log");
_logger->info ("hahahaha");
```

then will create file `20210225.log`

### DIFF
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -7,6 +7,7 @@
 #include <spdlog/details/file_helper.h>
 #include <spdlog/details/null_mutex.h>
 #include <spdlog/fmt/fmt.h>
+#include <spdlog/fmt/chrono.h>
 #include <spdlog/sinks/base_sink.h>
 #include <spdlog/details/os.h>
 #include <spdlog/details/circular_q.h>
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/daily_file_sink.h
```diff
@@ -34,6 +39,98 @@ struct daily_filename_calculator
...
+                    dest += '%';
+                    break;
+                default:
+                    throw_spdlog_ex("daily_file_sink: Invalid format");
+                }
+            } else {
+                dest += filename [i];
+            }
+        }
+        return dest;
```
gabime, line 132: why not replace all this and  receive param of  fmt chrono format string? this would make the impl trivial (https://fmt.dev/dev/api.html#chrono-api)

include/spdlog/sinks/daily_file_sink.h
```diff
@@ -46,21 +143,23 @@ class daily_file_sink final : public base_sink<Mutex>
 {
 public:
     // create daily file sink which rotates on given time
-    daily_file_sink(filename_t base_filename, int rotation_hour, int rotation_minute, bool truncate = false, uint16_t max_files = 0)
+    daily_file_sink(filename_t base_filename, int rotation_hour, int rotation_minute, bool truncate = false, uint16_t max_files = 0, bool is_fmt = false)
```
gabime, line 146: No need another flag. this is a template class that get as a template param FileNameCalc, which can be set to fmt_filename_calc (from a new  factory function maybe to make it nicer to use)

gabime (changes requested):

gabime (approved):

gabime:
Thanks @fawdlstty . Merged.

PR #1851, opened 2021-02-27
Add build2 package support

Klaim:
WARNING: Please do not merge yet! See below.

We are currently in the process of packaging `spdlog` for `build2`, we have it working (even on unsupported Windows, see for example: https://ci.stage.build2.org/@a993b64e-8ba2-422e-97d7-250cdb5828e0?builds=&pv=&tc=*&cf=&mn=&tg=&rs=*)

This change simply adds the necessary information to use this package when using `build2`. Some notes:
 - not sure how to give the info succintly because `build2` allows packages to come from different sources, including the git repository of the package - in doubt I just linked to the future community repository address providing all the info, but the other package manager didn't do it like that so not sure if it's ok for you?; 
 - `build2` distinguish the package (`depends: spdlog <some-version-scheme>`, added in the `manifest` file of a `build2` project) and the target (`spdlog%lib{spdlog}` imported in a `buildfile`) because packages can contain several targets. I was not sure how to formulate that here, so feel free to tell me if I should just provide the name of the package and not more info?

The package isn't available yet so this have to be merged only once it is made available (probably in a few days).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ $ cmake .. && make -j
 * vcpkg: `vcpkg install spdlog`
 * conan: `spdlog/[>=1.4.1]`
 * conda: `conda install -c conda-forge spdlog`
+* build2: [package `spdlog`](https://cppget.org/spdlog), target `spdlog%lib{spdlog}`
 
 
 ## Features


gabime:
what is build2? never heard of it

Klaim:
It's a build-system combined with a package manager, see https://build2.org
A bit like if you used conan+make or conan+cmake/make or etc.

It's known to be one of the first buildsystem handling C++ modules (when preview implementations were made available), among other things. 

Once the package will be published, it will be available on https://cppget.org which is the OSS community central repository. But build2 also allows to use build2 packages/projects inside git repositories, so that's another option (when you want to use a tweaked version of the package for example).

Klaim:
The build2 package for spdlog v1.8.2 is currently in the publication queue: https://queue.cppget.org/spdlog/1.8.2+1
There you can also see the build output (only the tests are executed for testing).
There are a few warnings on Windows (which I understand is not officially supported, but it actually works as long as you link with the right system library) and we deactivated gcc/mingw on Windows because some tests fails (but build works).

Waiting for the publication to be done.

gabime:
> There are a few warnings on Windows (which I understand is not officially supported, but it actually works as long as you link with the right system library) 

What do you mean ? Windows  is officially supported

Klaim:
> > There are a few warnings on Windows (which I understand is not officially supported, but it actually works as long as you link with the right system library)
> 
> What do you mean ? Windows is officially supported

Sorry, I meant spdlog as **shared library** on Windows.
Or maybe[ this comment](https://github.com/gabime/spdlog/blob/v1.x/CMakeLists.txt#L119) is not up to date?

Klaim:
`spdlog` package `v1.8.2` is now online: https://cppget.org/spdlog

You can merge this now if you want, or point me tweaks you would like me to do.

gabime:
> Or maybe this comment is not up to date?

It is not updated. Shared build is supported.

gabime:
@Klaim Something doesn't look right in the command example, so I reverted.
What is the exact command to use?

Klaim:
@gabime There is no specific "command": you just add the package as dependency in your `build2` project's `manifest` file like this `depends: spdlog ^1.8.2` (for any version compatible with 1.8.2 for example) and the buildsystem will invoke the package manager automatically at the next command to retrieve and initialize that package, or will do so when you initialize the project in a new configuration (so you have as many libraries built as configurations - note that build2 only have source packages).

After that you can then import the `spdlog%lib{spdlog}` target to make your own targets depends on it (in `buildfile`s where your targets are defined).
This is assuming your project relies on cppget.org to get it's packages, which is can be done manually or by adding it in another project's file. Most `build2` projects rely on that repository.

It's similar to `conan` in the sense that there is a file in your project listing the packages that project depends on, and then the targets in it are made available to your build scripts (CMake scripts for example).

So here I just named the package and the target to import, similarly to the `conan` line. I'm assuming the user knows `build2` so they can easilly use that information, as for the other package managers. I tried to be short because the other ones are short, and the conan one just name the package and probably assume the user knows what to do with it.

Or do I misunderstand what you want in there?

gabime:
so, mayne something like 
```build2: depends: spdlog ^1.8.2``` ?

Klaim:
Works for me.
It might be obscure for newcomers (where do they write add?) but it's obvious to people used to `build2`

gabime:
btw I am a big fan of cargo, so I really hope build2 succeeds!

Klaim:
Nice :D

I updated the branch, should I make another PR?

gabime:
No need. I already updated

Klaim:
Thanks!

Issue #1850, opened 2021-02-26
Daily file sink does not clean old files of previous program executions


rubenpraets:
As far as I can see, the daily file sink will only work as intended under some specific circumstances. I'll try to explain what I think is wrong with it, but I don't have a lot of time now so feel free to ask for more info. 

The problem seems to be that this logger doesn't take into account that the process using spdlog is not guaranteed to be running continuously. Some problems this entails:
- Deleting old files only ever happens if the process is active when the rotation time point comes around. If not, spdlog will never detect this and old files will pile up.
- If the process is stopped and then started the next day, a new log file is created no matter the intended rotation time. When the rotation time point rolls around, it is assumed that a new file will be created and an old one is removed. Instead, the current file name is entered a second time into the filenames_q_ and not the oldest but second oldest log file is deleted, leaving the oldest one there for eternity.

My description is probably not complete, but if you reason about it a bit you might see what I mean, or otherwise prove me wrong. In any case, I also have a slightly simplified version of your daily file sink that I'm willing to share. It always rotates at midnight, and that deals with some of the problems I describe above.
Thanks anyway for the great logging library you gave us ;)

gabime:
Thanks for reporting. Both of the problems are due to the fact that spdlog, for performance reasons avoids to accessing the disk to search existing files,  but uses in memory q jnstead.

It might be a good idea to replace with something more robust even at the expense of accessing the disk. PRs are welcome.

rubenpraets:
I'll make a PR for my adaptations when I'm sure they are correct, but in the meantime I can also recommend you to look at this Python class: [TimedRotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.TimedRotatingFileHandler) (source code available [here](https://github.com/python/cpython/blob/3.9/Lib/logging/handlers.py))
It offers the same basic functionality as your daily file sink but also some extra stuff, and might be a good source of inspiration.

PR #1858, opened 2021-03-04
Add simpler version of daily file sink

rubenpraets:
This PR is in response to issue #1850 and addresses some of the problems with the daily file sink. I have added a new sink instead of changing the existing one for the following reasons:

- My proposed solution does not offer the functionality of rotating on a specific time, but is fixed to rotate at midnight
- It still does not address all problems. In particular, if a program using spdlog is not run at all in an entire calendar day, log files from before that day will not count towards the max_files count and will not be deleted automatically (as is also the case for the current daily file sink).

This implementation has some important advantages though. The functionality of rotating on a specific time is not implemented in a robust way in the current daily file sink. This can cause spdlog to throw an exception and to fail to remove an old log file, while it is also possible that spdlog rotates a log file too early. 
Another contribution is that old log files are now also deleted if the program was not running when the rotation time comes around.

Like @gabime said, that last problem with removing old log files could be solved by scanning the whole directory for files that match the format, but that is something I leave to be done. So, do with this PR what you will, complete it, use my code to replace the current daily file sink or ignore it.

### DIFF
--- a/include/spdlog/sinks/daily_file_sink_simple.h
+++ b/include/spdlog/sinks/daily_file_sink_simple.h
@@ -1,0 +1,233 @@
+// Copyright(c) 2015-present, Gabi Melman & spdlog contributors.
+// Distributed under the MIT License (http://opensource.org/licenses/MIT)
+
+#pragma once
+
+#include <spdlog/common.h>
+#include <spdlog/details/file_helper.h>
+#include <spdlog/details/null_mutex.h>
+#include <spdlog/fmt/fmt.h>
+#include <spdlog/fmt/chrono.h>
+#include <spdlog/sinks/base_sink.h>
+#include <spdlog/details/os.h>
+#include <spdlog/details/circular_q.h>
+#include <spdlog/details/synchronous_factory.h>
+
+#include <chrono>
+#include <cstdio>
+#include <ctime>
+#include <mutex>
+#include <string>
+
+// TODO Remaining problem: does not remove old files from before an interruption in execution that includes at least an entire calendar day (e.g. a weekend).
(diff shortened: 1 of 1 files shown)


gabime:
Thanks for the pr. 
> The functionality of rotating on a specific time is not implemented in a robust way in the current daily file sink. This can cause spdlog to throw an exception and to fail to remove an old log file, while it is also possible that spdlog rotates a log file too early.

How this be reproduced? It was never reported before. How does this PR fix it?

rubenpraets:
Well I hadn't actually tested this because it would take quite some time and it seems I was mistaken. I misinterpreted your code in that it would throw an exception if a file in `filenames_q_` did not exist on disk, but after a second inspection this does not seem to be the case. I'll give you my reasoning below, but you can basically ignore it.

Let's assume you set your rotation time at noon and start running your program somewhere in the morning. Spdlog does not check the rotation time, opening a new file which is added to the in-memory `filenames_q_` in `init_filenames_q_()`. The program runs some time and noon comes around. Spdlog sees that it should rotate, and thinks it opens a new file while in reality it opens the same file it created earlier. The filename is put in the `filenames_q_` anyway from `delete_old_()`, so now it contains the same filename twice. Now you let the program run uninterrupted and eventually spdlog will try to remove a file that is no longer there. That's where I thought an exception would be thrown, which is not correct.

So anyway, this PR makes the above scenario impossible, which is somewhat less of an improvement if you consider that throwing an exception is not what would have happened anyway. It does reduce the trouble with leftover log files, but if you don't think that's worth the bother then I won't blame you.

PR #1860, opened 2021-03-04
CMake: Add SPDLOG_STATIC_VCRT to choose static MSVC runtime

rofferom:
This PR simply adds a flag for Windows build to change the target runtime (`/MT`/`/MD`).

The naming comes from SDL2's cmake: https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L226.

I based my tests with this doc: https://stackoverflow.com/a/19487637.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@
 
 cmake_minimum_required(VERSION 3.10)
 
+cmake_policy(SET CMP0091 NEW)
+
 # ---------------------------------------------------------------------------------------
 # Start spdlog project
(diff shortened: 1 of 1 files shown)


gabime (approved):

gabime:
Thanks! Merged

gabime:
@rofferom Needed to revert. Apparently it breaks builds  with CMake versions prior to  3.15

rofferom:
Woops, I'll check how to update the patch. What is the minimum cmake version you target ?

gabime:
3.10

PR #1882, opened 2021-03-19
add default cases

imsherlock:
Adding default case for cases where the compilation flag
-Wswitch-default is present on the command line when spdlog is
included in external projects.

Signed-off-by: Ryan Sherlock <ryan.m.sherlock@gmail.com>

### DIFF
--- a/include/spdlog/fmt/bundled/format.h
+++ b/include/spdlog/fmt/bundled/format.h
@@ -2814,6 +2814,8 @@ FMT_CONSTEXPR const Char* parse_align(const Char* begin, const Char* end,
     case '^':
       align = align::center;
       break;
+    default:
+      break;
     }
     if (align != align::none) {
       if (p != begin) {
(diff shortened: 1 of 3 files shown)


include/spdlog/fmt/bundled/format.h
```diff
@@ -2814,6 +2814,8 @@ FMT_CONSTEXPR const Char* parse_align(const Char* begin, const Char* end,
     case '^':
       align = align::center;
       break;
+    default:
+      break;
```
gabime, line 2818: this is bundled copy from the fmt project, so should be fixed there.

include/spdlog/fmt/bundled/format.h
```diff
@@ -2897,6 +2899,8 @@ FMT_CONSTEXPR const Char* parse_format_specs(const Char* begin, const Char* end,
     handler.on_space();
     ++begin;
     break;
+  default:
+    break;
```
gabime, line 2903: this is bundled copy from the fmt project, so should be fixed there

gabime (changes requested):

include/spdlog/fmt/bundled/format.h
imsherlock: I'll fix it in the upstream fmtlib.

PR #1883, opened 2021-03-19
Add option to select C++20 std format

Oipo:
A minor change that would help reduce the amount of dependencies in my project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,10 +82,11 @@ option(SPDLOG_BUILD_WARNINGS "Enable compiler warnings" OFF)
 option(SPDLOG_INSTALL "Generate the install target" ${SPDLOG_MASTER_PROJECT})
 option(SPDLOG_FMT_EXTERNAL "Use external fmt library instead of bundled" OFF)
 option(SPDLOG_FMT_EXTERNAL_HO "Use external fmt header-only library instead of bundled" OFF)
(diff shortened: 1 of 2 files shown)


gabime:
How does it work? spdlog use fmt:: namespace to do formatting(https://github.com/gabime/spdlog/blob/44e1f9f6829cd8acbfea282d390560452e889aff/include/spdlog/logger.h#L340)

Oipo:
Well, this is embarrassing. Just because the tests compile and run doesn't mean the code path I thought actually got taken...

Moreover, std::format is not even provided by [any of the major compilers](https://en.cppreference.com/w/cpp/compiler_support) (P0645R10).  

Once they do and I have drunk some coffee that day, I'll make a new attempt.

Oipo:
Regarding your question though, perhaps it would be an idea to stick `fmt` in a define and swap it out with `std` when the CMake define is set.

gabime:
No worries, thanks for the attempt . 
The thing is that spdlog use internally some fmt utils that are not part of the standard, so it is not very easy to replace entirely fmt with std (and why would we? fmt is great, and it will take a while until std::format is properly supported by major compilers).

Oipo:
fmt is a great library, don't get me wrong. My day job is working in an enterprise which prefers to keep the amount of dependencies as low as possible. Everything open source we add has to be meticulously added to a database per version, hence the effort.

But if it's (way) more effort to modify spdlog, I'll gladly add fmt to the database as well.

Issue #1823, opened 2021-02-08
use std::source_location and replace macros with constexpr functions


L0ric0:
c++ 20 added [std::source_location](https://en.cppreference.com/w/cpp/utility/source_location) which provides the current file, line, collumn and function name. as I understand the SPDLOG_info ... macros are only needed to get this informations, so it would be nice to replace the macros with constexpr functions for c++ 20 code (std::source_location is currently supported on gcc trunk)

gabime:
How could you get the call site location without macros?

franklange:
Experimental PR was closed and revealed the following issues:
1) Default use of std::source_location means adding it as a default parameter to the log function. If the log function is a variadic template of the form `log(fmtstr, args...)` adding a default parameter is non-trivial.
2) Replacing the macros with constexpr functions makes no sense because ultimately they will call logger->log(), ie a non-constepxr function. That only leaves using `if constexpr ()` within the function but that still would mean that function arguments would be evaluated which they are not in case of disabled macros.
3) It's unclear whether it makes sense to keep the macros and rather add std::source_location to the existing non-macro log functions. That would mean always logging source_location info, whereas currently the caller can choose whether they want location information or not by using either the functions or the macros.

So overall I wouldn't know how to add std::source_location to this project without disrupting the existing design.

sylveon:
I think we should at least replace spdlog::source_loc by std::source_location when it's available, to make consumption from C++20 nicer.

oct15demo:
#define fl_debug(...) log(spdlog::source_loc{__FILE__, __LINE__, static_cast<const char *>(__FUNCTION__)}, spdlog::level::trace, __VA_ARGS__)

Just using the above gives you the option of not using SPDLOG_DEBUG to see filename:line# without changing anything in the current code base. You get individual loggers, not just the default, with the file number, line number, handy in development when you want console output to link to the source code. 

I went a bit further and changed the existing level function names in spdlog.h and logger.h for my own use. Plus some dereferencing the pointer to give my an easy syntax logger.debug("my message");  That part could probably be made configurable, and is just syntactic sugar.

But getting back to the marco above, perhaps you could consider offering that in an example, or footnote to the formatting table for %s and %#.

USAFrenzy:
[...]

PR #1884, opened 2021-03-19
WIP: replace SPDLOG_* macros with constexpr funcs + source_location

franklange:
I wanted to investigate what it would take to replace the existing log macros (eg. SPDLOG_TRACE) with constexpr functions that make use of the new `std::source_location` [1] that will be available in GCC11 [2], as originally discussed here:
https://github.com/gabime/spdlog/issues/1823

The idea of `std::source_location` is that you can provide it as a default-value parameter for your log function and it will still log all the info from the call-site.
However, combining default-value params with variadic template functions is almost impossible or at least quite involved. I opted for using `user-defined deduction guides` [3] as advocated on SO [4].

This is more an experiment but please let me know what you think :)
Would this be a direction worth pursuing?

________________________________________________
[1] https://en.cppreference.com/w/cpp/utility/source_location
[2] https://en.cppreference.com/w/cpp/compiler_support
[3] https://en.cppreference.com/w/cpp/language/class_template_argument_deduction#User-defined_deduction_guides
[4] https://stackoverflow.com/questions/57547273/how-to-use-source-location-in-a-variadic-template-function

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ endif ()
 # Compiler config
 # ---------------------------------------------------------------------------------------
 if (NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD 17)
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
 endif()
 
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -26,6 +26,18 @@ void custom_flags_example();
 
 int main(int, char *[])
 {
+    spdlog::set_level(spdlog::level::debug);
+
+    spdlog_trace(">>>TRACE");
+    spdlog_logger_trace(spdlog::default_logger_raw(), ">>>TRACE");
+
+    spdlog_debug(">>>DEBUG");
+    spdlog_logger_debug(spdlog::default_logger_raw(), ">>>DEBUG");
+
+    spdlog_critical(">>>CRITICAL");
+    spdlog_logger_critical(spdlog::default_logger_raw(), ">>>CRITICAL");
+
+
     // Log levels can be loaded from argv/env using "SPDLOG_LEVEL"
     load_levels_example();
 
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -10,6 +10,7 @@
 #include <chrono>
 #include <initializer_list>
 #include <exception>
+#include <experimental/source_location>
 #include <string>
 #include <type_traits>
 #include <functional>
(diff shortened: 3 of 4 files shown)


example/example.cpp
```diff
@@ -26,6 +26,18 @@ void custom_flags_example();
...
+
+    spdlog_trace(">>>TRACE");
+    spdlog_logger_trace(spdlog::default_logger_raw(), ">>>TRACE");
+
+    spdlog_debug(">>>DEBUG");
+    spdlog_logger_debug(spdlog::default_logger_raw(), ">>>DEBUG");
+
+    spdlog_critical(">>>CRITICAL");
+    spdlog_logger_critical(spdlog::default_logger_raw(), ">>>CRITICAL");
+
```
franklange, line 39: This compiled and worked for me on Linux + GCC 10.2.0.

CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ endif ()
 # Compiler config
 # ---------------------------------------------------------------------------------------
 if (NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD 17)
```
franklange, line 29: Apparently needed for deduction guides.

gabime:
Interesting approach. Thanks, for sharing. I wonder if the struct trick is the only way to achieve this. Ideally, this  would be added directly to the existing functions.

franklange:
> Interesting approach. Thanks, for sharing. I wonder if the struct trick is the only way to achieve this. Ideally, this would be added directly to the existing functions.

Heyhey, thanks for the quick reply!
Unfortunately I couldn't find any other way of combining a variadic template with a default-value parameter. So I guess this might be the way to go for all the fmt-style functions.

Regarding bigger picture: so you envision using source_location for all the log functions? eg:
```c++
// spdlog.h

template<typename T>
inline void trace(const T &msg, std::source_location loc = std::source_location::current())
{
    default_logger_raw()->trace(loc, msg); // needs mods in logger.h to take + forward source_location
}

```
I wasn't quite sure if it's desirable to "force" logging the source location in every log function. If I understand the API correctly then currently having these additional macros allows me to chose wether I'm interested in src_location or not.

So ya, let me know what you think. I can also add this to the existing functions but then I'd have to adjust `logger.h` as well.

gabime:
> So ya, let me know what you think. I can also add this to the existing functions but then I'd have to adjust logger.h as well.

If the only benefit is that one can call ```spdlog_trace``` instead of ```SPDLOG_TRACE```,  then I don't think it worth the extra complexity (not to mention it only supported in  c++20). 

Also, the macros are zero code if the level is disabled at compile time, so they can't be completely replaced with the proposed.

franklange:
> Also, the macros are zero code if the level is disabled at compile time, so they can't be completely replaced with the proposed.
Shouldn't the constexpr if inside the functions here make sure that they're empty and optimized away if the corresponding level is disabled?

I don't mind throwing this away. I started this purely based on your "thumbs up" on the original idea posted here: https://github.com/gabime/spdlog/issues/1823

What did you have in mind there?

gabime:
> Shouldn't the if constexpr inside the functions here make sure that they're empty and optimized away if the corresponding level is disabled?

Its not exactly the same. For example the args to the function would still be evaluated (`spdlog_debug(“{}”, some_expensive_func());`


> What did you have in mind there?

That should be considered,  but like I said, seems not worth it in the current state of affairs.

PR #1886, opened 2021-03-22
Add required libraries for Android CMake build

ahmedyarub:
This is to avoid
`ld: error: undefined symbol: __android_log_write`


gabime:
Thanks @ahmedyarub

PR #1888, opened 2021-03-24
added spdlog::level::set_string_view to enable alternate log level na…

stevenlunt:
…mes without changing the build via SPDLOG_LEVEL_NAMES.  presently we need to clone the repo and create a custom conan package for spdlog just to control the log level strings that appear in the log

### DIFF
--- a/include/spdlog/common-inl.h
+++ b/include/spdlog/common-inl.h
@@ -22,6 +22,11 @@ SPDLOG_INLINE const string_view_t &to_string_view(spdlog::level::level_enum l) S
     return level_string_views[l];
 }
 
+SPDLOG_INLINE void set_string_view(spdlog::level::level_enum l, const string_view_t &s) SPDLOG_NOEXCEPT
+{
+    level_string_views[l] = s;
+}
+
 SPDLOG_INLINE const char *to_short_c_str(spdlog::level::level_enum l) SPDLOG_NOEXCEPT
 {
     return short_level_names[l];
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -167,6 +167,7 @@ enum level_enum
 #endif
 
 SPDLOG_API const string_view_t &to_string_view(spdlog::level::level_enum l) SPDLOG_NOEXCEPT;
+SPDLOG_API void set_string_view(spdlog::level::level_enum l, const string_view_t &s) SPDLOG_NOEXCEPT;
 SPDLOG_API const char *to_short_c_str(spdlog::level::level_enum l) SPDLOG_NOEXCEPT;
 SPDLOG_API spdlog::level::level_enum from_str(const std::string &name) SPDLOG_NOEXCEPT;
 
--- a/tests/test_misc.cpp
+++ b/tests/test_misc.cpp
@@ -55,6 +55,14 @@ TEST_CASE("level_to_string_view", "[convert_to_string_view")
     REQUIRE(spdlog::level::to_string_view(spdlog::level::off) == "off");
 }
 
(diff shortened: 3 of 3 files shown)


gabime:
Thanks @stevenlunt

include/spdlog/common-inl.h
```diff
@@ -22,6 +22,11 @@ SPDLOG_INLINE const string_view_t &to_string_view(spdlog::level::level_enum l) S
     return level_string_views[l];
 }
 
+SPDLOG_INLINE void set_string_view(spdlog::level::level_enum l, const string_view_t &s) SPDLOG_NOEXCEPT
+{
+    level_string_views[l] = s;
```
sjanel, line 27: Hello,

I just have this compile error within my project (which used v1.8.2 previously without error):

```
FAILED: _deps/spdlog-build/CMakeFiles/spdlog.dir/src/spdlog.cpp.o 
/usr/local/bin/ccache /usr/bin/c++ -DSPDLOG_COMPILED_LIB -I_deps/spdlog-src/include -g -std=c++2a -MD -MT _deps/spdlog-build/CMakeFiles/spdlog.dir/src/spdlog.cpp.o -MF _deps/spdlog-build/CMakeFiles/spdlog.dir/src/spdlog.cpp.o.d -o _deps/spdlog-build/CMakeFiles/spdlog.dir/src/spdlog.cpp.o -c _deps/spdlog-src/src/spdlog.cpp
In file included from _deps/spdlog-src/src/spdlog.cpp:9:
_deps/spdlog-src/include/spdlog/common-inl.h: In function ‘void spdlog::level::set_string_view(spdlog::level::level_enum, const string_view_t&)’:
_deps/spdlog-src/include/spdlog/common-inl.h:27:29: error: passing ‘const string_view_t’ {aka ‘const fmt::v7::basic_string_view<char>’} as ‘this’ argument discards qualifiers [-fpermissive]
   27 |     level_string_views[l] = s;
      |                             ^
In file included from _deps/spdlog-src/include/spdlog/fmt/fmt.h:22,
                 from _deps/spdlog-src/include/spdlog/common.h:36,
                 from _deps/spdlog-src/include/spdlog/spdlog.h:12,
                 from _deps/spdlog-src/include/spdlog/spdlog-inl.h:7,
                 from _deps/spdlog-src/src/spdlog.cpp:8:
_deps/spdlog-src/include/spdlog/fmt/bundled/core.h:355:32: note:   in call to ‘constexpr fmt::v7::basic_string_view<char>& fmt::v7::basic_string_view<char>::operator=(const fmt::v7::basic_string_view<char>&)’
  355 | template <typename Char> class basic_string_view {
      |                                ^~~~~~~~~~~~~~~~~
```
Looks like it is linked to this change, could you please take a look ?

sjanel: In C++17, there is a `constexpr` qualifier on `level_string_views` so it makes assignment impossible, hence compilation error.

stevenlunt: I will remove constexpr as it does not seem to add any significant value in this case.  Sorry i did not notice this before.

sjanel: No problem, and thanks for this lib, I find it very useful!

gv-me: @stevenlunt can you check my comment [here](https://github.com/gabime/spdlog/pull/1889#issuecomment-834200256)?
the `constexpr` actually is very important to ensure the log levels are initialized properly, and to [prevent static initialization order fiasco](https://en.cppreference.com/w/cpp/language/siof) .
Removal of `constexpr` breaks a lot of things for us when use spdlog to log things happening before entering `main()`.

Issue #1791, opened 2021-01-08
static initialization order for level_string_views


gv-me:
In file common-inl.h level names are defined as `string_view`s. The declarations can be made `constexpr` so they are always initialized at start of the program.

Currently, any logs created before entering `main()` or even any operations dealing with level names like loading levels from env variables cannot be performed reliably before entering `main()` function of the program. Depending on the static initialization order, sometimes level names might not be present at the time.

For `constexpr` declaration of `level_string_views` only minor changes are needed. I can submit a PR if that is preferred.

gabime:
Sounds good. Thanks. 

Keep in mind this should still compile in c++11

gcerretani:
Are there reliable solutions for C++14? I'm developing a shared library using spdlog, and I would like to reliably invoke `spdlog::cfg::load_env_levels()` from a function marked as `__attribute__((constructor))`. The constexpr solution has no impact on C++14, right?

gcerretani:
Seems that the problem is solved also with C++14 using the x1.x branch since commit 68aed6a5ebf55aecc67351513ad4aea3e38dc7d9, but I can't find any workdaround for 1.8.5.

PR #1889, opened 2021-03-24
remove constexpr on level_string_views to fix compilation on C++17 fr…

stevenlunt:
…om addition of set_string_view

### DIFF
--- a/include/spdlog/common-inl.h
+++ b/include/spdlog/common-inl.h
@@ -10,9 +10,6 @@
 namespace spdlog {
 namespace level {
 
-#if __cplusplus >= 201703L
-constexpr
-#endif
 static string_view_t level_string_views[] SPDLOG_LEVEL_NAMES;
 
 static const char *short_level_names[] SPDLOG_SHORT_LEVEL_NAMES;


gabime (approved):

sjanel:
@gabime  Please merge this, spdlog is not usable in C++17 without it ;)

gabime (approved):

gabime:
will release new version

gv-me:
Hi, this removes the constexpr declaration for log levels, which re-adds the problems caused in issue #1791.

Static initialization will now cause problems again, now that the level names are not initialized reliably before entering `main()`

I cannot see the utility in saving a bit of compilation time while changing the log levels (hich was the feature that caused the removal of `constexpr` here) while sacrificing the runtime stability and initialization order.

@gabime even then if people want this, a compile option be added to choose between either constexpr log levels vs the ability to change them at runtime?

gabime:
> @gabime even then if people want this, a compile option be added to choose between either constexpr log levels vs the ability to change them at runtime?

Sounds right. But what about pre cpp 17 ?

gv-me:
for pre cpp 17, we can't declare `string_view` as `constexpr` anyway, due to the constructor that's being used, which is does not have `constexpr` until cpp 17.

gabime:
This means this option should be enabled only if cpp17 or newer is detected. PR is welcome.

gv-me:
yes, that's correct. I will take a look how it can be done, maybe in cmake.

stevenlunt:
I am fine with a compile-time option for c++17 and beyond.  it is still better than it was before, where I had to patch the spdlog package in order to change the strings.

gv-me:
@stevenlunt can you elaborate on what do you mean by "patching the package"?

If you use spdlog as header only lib, you only have define `SPDLOG_LEVEL_NAMES` and `SPDLOG_SHORT_LEVEL_NAMES` before you include spdlog.
If you compile spdlog from source, you only have to add the definition at compile time.
If you include spdlog in your project using cmake as `add_subdirectory`, you can define it in cmake as `target_compile_definitions`.

I cannot see the real use for setting the level names in a function in c++ code, when you should not be changing them often anyway.

stevenlunt:
i've tried updating the recipe to allow you to pass an option with the level names but it is very difficult if not impossible to convince CMake to define something like { "trace", "dbg", ...

It would be better if rather than one preprocessor token there were instead separate tokens for each level string.  this way it would be much more straightforward to override one or more of them

gv-me:
I just tested once with cmake command line and this works fine for me

```bash
cmake . -DSPDLOG_LEVEL_NAMES='{'\"mytrace\",\"mydebug\",\"myinfo\",\"mywarn\",\"mycritical\",\"myerror\",\"myoff\"'}'
```

in my CMakeLists I added
```cmake
add_subdirectory(./path/to/spdlog spdlog)
if(SPDLOG_LEVEL_NAMES)
    target_compile_definitions(spdlog PUBLIC SPDLOG_LEVEL_NAMES=${SPDLOG_LEVEL_NAMES})
endif()
```
This lines after `add_subdirectory` can actually be added to spdlog CMakeLists.txt as well to directly make it work with no changes in user projects.

But you are right that if there was a token for each level it would look better.

stevenlunt:
i was trying to pass options to conan create, something like:

conan create . 1.8.5@third-party/stable -o spdlog:level_name_debug=DBG -o spdlog:level_name_info=INF -o spdlog:level_name_warning=WRN -o spdlog:level_name_error=ERR

gv-me:
I have not used conan options before, would these then be passed through as CMake variables from conan?

stevenlunt:
it does not seem possible through conan options to pass a macro through to gcc like {"string","string",...}

gv-me:
would something like this help perhaps? 
https://github.com/conan-io/conan/issues/4754#issuecomment-474285354

stevenlunt:
i will create a new pull request to revert my changes in from this PR and add separate string #defines for the individual level names

stevenlunt:
https://github.com/gabime/spdlog/pull/1948

PR #1890, opened 2021-03-25
Remove version requirement from fmt find_package

prince-chrismc:
By introducing 'no module' support https://github.com/gabime/spdlog/commit/3d8f71c4d23fd2df08e4f6284155958b0df47b33 this inadvertently triggered the find_package to use the "full version" version selection.

From https://cmake.org/cmake/help/latest/command/find_package.html#version-selection:

> When the [version] argument is given, Config mode will only find a version of the package that claims compatibility with the requested version (see format specification).

FMT does not set this configuration https://github.com/fmtlib/fmt/blob/7.1.3/support/cmake/fmt-config.cmake.in... but regardless it would not be "any compatibility" based on version semantics.

Which causes this error. v1.8.2 builds perfectly fine.

```
CMake Error at CMakeLists.txt:181 (find_package):
  Could not find a configuration file for package "fmt" that is compatible
  with requested version "5.3.0".

    /home/proj/build/e98598522b7c484d9220bffc7ec84474515facbc/fmt-config.cmake, version: 7.1.3
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,7 +178,7 @@ target_link_libraries(spdlog_header_only INTERFACE Threads::Threads)
 # ---------------------------------------------------------------------------------------
 if(SPDLOG_FMT_EXTERNAL OR SPDLOG_FMT_EXTERNAL_HO)
     if(NOT TARGET fmt::fmt)
-        find_package(fmt 5.3.0 CONFIG REQUIRED)
+        find_package(fmt CONFIG REQUIRED)
     endif()
     target_compile_definitions(spdlog PUBLIC SPDLOG_FMT_EXTERNAL)
     target_compile_definitions(spdlog_header_only INTERFACE SPDLOG_FMT_EXTERNAL)


gabime:
so how can spdlog state the minimum fmt version to use?

prince-chrismc:
Hmm, good question 

- Remove the `CONFIG` for sure
- check the package version string of it's found (less modern CMake)

gabime:
Thanks.  Merged

Ryan-rsm-McKenzie:
fmt does set version compatibility. See this line [here](https://github.com/fmtlib/fmt/blob/9cb347b4b2e80fc9fbf57b8621746663c3f870f6/CMakeLists.txt#L285). Did you check that `fmt-config-version.cmake` is installed correctly?

whyman:
What was the solution to the version check? What are acceptable to fmt versions?

prince-chrismc:
To the best of my knowledge... If you want the latest compatible `{fmt}` for spdlog

```py
        if version >= "1.9.0":
            requires("fmt/8.0.1")
        elif version >= "1.7.0":
            requires("fmt/7.1.3")
        elif version >= "1.5.0":
            requires("fmt/6.2.1")
        else: # Going back to 1.4.2
            requires("fmt/6.0.0")
```

gabime:
only if using external fmt instead of the one bundled with spdlog.

whyman:
Sure, this is with external fmt. 

Would it not be nicer for everyone if that logic was in this projects CMake? Instead of everyone having to have some similar logic? Or as it is now and the build just fails with incompatible versions.

gabime:
well, the point of this pr was to remove this logic because it wasn’t reliable. PR to add good support for this is welcome.

Issue #1790, opened 2021-01-08
Cmake build failed with v140_xp, v141_xp toolset flags


wwriter:
First of all, thank you for providing and maintaining this nice library.
I am posting this to ask if there's some way to build spdlog with `_xp` toolset flag.

When I use this script to build spdlog
``` sh
cmake -T v141_xp (or v140_xp) ... (other flags) ..
cmake --build . --config Debug (or Release)
```

I got this error message.
```
stdout_sink-inl.h (18) fileapi.h : no such file or directory
```
So I opened the inl file and found these lines.
``` cpp
#ifdef _WIN32
// under windows using fwrite to non-binary stream results in \r\r\n (see issue #1675)
// so instead we use ::FileWrite
#include <spdlog/details/windows_include.h>
#include <fileapi.h> // WriteFile (..) //< ******this seemed to cause the error*******
#include <io.h>      // _get_osfhandle(..)
#include <stdio.h>   // _fileno(..)
#endif               // WIN32
```
After doing some tests, I figured out that one cannot include `fileapi.h` when the version of Windows SDK for the project is set to 7.
Then I changed the code somewhat like this and was able to get the lib file.
``` cpp
#ifdef _WIN32
// ... 
#ifndef _USING_V110_SDK71_
#include <fileapi.h> // WriteFile (..) //< ******this seemed to cause the error*******
#endif
// ...
#endif               // WIN32
```

My question is, is there any flag or option that I can set to build a lib for Windows xp?

gabime:
No, WinXP is not supported.

wwriter:
I thought it might support windows xp because earlier today, I found some questions about crashes on windows xp or whatnot.
Anyway, thanks for your quick response !

swex:
@wwriter If you just remove `#include fileapi.h` it would compile and work just ok.
btw @gabime why windows xp is not supported anyway?

gabime:
It didn't compile cleanly and I have no time to support it.
Prs are welcome to make it work.

PR #1906, opened 2021-04-05
fix bug #1790

LonghronShen:
I'm also stuck at the problem described in issue #1790.
So the PR fix the problem as is in the issue, it just works.

By the way, fully remove **#include <fileapi.h>** is also ok, but as for the code consistency, I think keeping it is a better way.

### DIFF
--- a/include/spdlog/sinks/stdout_sinks-inl.h
+++ b/include/spdlog/sinks/stdout_sinks-inl.h
@@ -15,7 +15,11 @@
 // under windows using fwrite to non-binary stream results in \r\r\n (see issue #1675)
 // so instead we use ::FileWrite
 #include <spdlog/details/windows_include.h>
+
+#ifndef _USING_V110_SDK71_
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/stdout_sinks-inl.h
```diff
@@ -15,7 +15,11 @@
 // under windows using fwrite to non-binary stream results in \r\r\n (see issue #1675)
 // so instead we use ::FileWrite
 #include <spdlog/details/windows_include.h>
+
+#ifndef _USING_V110_SDK71_
```
gabime, line 19: Please add comment about the purpose of this ifdef. It woild hard to understand without it.

LonghronShen: The **_USING_V110_SDK71_** guard works when building based on a pre-8 version of Windows SDK, or trying to use the XP-toolset like **vc141_xp**.

PR #1909, opened 2021-04-07
Removed forward declarations

ChristianPanov:
Since forward declarations are present in the fwd.h header, having them here is redundant


gabime:
But fwd.h need to be included instead of the removed code

PR #1913, opened 2021-04-08
Update pattern_formatter-inl.h full_formatter comment to describe the default logging pattern expression accurately

bsergean:
Comment describing the default logging pattern is missing [%s:%#], which is the abbreviated source file + the line number.

I tried to customize our own logger by copy pasting this info, and then I noticed we had lost that information.


gabime:
Thanks @bsergean

bsergean:
You’re welcome !

PR #1918, opened 2021-04-22
Made mutex member variable mutable

matt77hias:
Classes inheriting from `base_sink` can now lock the base mutex inside their `const` member methods (e.g., basic accessors).


matt77hias:
For example, `ringbuffer_sink` could have a `const` member method to retrieve the number of log messages in the buffer, which is simultaneously read from/written to by different threads.

gabime:
Thanks @matt77hias

PR #1919, opened 2021-04-22
Ref-qualified memfns and added some move semantics

jharmer95:


gabime:
Tnanks @jharmer95 . Whats the use case here? I

jharmer95:
Right now this is still a draft, but the idea is to catch errors at compile-time.

- For example, if someone tried to get a reference to a member variable of an object that is about to be destroyed, it would not compile rather than that reference becoming dangling when the object is destroyed

```C++
// I know this isn't how a circular_q is constructed and wouldn't compile,
// but it demonstrates in a lot less lines
// This now won't compile, preventing a dangling reference
auto& x = circular_q { 1, 2, 3 }.front();

// This however would work because of the `&&` overload for `.front()` (moves T out of the queue)
auto x = circular_q { 1, 2, 3 }.front();
```

- Another example would be catching bugs from discarded operations. Setting the formatter for a sink that is going to be destroyed makes no sense as that formatter will never be used. So either you did not intend for the sink to be destroyed (finding a bug) or the `set_formatter` call is unneeded (improving performance). Admittedly, this isn't super useful because you would probably have to go out of your way to write code like this, but it is possible.

- Finally, by ref-qualifying assignment operators, you can detect bugs where `=` was used instead of `==` in equality conditions. It also ties in with the point above (why do you want to assign to an object that is about to be destroyed?)

```C++
struct foo
{
    foo(int x) : n(x) {}
    // Non-ref qualified default copy-assign
    foo& operator=(const foo&) = default;
    bool operator==(const foo& other) const { return n == other.n; }
    operator int() const { return n; }
    
    int n;
};

struct bar
{
    bar(int x) : n(x) {}
    // lvalue-ref qualified default copy-assign
    bar& operator=(const bar&) & = default;
    bool operator==(const bar& other) const { return n == other.n; }
    operator int() const { return n; }
    
    int n;
};

foo getFoo()
{
    return foo{4};
}

bar getBar()
{
    return bar{5};
}

void bool_func(bool B)
{
}

int main()
{
    // Oops, meant to check if they were equal, but wrote `=` instead of `==`
    
    // RUNTIME-BUG: implicitly converts foo to int to bool and assigns temp foo to 12
    bool_func(getFoo() = 12);
    
    // COMPILE-ERROR: `bar& operator=(bar) &&` does not exist
    bool_func(getBar() = 12);
}
```

- There's also some mutexes that I made `mutable` so that some trivial "getters" can be const-qualified

As I said, right now this is a draft, I want to make sure I do everything right.
It is of low priority, but I am digging deeper into C++ move semantics and improving my own projects and wanted to see how I could contribute to a library I use in pretty much everywhere. I appreciate your feedback and if this is coming across as unnecessary or unwanted, please let me know. I'm just happy to contribute!

gabime:
> bar& operator=(const bar&) & = default;
 
So move constructor/assignment is not possible anymore for bar?

While it might help mitigating few bugs, It would change lots of code and (probably break) API, so I personally don’t think it worth it.

jharmer95:
No sorry. This is just for this example code. In this case you would also have the move assignment defined as well. `bar& operator=(bar&&) & = default;`

I only added ref-qualifiers to assignment operators that were already explicitly defined in the codebase (r-value and l-value if both existed before). This does not make any breaking changes at all, current move operations are intact.

Sorry for the confusion

gabime:
Thanks for the clarification.
In any case I am reluctant to modify so much code for this, so I will close for now.

PR #1920, opened 2021-04-27
Check more I/O errors

pck:
Hello

With buffered I/O most of the work might be done by fflush().
Only checking fwrite() might not report common problems from failing (ro-remounted, removed) volumes.

I'd suggest to at least improve file_helper::flush.
I suppose it can't hurt to improve (potentially redirected) stdout/stderr/ansi-sinks as well.

Kind regards, Paul


gabime:
Thanks @pck . I will consider this. I agree about the fflush checking, but not sure about the fwrite checks in the stdout sinks. seems excessive and might even hurt performance.

PR #1927, opened 2021-05-06
Avoid harmless warning about unreachable statement in MSVS build

vadz:
Don't compile "return 0" after throw_spdlog_ex() when using MSVS as this
compiler (at least its 2019 edition) is smart enough to warn about this
statement being unreachable after throw inside this inline function.

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -246,7 +246,11 @@ SPDLOG_INLINE size_t filesize(FILE *f)
 #endif
 #endif
     throw_spdlog_ex("Failed getting file size from fd", errno);
+
+    // avoid C4702 warning about unreachable statement.
+#ifndef _MSC_VER
     return 0; // will not be reached.
+#endif
 }
 
 // Return utc offset in minutes or throw spdlog_ex on failure


gabime:
What would happen with older msvc version? we try to support ver 2103 and up.

vadz:
2103 is MSVS 2013, right?

The worst that _can_ happen is that other versions warn about the missing return and it's arguably better to for the old versions to warn about this than for new versions to warn about unreachable code. I could change this to check for MSVS >= 2017 only, but if it's really important I could try recompiling with 2013 (I didn't know that spdlog still supported it to be honest).

gabime:
> The worst that can happen is that other versions warn about the missing return

The worst is that they would issue error or even UB. who kmows

vadz:
No, the worst that could happen is really a warning. But I'll check with 2013 or restrict the check to 2015 and later (which are definitely fine, as can be already seen in the CI builds).

vadz:
Sorry, MSVS 2013 doesn't seem to be supported by the bundled fmt:
```
c:\...\spdlog\src>cl /c /EHsc /W4 /DSPDLOG_COMPILED_LIB /I../include spdlog.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 18.00.40629 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

spdlog.cpp
../include\spdlog/fmt/bundled/core.h(282) : error C2146: syntax error : missing ';' before identifier 'T'
```
and plenty of other errors follow.

So I think this can be applied without any problems as 2015, 2017 and 2019 all give the same "unreachable" warning.

gabime:
still, I believe we should suppress the warning,not the return statement itself which is perfectly legal.

vadz:
FWIW I disagree because this statement is really unreachable (even when exceptions are disabled, there is still a call to `abort()` before it), but, worse, using `#pragma warning(disable: 4702)` inside the function doesn't work, so it has to be done outside of it, which is IMHO more ugly.

But here is  #1930 doing it like this if you prefer.

gabime:
Thanks. I agree it is uglier but feels much better :)

PR #1931, opened 2021-05-06
Fix warning about testing _WIN64 which might be undefined

vadz:
This warning is disabled by default, but is pretty useful and worth
enabling for MSVC, just as -Wundef for gcc, so fix it in Win32 build.

---

Sorry for another trivial PR, but while testing #1930 I've also stumbled on this one in Win32 build.


gabime:
Thanks @vadz

PR #1933, opened 2021-05-09
small std::find conversion

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/include/spdlog/common-inl.h
+++ b/include/spdlog/common-inl.h
@@ -7,6 +7,9 @@
 #include <spdlog/common.h>
 #endif
 
+#include <algorithm>
+#include <iterator>
+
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. Please add includes for algorithm and for iterator.

neheb:
Iterator? I think algorithm is enough.

neheb:
Added both.

gabime:
The includes should be  only in the common-inl.h file(so user code not forced to include those  in compiled version)

neheb:
Done. It honestly sounds like a job for iwyu though...

gabime:
Thanks @neheb .

PR #1946, opened 2021-05-16
Add support for CMake policy CMP0077

### LINKED ISSUE #1945: Update cmake_minimum_requrired to 3.13
I am using spdlog as CMake subdirectory (through ```add_subdirectory```), and am unable to set configuration options (such as ```SPDLOG_FMT_EXTERNAL```) in my parent CMakeLists.txt file. This is because ```cmake_minimum_required(VERSION 3.10)``` unsets CMake policy ```CMP0077```.

```CMP0077``` unsets normal variables (i.e. a parent-project required setting for ```SPDLOG_FMT_EXTERNAL``` etc), and forces the default option. I can not force the policy to be set as the current ```cmake_minimum_required(VERSION 3.10)``` overrides my own policy / version settings.

The new (more intuitive) behaviour was introduced in cmake 3.13: https://cmake.org/cmake/help/latest/policy/CMP0077.html#policy:CMP0077

jafj:
Closes #1945

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,13 @@ message(STATUS "Build spdlog: ${SPDLOG_VERSION}")
 include(GNUInstallDirs)
 
 # ---------------------------------------------------------------------------------------
+# Set CMake policies to support later version behaviour
+# ---------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @jafj

PR #1958, opened 2021-05-30
Allow compilation with nvc++ (and possibly PGI)

mlund:
Using CMake, Nvidia's `nvc++` adds the `-A` compiler flag ("strict"), causing compilation error if newlines are missing from the end of source files. `nvc++` (not to be confused with `nvcc`) is detected as the PGI C++ compiler and this patch may therefore also apply to that. Proposed changes involve newlines, only.


gabime:
Thanks @mlund

PR #1970, opened 2021-06-13
Fix signed/unsigned mismatch in VS

bansan85:
When building under VS, I have:

```
[build] [...]\vcpkg\installed\x64-windows\include\spdlog\details\os-inl.h(362,25): warning C4365: 'argument': conversion from 'int' to 'DWORD', signed/unsigned mismatch [[...].vcxproj]
```

This small patch just fix it.

It should be compatible with `std::chrono::milliseconds` where the type of the argument is part of the template;

### DIFF
--- a/include/spdlog/async.h
+++ b/include/spdlog/async.h
@@ -46,7 +46,7 @@ struct async_factory_impl
         auto tp = registry_inst.get_tp();
         if (tp == nullptr)
         {
-            tp = std::make_shared<details::thread_pool>(details::default_async_q_size, 1);
+            tp = std::make_shared<details::thread_pool>(details::default_async_q_size, 1U);
             registry_inst.set_tp(tp);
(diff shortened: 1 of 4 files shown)


gabime:
Wouldnr that cause warning about int to unsigned int convertions? e.g here https://github.com/gabime/spdlog/blob/be14e60d9e8be31735dd9d2d132d8a4cd3482165/include/spdlog/details/file_helper-inl.h#L58

bansan85:
You're right. I should have checked. I only checked if the library build without warning only for my project and not for the whole library.

I also fix two other `signed/unsigned mismatch`.

I only modify source to remove warnings. I know that I probably should have wrote `unsigned int x = 10U` but VS didn't complains.

gabime:
Thanks @bansan85

PR #1971, opened 2021-06-15
Use std::function for the global error handler

SpriteOvO:
```cpp
// error C2664: 'void spdlog::set_error_handler(void (__cdecl *)(const std::string &))': cannot convert argument 1 from '<lambda_1>' to 'void (__cdecl *)(const std::string &)'
//
spdlog::set_error_handler([&](const std::string &message) { /* ... */ });
spdlog::set_error_handler([=](const std::string &message) { /* ... */ });
```

The pure pointer function type error handler cannot store `lambda with capture` or `std::bind`.

I think this might be a mistake? Because the error handler in `spdlog::logger` is using `err_handler` (`std::function`).

This PR changes the global error handler type from a pure pointer function (`void (*)(const std::string &msg)`) to `err_handler` (`std::function<void(const std::string &err_msg)>`) to fix that problem.

### DIFF
--- a/include/spdlog/details/registry-inl.h
+++ b/include/spdlog/details/registry-inl.h
@@ -195,14 +195,14 @@ SPDLOG_INLINE void registry::flush_every(std::chrono::seconds interval)
     periodic_flusher_ = details::make_unique<periodic_worker>(clbk, interval);
 }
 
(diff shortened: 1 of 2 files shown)


gabime:
Thanks @SpriteOvO . Merged.

PR #1972, opened 2021-06-16
Fix runtime when build with -fsanitize=cfi

bansan85:
-flto=thin -fvisibility=hidden -fno-sanitize-trap=cfi

See https://gcc.gnu.org/wiki/Visibility

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -22,13 +22,17 @@
 
 #ifdef SPDLOG_COMPILED_LIB
 #    undef SPDLOG_HEADER_ONLY
-#    if defined(_WIN32) && defined(SPDLOG_SHARED_LIB)
-#        ifdef spdlog_EXPORTS
-#            define SPDLOG_API __declspec(dllexport)
-#        else
-#            define SPDLOG_API __declspec(dllimport)
+#    if defined(SPDLOG_SHARED_LIB)
+#        if defined(_WIN32)
+#            ifdef spdlog_EXPORTS
+#                define SPDLOG_API __declspec(dllexport)
+#            else // !spdlog_EXPORTS
+#                define SPDLOG_API __declspec(dllimport)
+#            endif
+#        else // !defined(_WIN32)
+#            define SPDLOG_API __attribute__((visibility ("default")))
 #        endif
-#    else // !defined(_WIN32) || !defined(SPDLOG_SHARED_LIB)
+#    else // !defined(SPDLOG_SHARED_LIB)
 #        define SPDLOG_API
 #    endif
 #    define SPDLOG_INLINE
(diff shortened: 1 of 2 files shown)


bansan85:
I tried to build and run a program with `-fsanitize=cfi` (and `-flto=thin -fvisibility=hidden -fno-sanitize-trap=cfi`) according to the [clang documentation](https://clang.llvm.org/docs/ControlFlowIntegrity.html).

So with this option, all class are by default hidden.

According to [GCC documentation](https://gcc.gnu.org/wiki/Visibility), the _C++ visibility support_ should be `__attribute__ ((visibility ("default")))` for `SPDLOG_API`.

bansan85:
I now have a doubt. Maybe the `__attribute__ ((visibility ("default")))` should not be applied in header_only mode.

gabime:
> I now have a doubt. Maybe the __attribute__ ((visibility ("default"))) should not be applied in header_only mode / static lib.

Yes, need to be careful here not to break backward compatibility.

Here is the closest I could find (I didn't test it. Please make sure it indeed works):

```c++
#ifdef SPDLOG_COMPILED_LIB
#   if defined(SPDLOG_SHARED_LIB) && defined(_WIN32)
#       ifdef spdlog_EXPORTS
#           define SPDLOG_API __declspec(dllexport)
#       else
#           define SPDLOG_API __declspec(dllimport)
#       endif
#   elif defined(SPDLOG_SHARED_LIB) && (defined(__GNUC__) || defined(__clang__))
#       define SPDLOG_API __attribute__((visibility("default")))
#   else // if not SPDLOG_SHARED_LIB or other compiler
#       define SPDLOG_API
#   endif
#   define SPDLOG_INLINE
#   undef SPDLOG_HEADER_ONLY
#else // !defined(SPDLOG_COMPILED_LIB)
#   define SPDLOG_API
#   define SPDLOG_HEADER_ONLY
#   define SPDLOG_INLINE inline
#endif // #ifdef SPDLOG_COMPILED_LIB
```

gabime:
@bansan85 any news on this pr?

bansan85:
Sorry. Thanks for reminding me.

I have to take some time to do it. Give me 14 days so I can make so good tests.

gabime:
@bansan85 Any news?

bansan85:
I'm sorry :confused: I think the code should work but at the current time, I don't have the motivation to write some tests :weary:

bansan85:
I finally took some time to test.

Run the command:

```
/mnt/c/temp/spdlog-1.9.2 # rm -Rf build; cmake -S . -B build -DCMAKE_CXX_COMPILER=clang++ -DSPDLOG_BUILD_SHARED=yes -DCMAKE_CXX_FLAGS="-O3 -flto=thin -fvisibility=hidden -fno-sanitize-trap=cfi"; VERBOSE=1 cmake --build build/ --target all --config Release --parallel
```

gabime:
Thanks @bansan85

PR #1984, opened 2021-06-28
Add a color-terminal type

hctym1995:
I am using spdlog in RISC-V64 machine,
and the COLORTERM in this machine is "vt102",
so open a pr


gabime:
Thanks @hctym1995

Issue #1898, opened 2021-03-30
Logging in Qt Widget


Schoppenglas:
How can I log into a Qt Widget eg QTextEdit? In my application I dont have a console, so I want to log into a Qt Widget.

```c++
QTextEdit *myTextEdit = new QTextEdit;
myTextEdit->append( [where do I get the logging string from?] );
```

Or is there another way?

Schoppenglas:
Ok. That was easier then I tought. Here is my QTextEdit solution:

```c++
#pragma once
//---
#include "spdlog/sinks/base_sink.h"
#include "spdlog/details/null_mutex.h"
//--
#include <mutex>
//--
#include <QTextEdit>

template<typename Mutex>
class qtextedit_sink : public spdlog::sinks::base_sink<Mutex>
{
public:
    explicit qtextedit_sink(QTextEdit *qtTextEdit) : qtTextEdit_(qtTextEdit) {}
    qtextedit_sink(const qtextedit_sink &) = delete;
    qtextedit_sink &operator=(const qtextedit_sink &) = delete;

protected:
    void sink_it_(const spdlog::details::log_msg &msg) override
    {
        spdlog::memory_buf_t formatted;
        spdlog::sinks::base_sink<Mutex>::formatter_->format(msg, formatted);

        if (qtTextEdit_)
        {
            qtTextEdit_->insertPlainText(fmt::to_string(formatted).c_str());
        }
    }

    void flush_() override {}

private:
    QTextEdit *qtTextEdit_{ nullptr };
};

using qtextedit_sink_mt = qtextedit_sink<std::mutex>;
using qtextedit_sink_st = qtextedit_sink<spdlog::details::null_mutex>;
```

And then, add to an existing "logger" with:

```c++
auto qtSink = std::make_shared<qtextedit_sink_mt>(ui->textEdit);
spdlog::get("logger")->sinks().push_back(qtSink);
```

gabime:
Cool. Thanks for sharing.

WorstCodeWay:
Great work! Thanks.

PR #1986, opened 2021-06-28
Added QTextEdit and QPlainTextEdit sink

mguludag:
Hello I've added custom sink for show logs in QTextEdit or QPlainTextEdit widget. Its only dependency is Qt.


gabime:
Thanks. Could you please merge all to single file called “qt_sinks” ?

include/spdlog/sinks/qplaintextedit_sink.h
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    ~qplaintextedit_sink() { flush_(); }
+
+protected:
+    void sink_it_(const details::log_msg &msg) override {
+        memory_buf_t formatted;
+        base_sink<Mutex>::formatter_->format(msg, formatted);
+        auto str = std::string(formatted.begin(), formatted.end() - 2);
```
gabime, line 37: Is there a way to avoid the expensive creation of std string? maybe use a string view?

include/spdlog/sinks/qtextedit_sink.h
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  ~qtextedit_sink() { flush_(); }
+
+protected:
+  void sink_it_(const details::log_msg &msg) override {
+    memory_buf_t formatted;
+    base_sink<Mutex>::formatter_->format(msg, formatted);
+    auto str = std::string(formatted.begin(), formatted.end() - 2);
```
gabime, line 37: same

gabime (changes requested):

include/spdlog/sinks/qplaintextedit_sink.h
mguludag: Ok. I changed it to `spdlog::string_view_t` but widget's slots accepts `QString` so constructed `QString` and emit its signal.

gabime:
Thanks @mguludag

FlorianWolters:
It took me some time to figure out how-to use `qtextedit_sink_mt` in `qt_sinks.h` with CMake v3.21.0 and spdlog v1.9.0.

The following minimal CMake example demonstrates how-to create the MOC generated C++ source file required for `qt_sinks.h`  **without** using `CMAKE_AUTOMOC:BOOL=TRUE`. Using `CMAKE_AUTOMOC` could work in case spdlog is part of the CMake project itself and not an external "installation" by the means of `find_package` is used.

    cmake_minimum_required(VERSION 3.21)

    project(spdlog_qt_usage LANGUAGES CXX)

    find_package(Qt5 REQUIRED CONFIG COMPONENTS Widgets)
    find_package(spdlog REQUIRED CONFIG)

    get_target_property(spdlog_include_dir spdlog::spdlog INTERFACE_INCLUDE_DIRECTORIES)
    qt5_wrap_cpp(spdlog_qt_moc_source_files "${spdlog_include_dir}/spdlog/sinks/qt_sinks.h")

    add_executable(main)
    target_link_libraries(main
      PRIVATE
        Qt::Widgets
        spdlog::spdlog
      )
    target_sources(main
      PRIVATE
        main.cpp
        ${spdlog_qt_moc_source_files}
      )

Questions:

1. Are there any plans to simplify this?
2. Is it really required that the "private" classes in `qt_sinks.h` inherit from `QObject` and use `Q_OBJECT`? What are the actual use cases for this? Shouldn't it work without the direct dependency to `moc`?

I think it would be great to have a solution in spdlog that doesn't require to run `moc` on a spdlog header file (though I think I am missing something here), e.g. refer to a solution like https://github.com/gabime/spdlog/issues/1898, which doesn't use the class and macro.

gabime:
> Are there any plans to simplify this?

A PR to simplify would be most welcome.

> Is it really required that the "private" classes in qt_sinks.h inherit from QObject and use Q_OBJECT? What are the actual use cases for this? Shouldn't it work without the direct dependency to moc?

If you could find a way to simplify this as well, Would be great.

mguludag:
> It took me some time to figure out how-to use `qtextedit_sink_mt` in `qt_sinks.h` with CMake v3.21.0 and spdlog v1.9.0.
> 
> The following minimal CMake example demonstrates how-to create the MOC generated C++ source file required for `qt_sinks.h` **without** using `CMAKE_AUTOMOC:BOOL=TRUE`. Using `CMAKE_AUTOMOC` could work in case spdlog is part of the CMake project itself and not an external "installation" by the means of `find_package` is used.
> 
> ```
> cmake_minimum_required(VERSION 3.21)
> 
> project(spdlog_qt_usage LANGUAGES CXX)
> 
> find_package(Qt5 REQUIRED CONFIG COMPONENTS Widgets)
> find_package(spdlog REQUIRED CONFIG)
> 
> get_target_property(spdlog_include_dir spdlog::spdlog INTERFACE_INCLUDE_DIRECTORIES)
> qt5_wrap_cpp(spdlog_qt_moc_source_files "${spdlog_include_dir}/spdlog/sinks/qt_sinks.h")
> 
> add_executable(main)
> target_link_libraries(main
>   PRIVATE
>     Qt::Widgets
>     spdlog::spdlog
>   )
> target_sources(main
>   PRIVATE
>     main.cpp
>     ${spdlog_qt_moc_source_files}
>   )
> ```
> 
> Questions:
> 
> 1. Are there any plans to simplify this?
> 2. Is it really required that the "private" classes in `qt_sinks.h` inherit from `QObject` and use `Q_OBJECT`? What are the actual use cases for this? Shouldn't it work without the direct dependency to `moc`?
> 
> I think it would be great to have a solution in spdlog that doesn't require to run `moc` on a spdlog header file (though I think I am missing something here), e.g. refer to a solution like #1898, which doesn't use the class and macro.

Thanks for the idea. I will test a simple solution and PR :)

PR #2002, opened 2021-07-17
[feature] Support dynamic tag setting in android.

lilinxiong:
[feature] Support dynamic tag setting in android.
[change] Modify the Android example, and other examples fail to run error.

Support for dynamically modifying the TAG value in Android is achieved by adding a tag parameter to sink_it_ in base_sink.

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -261,8 +261,8 @@ void syslog_example()
 void android_example()
 {
     std::string tag = "spdlog-android";
-    auto android_logger = spdlog::android_logger_mt("android", tag);
-    android_logger->critical("Use \"adb shell logcat\" to view this message.");
+    auto android_logger = spdlog::android_logger_mt("android");
+    android_logger->critical("cus_critical_tag", "Use \"adb shell logcat\" to view this message.");
 }
 #endif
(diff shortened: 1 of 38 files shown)


lilinxiong:
Oh~ I'm sorry, I made the modification under Mac yesterday, and forgot to consider win, I will modify it now.

gabime:
Sorry seems there is misunderstanding. All that is needed is to modify a *single* file - the android sink, not anything else.

lilinxiong:
@gabime I tried to modify only android_sink. Although it seems that you can change different TAGs in the log, if you want to use macros (such as obtaining the name of the cpp file, the number of lines, and the name of the function), it is not very convenient. That's why I changed base_sink.

lilinxiong:
After the modification, you can print the log as normal, only ```LOG_INFO("TAG", "code: {0}", 1);``` can print the cpp file name, line number, function name, etc. as before Up.

gabime:
Sorry, too many changes. Closing.

lilinxiong:
Well, it doesn't matter. Thank you for opening up such a great log library.

PR #2008, opened 2021-07-21
Enable format string compile time validation

dkavolis:
`fmt` 8.0 changed how format strings are validated but `spdlog` converted the strings to string views before the validation was done.


gabime:
Thanks. Unfortunately it won't compile under  c++20

for example:
```c++
spdlog::info("{:d}", 42);
```

produces  (gcc 11, c++20)
```console
In file included from /home/gabi/spdlog/include/spdlog/spdlog.h:14,
                 from /home/gabi/spdlog/example/example.cpp:24:
/home/gabi/spdlog/include/spdlog/logger.h: In instantiation of ‘void spdlog::logger::log(spdlog::source_loc, spdlog::level::level_enum, const FormatString&, Args&& ...) [with FormatString = char [5]; Args = {const char (&)[13]}]’:
/home/gabi/spdlog/include/spdlog/logger.h:89:12:   required from ‘void spdlog::logger::log(spdlog::level::level_enum, const FormatString&, Args&& ...) [with FormatString = char [5]; Args = {const char (&)[13]}]’
/home/gabi/spdlog/include/spdlog/logger.h:160:12:   required from ‘void spdlog::logger::info(const FormatString&, Args&& ...) [with FormatString = char [5]; Args = {const char (&)[13]}]’
/home/gabi/spdlog/include/spdlog/spdlog.h:157:31:   required from ‘void spdlog::info(const FormatString&, Args&& ...) [with FormatString = char [5]; Args = {const char (&)[13]}]’
/home/gabi/spdlog/example/example.cpp:30:17:   required from here
/home/gabi/spdlog/include/spdlog/logger.h:83:13: error: ‘fmt’ is not a constant expression
   83 |         log_(loc, lvl, fmt, std::forward<Args>(args)...);
      |         ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Under c++17/14/11 it does compile but I didn't get  any validation happening.
For example
```c++
spdlog::info("{:d}", "not a number");
```

 didn't produce any compile time errrors.

dkavolis:
Ok, `fmt::basic_format_string` needs to be constructed in the method that's called, otherwise it will result in non-constant expression. This adds separate overloads for `fmt::format_string` and `fmt::wformat_string` since `Char` template argument cannot be deduced. One caveat though is that now every type convertible to string views is treated as format string since `FMT_STRING` is no longer needed to distinguish between compile time and runtime format strings.

`consteval` in MSVC is broken so you still need `FMT_STRING` to validate format strings at compile time.

gabime:
Thanks. Tried your latest commits.

Seems to produces compile-time error only wth c++20.

So, I wonder whether all those changes worth it considering that c++20 is yet to gain track and not very well supported.

BTW, how come moving the ```#include "spdlog/fmt/ostr.h"``` to the top fixes the ```user_defined_example()``` under c++20?

dkavolis:
> BTW, how come moving the `#include "spdlog/fmt/ostr.h"` to the top fixes the `user_defined_example()` under c++20?

Not sure, something before the second `#include "spdlog/fmt/ostr.h"` uses the fallback formatter before a partial specialization in `fmt/ostr.h`.

> So, I wonder whether all those changes worth it considering that c++20 is yet to gain track and not very well supported.
 
Some changes are needed to even support `FMT_STRING` compile time validation since the strings were converted to `string_view` before it was done. `fmt` already supports module import even though modules are largely broken on all compilers. And a useful subset of C++20 already works.

gabime:
Lets go for the minimal support for FMT_STRING (which I beleive can be achieved just by fixing the log_(..) functions and nothing else).
For later releases, we can gradualy add the more support compile time strings..

dkavolis:
Compile time format string validation is done in `fmt::basic_format_string` constructor which is changed since `fmt` 8.0. Just tried GCC C++20 but constructing `fmt::basic_format_string` from a string wrapped in `FMT_STRING` will produce a non-constant expression error. So it's either all strings will be checked at runtime in C++20 and will have to be wrapped in `fmt::runtime` or it will be simpler to use `fmt::basic_format_string` from the start rather than trying to do validation only in `log_` functions. I can add zero argument overloads for string view convertible types to do no validation to keep the current behaviour instead but I'm not sure this behaviour is desirable.

gabime:
> but I'm not sure this behaviour is desirable.

why not?

dkavolis:
`"Format string {}"` would compile without errors even if the format argument is missing.

gabime:
On the other hand the vast use cases are just simple strings, so something  simple  info(“some log”) should just simply logged without any expensive compile time/runtime extra formatting stages.

> fmt::make_args_checked works but string checking behaviour would differ from fmt in C++20 which may be unexpected to some users.

So lets go for it. Small steps is my preference here.

dkavolis:
#2010 I'll leave this up since log functions will still have to use `fmt::basic_format_string` arguments to match `fmt`.

> On the other hand the vast use cases are just simple strings, so something simple info(“some log”) should just simply logged without any expensive compile time/runtime extra formatting stages.

Instead of using a separate log function, the format function could be overloaded to do no formatting when `sizeof(Args...) == 0` but you would still get compile time checks with `FMT_STRING`/C++20

dkavolis:
Ok, I've only replaced `FormatString` template arguments with `fmt::basic_format_string` so there's no functionality change except C++20 validating format string literals at compile time with working `consteval`.

gabime:
1. Shouldn’t it be `const fmt::format_string<Args...> &fmt` instead by value?
2. Do the FMT_STRING still works with c++11/14/17 ?

dkavolis:
> 
> 
>     1. Shouldn’t it be `const fmt::format_string<Args...> &fmt` instead by value?
> 
>     2. Does the FMT_STRING still work with c++11/14/17 ?

1. `fmt::basic_format_string` is a wrapper around `basic_string_view` so it's fine passing it by value, may be even faster than passing by reference
2. Yes, without `consteval` `basic_format_string` constructor falls back to using the old argument check function

https://github.com/fmtlib/fmt/blob/8465869d7b04ae00f657332b3374ecf95b5e49c0/include/fmt/core.h#L2827-L2854

gabime:
This is great. Thanks !

PR #2015, opened 2021-07-27
Trim the newline-chars instead of removing of 2 chars

MadMax411:
Removing 2 chars (CR/LF) is under Windows okay, but on Linux only one char (LF) has to be deleted. Otherwise a char of the content is removed. So the fix is to use the QString method `trimmed()` to remove the CR/LF or LF chars.


gabime:
Thanks @MadMax411

PR #2016, opened 2021-07-27
Simplified Qt sinks

mguludag:
Thanks @FlorianWolters for [this](https://github.com/gabime/spdlog/pull/1986#issuecomment-886739045) idea. New header is more simpler and no private classes derived from QObject. Its also thread-safe (logging from any thread is safe)


mguludag (commented):
trim newline chars instead of remove 2 chars

include/spdlog/sinks/qt_sinks.h
```diff
@@ -1,200 +1,129 @@
...
+
+//
+// qtextedit_sink class
+//
+namespace spdlog {
+namespace sinks {
+template <typename Mutex>
+class qtextedit_sink : public base_sink<Mutex> {
+public:
+  qtextedit_sink(QTextEdit *textedit = nullptr) {
```
gabime, line 27: why allow nullptr by default if it throws if nullptr?

include/spdlog/sinks/qt_sinks.h
```diff
@@ -1,200 +1,129 @@
...
+        qtextedit = textedit;
+    } else {
+      throw spdlog_ex("Error opening QTextEdit");
+    }
+  }
+
+  ~qtextedit_sink() { flush_(); }
+
+protected:
+  void sink_it_(const details::log_msg &msg) override {
```
gabime, line 38: can never be null. why check it?

include/spdlog/sinks/qt_sinks.h
```diff
@@ -1,200 +1,129 @@
...
+    memory_buf_t formatted;
+    base_sink<Mutex>::formatter_->format(msg, formatted);
+    string_view_t str = string_view_t(formatted.data(), formatted.size());
+    QMetaObject::invokeMethod(qtextedit,"append", Qt::AutoConnection, Q_ARG(QString, QString::fromUtf8(str.data(), static_cast<int>(str.size())).trimmed()));
+  }
+
+  void flush_() override {}
+
+private:
+    QTextEdit* qtextedit = nullptr;
```
gabime, line 50: please rename to qtextedit_ like rest of spdlog

gabime (changes requested):

mguludag (commented):
removed nullptr checks and renamed member vars

gabime:
Thanks. Please fix the conflict so i can merge

include/spdlog/sinks/qt_sinks.h
```diff
@@ -1,125 +1,125 @@
...
+
+  ~qtextedit_sink() { flush_(); }
+
+protected:
+  void sink_it_(const details::log_msg &msg) override {
+    memory_buf_t formatted;
+    base_sink<Mutex>::formatter_->format(msg, formatted);
+    string_view_t str = string_view_t(formatted.data(), formatted.size());
+    QMetaObject::invokeMethod(qtextedit_,"append", Qt::AutoConnection, Q_ARG(QString, QString::fromUtf8(str.data(), static_cast<int>(str.size())).trimmed()));
+  }
```
gabime, line 43: why the change from the simple append()? seems efficiant

mguludag: Direct calling GUI widget functions from another threads in Qt app results crash. Qt forces us to use their own Q_OBJECT macro and connect function or QMetaObject::invokemethod function.

gabime:
Thanks @mguludag

PR #2018, opened 2021-07-28
Added common class for all qt objects

mguludag:
Removed separate class layout for qt_sinks and also send logs to any custom qt (QObject) classes (QML, QFile, custom Widget etc.)

### DIFF
--- a/include/spdlog/sinks/qt_sinks.h
+++ b/include/spdlog/sinks/qt_sinks.h
@@ -4,8 +4,8 @@
 #pragma once
 
 //
-// Custom sink for QPlainTextEdit or QTextEdit and its childs(QTextBrowser... etc)
-// Building and using requires Qt library.
+// Custom sink for QPlainTextEdit or QTextEdit and its childs(QTextBrowser...
+// etc) Building and using requires Qt library.
 //
 
 #include "spdlog/common.h"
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/qt_sinks.h
```diff
@@ -4,122 +4,81 @@
...
 namespace sinks {
-template <typename Mutex>
-class qtextedit_sink : public base_sink<Mutex> {
+template <typename Mutex> class qt_sink : public base_sink<Mutex> {
 public:
-  qtextedit_sink(QTextEdit *textedit = nullptr) {
-    if (textedit != nullptr) {
-        qtextedit_ = textedit;
+  qt_sink(QObject *qt_object = nullptr, const std::string &meta_method = "") {
+    if (qt_object != nullptr) {
```
gabime, line 28: i dont understand why to allow default nullptr in ctor if it get thrown afgerward if nullptr

gabime (changes requested):

include/spdlog/sinks/qt_sinks.h
```diff
@@ -4,122 +4,81 @@
...
-                    QPlainTextEdit *qplaintextedit = nullptr) {
-    return Factory::template create<sinks::qplaintextedit_sink_st>(logger_name,
-                                                              qplaintextedit);
+qt_logger_st(const std::string &logger_name, T *qt_object = nullptr, const std::string &meta_method = "") {
+  std::string meta_method_ = meta_method;
+  if (std::is_base_of<QTextEdit, T>::value)
+    meta_method_ = "append";
+  else if (std::is_same<T, QPlainTextEdit>::value)
+    meta_method_ = "appendPlainText";
+  return Factory::template create<sinks::qt_sink_st>(logger_name, qt_object, meta_method_);
```
gabime, line 82: T is known at compile time, so template specialization should be used instead of the if statements.

gabime (changes requested):

include/spdlog/sinks/qt_sinks.h
mguludag: Actually it doesn't matter(nullptr or qobject*) in last solution 🤔

gabime: why not?

include/spdlog/sinks/qt_sinks.h
mguludag: I will convert solution with template specialization

include/spdlog/sinks/qt_sinks.h
mguludag: I think pass nullptr into QMetaObject::invokemethod function is valid but does nothing with nullptr. But with a valid QObject*  object it will invoke a function named in meta_method string.

mguludag (commented):
Cleaned qt_sink ctor from unnecessary nullptr checks and added factory function overloads

PR #1981, opened 2021-06-27
added mongodb sink

mguludag:
Hello I've added custom sink for writing logs to mongodb.  If I will have free time I want to make dynamic mongodb document structure based on format string pattern.

## Dependencies
* [mongocxx](http://mongocxx.org/mongocxx-v3/installation/)

### asynchronous logger
```C++
#include <spdlog/async.h>
#include <spdlog/sinks/mongo_sink.h>
#include <spdlog/spdlog.h>
#include <thread>

int main() {
  std::thread t1([&] {
    auto logging = spdlog::async_factory::create<spdlog::sinks::mongo_sink>(
        "mongo_logger_1", "db", "collection");
    for (auto i = 0; i < 10; ++i) {
      logging->info("thread_1 iter: {}",i);
      std::this_thread::sleep_for(std::chrono::milliseconds(500));
    }
  });

  std::thread t2([&] {
    auto logging = spdlog::async_factory::create<spdlog::sinks::mongo_sink>(
        "mongo_logger_2", "db", "collection");
    for (auto i = 0; i < 10; ++i) {
      logging->info("thread_2 iter: {}",i);
      std::this_thread::sleep_for(std::chrono::milliseconds(100));
    }
  });

  std::this_thread::sleep_for(std::chrono::seconds(10));

  t1.join();
  t2.join();

  return 0;
}
```

### synchronous logger
```C++
#include <spdlog/sinks/mongo_sink.h>
#include <spdlog/spdlog.h>
#include <thread>

int main() {
    auto logging = spdlog::create<spdlog::sinks::mongo_sink>(
        "mongo_logger", "db", "collection");
  std::thread t1([&] {
    for (auto i = 0; i < 10; ++i) {
      logging->info("thread_1" + std::to_string(i));
      std::this_thread::sleep_for(std::chrono::milliseconds(500));
    }
  });

  std::thread t2([&] {
    for (auto i = 0; i < 10; ++i) {
      logging->info("thread_2" + std::to_string(i));
      std::this_thread::sleep_for(std::chrono::milliseconds(100));
    }
  });

  std::this_thread::sleep_for(std::chrono::seconds(10));

  t1.join();
  t2.join();

  return 0;
}
```

### Example document from mongodb
```JSON
{
  "_id": {
    "$oid": "60d87492314f000087006980"
  },
  "timestamp": {
    "$date": "2021-06-27T12:52:34.985Z"
  },
  "level": "info",
  "message": "thread_1 5",
  "logger_name": "mongo_logger1",
[...]

PR #2025, opened 2021-08-02
Removal of C++14-specific std::make_unique from mongo_sink.h

jabartek:
mongo_sink.h used `std::make_unique` instead of `spdlog::details::make_unique`, thus making it impossible to compile a project under C++11 which spdlog itself targets. While mongocxx requires use of polyfills to work under C++11, it should not be a case with a spdlog project that just links with it

### DIFF
--- a/include/spdlog/sinks/mongo_sink.h
+++ b/include/spdlog/sinks/mongo_sink.h
@@ -33,7 +33,7 @@ public:
     {
         try
         {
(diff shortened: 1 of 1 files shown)


jabartek:
I'm not sure how exactly #1981 passed the `GCC_VERSION=4.8 BUILD_TYPE=Release CPP=11` check - maybe this should be looked into as well. Sadly, I am not a CI person, so cannot really help with that

gabime:
Thanks @jabartek

PR #2026, opened 2021-08-03
fix include file twice in the same file

hbwang15:
there is one head file be included twice in the thread_pool.h, I just delete duplicate one.


gabime:
Thanks @hbwang15

PR #2029, opened 2021-08-04
Ensure exported package is relocatable

daverigby:
As per CMake's Importing and Exporting Guide[1],
configure_package_config_file() should be used for configuring the
package configuration file, not the regular configure_file() function.

This ensures that a spdlog package built on one system (with a given
directory tree) can be imported from a different system -
e.g. creating a pre-compiled spdlog package for use on different
systems.

[1]: https://cmake.org/cmake/help/git-stage/guide/importing-exporting/index.html#id8


gabime:
Thanks @daverigby

daverigby:
Thank you!

PR #2030, opened 2021-08-05
remove std::distance usage

neheb:
std::distance internally runs a loop, which may or may not be optimized
away. Just use simple arithmetic.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/include/spdlog/common-inl.h
+++ b/include/spdlog/common-inl.h
@@ -34,7 +34,7 @@ SPDLOG_INLINE spdlog::level::level_enum from_str(const std::string &name) SPDLOG
 {
     auto it = std::find(std::begin(level_string_views), std::end(level_string_views), name);
     if (it != std::end(level_string_views))
-        return static_cast<level::level_enum>(std::distance(std::begin(level_string_views), it));
+        return static_cast<level::level_enum>(it - std::begin(level_string_views));
 
     // check also for "warn" and "err" before giving up..
     if (name == "warn")


gabime:
Thanks @neheb

kelbon:
lol what

kin4stat:
```cpp
template<class InputIterator>
  constexpr typename iterator_traits<InputIterator>::difference_type
    distance(InputIterator first, InputIterator last);
```
Distance guarantees a substraction operator, which is the same as your "fix" in your PR, if InputIterator type meets RandomAccessIterator requirements.
[iterator.operations#5](http://eel.is/c%2B%2Bdraft/iterator.operations#5)

`std::begin` and `std::end` with C-array in arguments returns pointer to element, [[iterator.range-4]](https://eel.is/c++draft/iterators#iterator.range-4), [[iterator.range-55]](https://eel.is/c++draft/iterators#iterator.range-5)

`RandomAccessIterator` requirement can be seen here [[random.access.iterators#1]](https://eel.is/c++draft/random.access.iterators#1).

We can add and substract two pointers and pointer + intergral and get implementation defined integral result: [[expr.add-2.2]](https://eel.is/c++draft/expr.compound#expr.add-2.2) [[expr.add-2.3]](https://eel.is/c++draft/expr.compound#expr.add-2.3)
We can compare two pointers, and get bool result: [[expr.rel-4]](https://eel.is/c++draft/expr.compound#expr.rel-4)
We can subscript pointer with resulting type X&: [[expr.sub-2]](https://eel.is/c++draft/expr.compound#expr.sub-2)
We can add/substract and assign pointers in single expression: [[expr.ass-6]](https://eel.is/c++draft/expr.compound#expr.ass-6)

Thus, the pointer meets the `RandomAccessIterator` requirements, and then `std::distance` from `it` and `std::begin(level_string_views)` is the same as `it - std::begin(level_string_views)`

kin4stat:
@gabime please roll back changes

tt4g:
@kin4stat It would be better to open PR to revert changes.

PR #2038, opened 2021-08-09
add assert for null sink

hbwang15:
add assert  null sink


gabime:
Thanks. However this is hot path and i prefer keep it as fast as possible even in debug builds.

PR #2047, opened 2021-08-13
better file name for hourly file sink

seker:


seker:
better file name for hourly file sink. accordding to daily file sink

gabime:
Thanks @seker

PR #2048, opened 2021-08-13
Added additional information for error handler

D-r-P-3-p-p-3-r:
Useful when formatting log messages fails. Now you can tell which log message caused the problem.

### DIFF
--- a/include/spdlog/logger.h
+++ b/include/spdlog/logger.h
@@ -28,10 +28,24 @@
 #include <vector>
 
 #ifndef SPDLOG_NO_EXCEPTIONS
-#    define SPDLOG_LOGGER_CATCH(additional_info)                                                                                                          \
+#    define SPDLOG_LOGGER_CATCH(location)                                                                                                  \
         catch (const std::exception &ex)                                                                                                   \
         {                                                                                                                                  \
-            err_handler_(fmt::format("{} ({})", ex.what(), additional_info));                                                                                                       \
+            if(location.filename)                                                                                                          \
+            {                                                                                                                              \
+                try                                                                                                                        \
+                {                                                                                                                          \
+                    err_handler_(fmt::format("{} [{}({})]", ex.what(), location.filename, location.line));                                \
+                }                                                                                                                          \
+                catch (const std::exception &ex)                                                                                           \
+                {                                                                                                                          \
+                    err_handler_(ex.what());                                                                                               \
+                }                                                                                                                          \
+            }                                                                                                                              \
+            else                                                                                                                           \
+            {                                                                                                                              \
+                err_handler_(ex.what());                                                                                                   \
+            }                                                                                                                              \
         }                                                                                                                                  \
         catch (...)                                                                                                                        \
         {                                                                                                                                  \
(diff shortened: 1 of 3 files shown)


D-r-P-3-p-p-3-r:
Currently, there's still a problem under Windows with wchars...

gabime:
can you fix it?

D-r-P-3-p-p-3-r:
It's work in progress. I will continue checking it out when I'm back at work next Tuesday.

D-r-P-3-p-p-3-r:
Ok, I found the root cause.

It's actually SPDLOG_WCHAR_TO_UTF8_SUPPORT. The wstring_view_t format string/message can't just be dumped into fmt::format together with the exception message from what() in two versions of log_.

For now, I'm passing an empty string to SPDLOG_LOGGER_CATCH in those two versions of log(). Like that it's no worse than before under Windows.

But I'll still check for a nicer solution next week.

gabime:
Ok.  Thanks. I will leave this pr open until fully implemented.

D-r-P-3-p-p-3-r:
I got a working solution now which also works smoothly under Windows. Instead of emitting the offending log message, the location of the offending log message is reported (filename and line).
Thus, finding the problematic log message is still easy, but there is no trouble with wide chars. Also, using the original log message (using wide chars or not), could trigger the same exceptions that the original logging attempt could have. This problem is avoided as well.

include/spdlog/logger.h
```diff
@@ -28,18 +28,32 @@
...
+#    define SPDLOG_LOGGER_CATCH(location)                                                                                                  \
         catch (const std::exception &ex)                                                                                                   \
         {                                                                                                                                  \
-            err_handler_(ex.what());                                                                                                       \
+            if(location.filename)                                                                                                          \
+            {                                                                                                                              \
+                try                                                                                                                        \
+                {                                                                                                                          \
```
gabime, line 37: I dont think an extra try catch needed here. why would it fail? the lib is in total control of the source loc param

gabime (changes requested):

gabime:
Alos please note that by default the source_loc is empty unless the SPDLOG_INFO/DEBUG etc macros are used.

include/spdlog/logger.h
D-r-P-3-p-p-3-r: I just figured as fmt::format can throw (e.g. in case of OOM), it would be the safest bet to catch.
I'll remove it.

D-r-P-3-p-p-3-r:
> Alos please note that by default the source_loc is empty unless the SPDLOG_INFO/DEBUG etc macros are used.

In our case, our own log macros forward to SPDLOG_LOGGER_CALL or add the source_loc themselves. So that's cool.

include/spdlog/logger.h
gabime: Thanks, please remove it and I will merge.

gabime (approved):

gabime:
Thanks @D-r-P-3-p-p-3-r . Merged.

PR #2063, opened 2021-08-21
Fix uninitialized member warnings in v1.x as reported by GCC's `[-Weffc++]` mode

kcgen:
Also addresses one false-positive, as noted in the commit.

After this, spdlog v1.x builds clean against gcc 11.1.0, even with relatively strict compiler warnings: `-Wcast-align -Wdouble-promotion -Wduplicated-branches -Wduplicated-cond -Wformat=2 -Wlogical-op -Wmisleading-indentation -Wnull-dereference -Wshadow -Wunused -Wstrict-aliasing=2 -Woverloaded-virtual -Wuseless-cast -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -Wpedantic -Weffc++ -Wextra-semi`

Prior to this - the following warnings (just a small portion) are shown:


kcgen:
@gabime, 

We're trying to migrate to `spdlog` on the DOSBox Staging project to replace the existing internal logging functionality. 

We also care about warnings, and use GCC `-Weffc++` flag to gate our CI on high payoff warnings, as they often can lead to bugs or be the source of exploits (in general). We'd like to keep us using `-Weffc++` as it provides valuable feedback.
 
Can you try building spdlog with `-Weffc++` using a modern-ish version of GCC - and then try my branch? Unfortunately, I can't reproduce your Travis environment locally.  Looking for ways we can reconcile this.

kcgen:
Hmm - I see the 2.x branch is doing much better, with only a handful of repeating issues.
Maybe we can work on making it build clean with `Weffc++`?

gabime:
Thanks. It doesn’t pass the tests though.. Also, not sure I agree with those warnings. Why need to initialize members objects with {} if they are already guaranteed  be default constructed? see https://stackoverflow.com/a/14002558

kcgen:
Thanks for the quick assessment @gabime.

Indeed, gcc-4.8 is simply too outdated to despite the changes compiling clean using `stdc++11` on newer compilers.

I see branch `v2.x` is in significantly better shape.

Are you amenable to addressing its handful or remaining `Weffc++` warnings, assuming we can pass the Travis build and tests?

PR #2087, opened 2021-09-03
add callback while rotate the log file

seker:
add callback while rotate the log file


seker:
fixed compile error firstly

seker:
fixed compile error already.

PR #2089, opened 2021-09-03
add callback while rotate the log file

seker:
add callback while rotate the log file

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -255,6 +255,8 @@ struct source_loc
     const char *funcname{nullptr};
 };
 
+typedef void (*rotate_file_callback)(const filename_t& pre, const filename_t& next);
+
 namespace details {
 // make_unique support for pre c++14
 
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -68,12 +68,13 @@ class daily_file_sink final : public base_sink<Mutex>
 {
 public:
     // create daily file sink which rotates on given time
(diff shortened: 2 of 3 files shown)


gabime:
Thanks, whats the use case here?

seker:
> Thanks, whats the use case here?

when rotates the log file ，my app hope to upload pre log file to server.

gabime:
we need to define if the callback  is called after the old file closed or not.. There are use cases you want it to be closed  (e.g. you want to zip and rename). On the other hand sometimes you want it to be open (add last line before the file closes). Here is an example of typical callbacks types (from old [branch](https://github.com/gabime/spdlog/tree/file_event_handlers)):


```c++
struct file_event_handlers 
{
    std::function<void(const filename_t &filename, std::FILE *file_stream)> after_open;
    std::function<void(const filename_t &filename, std::FILE *file_stream)> before_close;
    std::function<void(const filename_t &filename)> after_close;
};
```

I suggest to continue the work of this branch.

seker:
> we need to define if the callback is called after the old file closed or not.. There are use cases you want it to be closed (e.g. you want to zip and rename). On the other hand sometimes you want it to be open (add last line before the file closes). Here is an example of typical callbacks types (from old [branch](https://github.com/gabime/spdlog/tree/file_event_handlers)):
> 
> ```c++
> struct file_event_handlers 
> {
>     std::function<void(const filename_t &filename, std::FILE *file_stream)> after_open;
>     std::function<void(const filename_t &filename, std::FILE *file_stream)> before_close;
>     std::function<void(const filename_t &filename)> after_close;
> };
> ```
> 
> I suggest to continue the work of this branch.

ok，thanks~

PR #1746, opened 2020-11-25
add udp_sink

vbirds:
add udp_skin support for udp log

PR #2090, opened 2021-09-03
Add udp_sink

CJLove:
This PR updates the prior udp_sink PR (#1746), cleaning up some code and adding initial Windows support.  Winsock seems more prone to drop UDP packets if sent too quickly, so this sink seems more usable on Linux.  I'm not a Windows developer so I'm not sure if I'm missing something.

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -214,13 +214,7 @@ void udp_example()
     spdlog::sinks::udp_sink_config cfg("127.0.0.1", 11091);
     auto my_logger = spdlog::udp_logger_mt("udplog", cfg);
     my_logger->set_level(spdlog::level::debug);
-    for (int i = 0; i < 10; i++) {
-        my_logger->info("hello world {}", i);
-#ifdef _WIN32
-        // sendto() on winsock will drop packets if sent too quickly
-        std::this_thread::sleep_for(std::chrono::milliseconds(40));
-#endif
-    }
+    my_logger->info("hello world");
 }
 
 // A logger with multiple sinks (stdout and file) - each with a different format and log level.
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -212,7 +212,7 @@ void stopwatch_example()
 #include "spdlog/sinks/udp_sink.h"
 void udp_example()
 {
-    spdlog::sinks::udp_sink_config cfg("192.168.1.129", 11091);
+    spdlog::sinks::udp_sink_config cfg("127.0.0.1", 11091);
     auto my_logger = spdlog::udp_logger_mt("udplog", cfg);
     my_logger->set_level(spdlog::level::debug);
     my_logger->info("hello world");
@@ -231,15 +231,15 @@ void multi_sink_example()
 
     spdlog::logger logger("multi_sink", {console_sink, file_sink});
     logger.set_level(spdlog::level::debug);
-    logger.warn("this should appear in both console and file");
-    logger.info("this message should not appear in the console, only in the file");
-}
+    logger.warn("this should 
 
 // User defined types logging by implementing operator<<
 struct my_type
 {
     int i;
-    template<typename OStream>
+    template<typename OStream>appear in both console and file");
+    logger.info("this message should not appear in the console, only in the file");
+}
     friend OStream &operator<<(OStream &os, const my_type &c)
     {
         return os << "[my_type i=" << c.i << "]";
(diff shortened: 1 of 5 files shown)


example/example.cpp
```diff
@@ -198,13 +201,28 @@ void trace_example()
...
 
+void udp_example()
+{
+    spdlog::sinks::udp_sink_config cfg("127.0.0.1", 11091);
+    auto my_logger = spdlog::udp_logger_mt("udplog", cfg);
+    my_logger->set_level(spdlog::level::debug);
+    for (int i = 0; i < 10; i++) {
+        my_logger->info("hello world {}", i);
+#ifdef _WIN32
+        // sendto() on winsock will drop packets if sent too quickly
```
gabime, line 220: It depends on lots of variables. Since this is just a samole I would keep it simple and remove the loop and the ifdef.

include/spdlog/sinks/udp_sink.h
```diff
@@ -0,0 +1,87 @@
...
+
+#include <mutex>
+#include <string>
+#include <chrono>
+#include <functional>
+
+#pragma once
+
+// Simple udp client sink
+// Connects to remote address and send the formatted log.
```
gabime, line 23: There is no connection in udp. Please remove this comment

gabime (changes requested):

include/spdlog/details/udp_client-windows.h
```diff
@@ -0,0 +1,128 @@
+// Copyright(c) 2015-present, Gabi Melman & spdlog contributors.
+// Distributed under the MIT License (http://opensource.org/licenses/MIT)
+
+#pragma once
+
+#define WIN32_LEAN_AND_MEAN
```
gabime, line 6: Please surround with `#ifndef WIN32_LEAN_AND_CLEAN`

gabime (changes requested):

include/spdlog/details/udp_client.h
```diff
@@ -0,0 +1,85 @@
...
+    // Send exactly n_bytes of the given data.
+    // On error close the connection and throw.
+    void send(const char *data, size_t n_bytes)
+    {
+        ssize_t toslen = 0;
+        socklen_t tolen = sizeof(struct sockaddr);
+        if (( toslen = sendto(socket_, data, n_bytes, 0, (struct sockaddr *)&sockAddr_, tolen)) == -1)
+        {
+            throw_spdlog_ex("sendto(2) failed", errno);
+            close();
```
gabime, line 80: close() will never be reached.

gabime (changes requested):

gabime:
Also, since the socket creation might be expensive operation, I think it is better created once in the constructor and not called again (i.e. no need to check is_init() on each log call).

include/spdlog/details/udp_client-windows.h
```diff
@@ -0,0 +1,130 @@
...
+        close();
+    }
+
+    void send(const char *data, size_t n_bytes)
+    {
+        socklen_t tolen = sizeof(struct sockaddr);
+        if (sendto(socket_, data, static_cast<int>(n_bytes), 0, (struct sockaddr *)&addr_, tolen) == -1)
+        {
+            throw_spdlog_ex("sendto(2) failed", errno);
+            close();
```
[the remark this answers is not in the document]
CJLove: moved close() before the exception is thrown

include/spdlog/details/udp_client.h
CJLove: moved close() before the exception is thrown.

CJLove:
Agreed that the socket creation and initialization can/should be moved to the constructor.  I'm inclined to leave the `is_init()` check in place as this would allow the sink to re-initialize the socket if an error had occurred and the previous socket was closed.

That being said it's your call and I can remove the `is_init()` check if you'd prefer.

gabime:
> to re-initialize the socket if an error had occurred and the previous socket was closed.

Problem is that there is good chance it would fail again and again in each log call. Lets keep it simple and remove it.

CJLove:
Done.  Thanks!

gabime:
Thanks @CJLove

gabime:
@CJLove fyi I committed some cleanups in the udp code (9bb66c00e9da7f17bedfeba49a0054e093eca160, 9bb66c00e9da7f17bedfeba49a0054e093eca160, 9bb66c00e9da7f17bedfeba49a0054e093eca160) to make it RAII.

PR #2098, opened 2021-09-08
CMake: Support <PackageName>_ROOT

RedDwarf69:
I am no expert in all the policies, but I have seen no issues when using it with cmake 3.21.2.
My main user-case is getting `fmt_ROOT` to work (CMP0074).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,13 +16,6 @@ message(STATUS "Build spdlog: ${SPDLOG_VERSION}")
 include(GNUInstallDirs)
 
 # ---------------------------------------------------------------------------------------
-# Set CMake policies to support later version behaviour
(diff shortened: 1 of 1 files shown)


gabime:
The lines
```
if(POLICY CMP0077)
    cmake_policy(SET CMP0077 NEW) # option() honors variables already set
endif()
```

were added in pr #1946. Do you think it was wrong ?

RedDwarf69:
No, it wasn't wrong. But after using `cmake_minimum_required(VERSION 3.10...3.21)` it becomes redundant. With it every cmake version knowing about CMP0077 will use CMP0077 NEW. In fact every cmake version <= 3.21 knowing about any CMPXXXX will use CMPXXXX NEW.

gabime:
Thanks. Merged

PR #2102, opened 2021-09-12
fix mongo_sink<std::mutex>::instance_ template

yzz-ihep:


gabime:
in which compiler version it wont compile without it?

yzz-ihep:
> in which compiler version it wont compile without it?

gcc8.3

Though use "template<>" there are other errors!

Issue #2139, opened 2021-10-16
There's no need to define the type of `Mutex mutex_` as mutable since class base_sink has no const method. 


sunlong169:
The related code snippet is seen at https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/sinks/base_sink.h line 40.

PR #2140, opened 2021-10-16
No need to define the Mutex mutex_ as mutable there is no const method.

sunlong169:
There's no need to define the Mutex mutex_ as mutable since class base_sink has no const method. Related issue is seen at https://github.com/gabime/spdlog/issues/2139


gabime:
Thanks @sunlong169

PR #2141, opened 2021-10-17
the abstract interfaces should be private other than protected

sunlong169:
The abstract interfaces of `class base_sink` should be private other than protected. These methods won't be invoked by any derived class.


gabime:
It might be right, but since this would break api for users of base_sink, It better not change.

PR #2157, opened 2021-11-04
Centralize and make accessible default log level (to level::info)

sjanel:
Hello,

I would like to access to the default level for a `logger` (currently hardcoded to `level::info` in `logger.h`):

```
protected:
    std::string name_;
    std::vector<sink_ptr> sinks_;
    spdlog::level_t level_{level::info};
    spdlog::level_t flush_level_{level::off};
    err_handler custom_err_handler_{nullptr};
    details::backtracer tracer_;
```

This PR defines a `constexpr` variable `kDefault` (name could be discussed, `default` is a keyword in C++ so not possible) such that client code can retrieve it. I took the opportunity to replace the `level::info` used here and there with this constant value.

WDYT?

Thanks in advance

### DIFF
--- a/include/spdlog/cfg/helpers-inl.h
+++ b/include/spdlog/cfg/helpers-inl.h
@@ -88,7 +88,7 @@ SPDLOG_INLINE void load_levels(const std::string &input)
 
     auto key_vals = extract_key_vals_(input);
     std::unordered_map<std::string, level::level_enum> levels;
-    level::level_enum global_level = level::info;
+    level::level_enum global_level = level::kDefault;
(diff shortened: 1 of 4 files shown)


gabime:
Thanks. Since the default is always info what’s the point?

sjanel:
Hi, this is for documentation and to avoid having this definition at several places in the code, it's just cleaner. In addition, it makes it available from client code.

gabime:
Thanks. However there are already too  many level related consts/macros/enums so I prefer not to add another.

PR #2165, opened 2021-11-10
add file event handlers

seker:
add file event handlers

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -120,7 +120,22 @@ void basic_example()
 void rotating_example()
 {
     // Create a file rotating logger with 5mb size max and 3 rotated files.
-    auto rotating_logger = spdlog::rotating_logger_mt("some_logger_name", "logs/rotating.txt", 1048576 * 5, 3);
+    spdlog::file_event_handlers_t file_event_handlers;
+        file_event_handlers.after_open = [](spdlog::filename_t filename, std::FILE* fstream)
+    {
+        fputs("OPEN!\r\n", fstream);
+        spdlog::info("basic_example() : file_event_handlers.after_open      : {}", filename.c_str());
+    };
+    file_event_handlers.before_close = [](spdlog::filename_t filename, std::FILE* fstream)
+    {
+        fputs("CLOSE!\r\n", fstream);
+        spdlog::info("basic_example() : file_event_handlers.before_close    : {}", filename.c_str());
+    };
+    file_event_handlers.after_close = [](spdlog::filename_t filename)
+    {
+        spdlog::info("basic_example() : file_event_handlers.after_close     : {}", filename.c_str());
+    };
+    auto rotating_logger = spdlog::rotating_logger_mt("some_logger_name", "logs/rotating.txt", 1048576 * 5, 3, false, file_event_handlers);
 }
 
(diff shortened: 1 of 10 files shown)


include/spdlog/sinks/basic_file_sink-inl.h
```diff
@@ -14,7 +14,8 @@ namespace spdlog {
 namespace sinks {
 
 template<typename Mutex>
-SPDLOG_INLINE basic_file_sink<Mutex>::basic_file_sink(const filename_t &filename, bool truncate)
+SPDLOG_INLINE basic_file_sink<Mutex>::basic_file_sink(const filename_t &filename, bool truncate, const file_event_handlers_t& event_handlers)
+    : file_helper_{std::move(event_handlers)}
```
gabime, line 18: const reference cannot be really moved

include/spdlog/sinks/daily_file_sink.h
```diff
@@ -68,10 +68,11 @@ class daily_file_sink final : public base_sink<Mutex>
 {
 public:
     // create daily file sink which rotates on given time
-    daily_file_sink(filename_t base_filename, int rotation_hour, int rotation_minute, bool truncate = false, uint16_t max_files = 0)
+    daily_file_sink(filename_t base_filename, int rotation_hour, int rotation_minute, bool truncate = false, uint16_t max_files = 0, const file_event_handlers_t& event_handlers = {})
```
gabime, line 71: cannot really move const ref.

example/example.cpp
```diff
@@ -26,9 +26,32 @@ void custom_flags_example();
 #include "spdlog/spdlog.h"
 #include "spdlog/cfg/env.h"  // support for loading levels from the environment variable
 #include "spdlog/fmt/ostr.h" // support for user defined types
+#include "spdlog/common.h"
+
+spdlog::file_event_handlers_t file_event_handlers;
```
gabime, line 31: no need to make it global

example/example.cpp
```diff
@@ -26,9 +26,32 @@ void custom_flags_example();
...
+
+void init_file_event_handler()
```
gabime, line 33: please move this towards the end of example.cpp

gabime (changes requested):

example/example.cpp
seker: fixed.

example/example.cpp
seker: file_event_handlers is referenced by other example functions.

include/spdlog/sinks/daily_file_sink.h
seker: fixed

include/spdlog/sinks/basic_file_sink-inl.h
seker: fixed

example/example.cpp
gabime: I prefer not to change the original example functions and keep it simple. Lets just have single function in the end that shows how to use it.

seker: ok.
I have changed that. Just pass file_event_handlers to the rotating_logger

gabime:
Merged. Thanks @seker

PR #2169, opened 2021-11-13
file_event_handlers add before_open function

seker:
file_event_handlers add before_open function

### DIFF
--- a/include/spdlog/details/file_helper-inl.h
+++ b/include/spdlog/details/file_helper-inl.h
@@ -37,6 +37,10 @@ SPDLOG_INLINE void file_helper::open(const filename_t &fname, bool truncate)
     auto *mode = SPDLOG_FILENAME_T("ab");
     auto *trunc_mode = SPDLOG_FILENAME_T("wb");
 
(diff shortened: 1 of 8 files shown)


gabime:
why do we need before open?

seker:
> why do we need before open?

in my project, i need to check avaliable disk spcace before open a log file while my app start up.

include/spdlog/details/file_helper-inl.h
```diff
@@ -54,6 +54,11 @@ SPDLOG_INLINE void file_helper::open(const filename_t &fname, bool truncate)
             }
             std::fclose(tmp);
         }
+        
+        if (event_handlers_.before_open)
```
gabime, line 58: I think its better called before the loop, otherwise it would be calked multiple times

gabime (changes requested):

include/spdlog/details/file_helper-inl.h
seker: god, my mistake~~

fixed.

PR #2172, opened 2021-11-14
example.cpp failes to build on FreeBSD

keith-dev:
We seem to be missing include file netinet/in.h in udp_client.h.

I've tested on Linux and FreeBSD, but don't have access to other platforms.


gabime:
Merged. Thanks @keith-dev

PR #2180, opened 2021-11-19
Avoid MSVC 2013 workaround used in gcc/clang

ibmibmibm:


gabime:
This code is just a copy from fmt lib repo. Please open this PR there (https://github.com/fmtlib/fmt).

PR #2182, opened 2021-11-19
Fix(tweakme): SPDLOG_FUNCTION

Light3039:
Uncommenting `SPDLOG_FUNCTION` will make MSVC fail to compile,

Fixed it by adding an `#ifdef` around `__PRETTY_FUNCTION__`


gabime:
Thanks @Light3039

PR #2187, opened 2021-11-23
Updated MIT License

joda01:
Updated to github MIT license to make it accessible by the REST API and project overview

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
 
-Copyright (c) 2016 Gabi Melman.                                       
+Copyright (c) 2021 Gabi Melman
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


gabime:
why did you remove the fmt note at the end ?

joda01:
@gabime Hello Gabi, when using a standard license from github it can automatically processed using the github REST API.
By the way, very cool library :)

gabime:
@vitaut Do you have objection of removing the note about fmt lib from of the spdlog license?
Please advise If I should put it somewhere else.

vitaut:
Since {fmt} is included in the repo you should specify that a different license applies to a portion of the code. I have no opinion regarding how it is done.

gabime:
@joda01 The license file seems to be the best place to put additional license info, so I will close this PR.

gabime:
On second thought we could add a license section to readme as recommended [here](https://softwareengineering.stackexchange.com/a/304895)

@joda01 Please add such section to the readme in this PR

PR #2195, opened 2021-12-01
Fix usage of ranges and to_hex in the same compile unit

patrickroocks:
Hello!

First of all, thanks a lot for this great package!

I encountered a compile problem when using ranges and bin_to_hex in the same compile unit. After doing some research I came to the conclusion that probably the only way so solve this is to hide the iterators used in bin_to_hex from the formatter in ranges. 

My change in detail:

When trying to use spdlog/fmt/bin_to_hex.h in the same compile unit as spdlog/fmt/bundled/ranges.h you got a compile error because there was a multiple definitions for iterable classes. This fix renames the begin() and end() getters in dump_info into getBegin()/getEnd() in order to avoid this collision.

Added an example of ranges in example.cpp to show that it actually works (a to_hex example was already there)

What do you think?

All the best
Patrick

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -14,6 +14,7 @@ void rotating_example();
 void daily_example();
 void async_example();
 void binary_example();
+void vector_example();
 void stopwatch_example();
 void trace_example();
 void multi_sink_example();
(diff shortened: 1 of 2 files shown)


gabime:
Thanks. Please use snake_case names (get_begin(), get_end())

patrickroocks:
fixed cody style

PR #2200, opened 2021-12-04
Implements shared library compatible global set_default_logger.

rioki:
The key difference is that registry::instance is moved to a cpp when
compiling the library.

THIS IS A BREAKING CHANGE IN BEHAVIOR! 
This PR was created to illustrate an approach. If this code is adapted, the rollout should be coordinated.


gabime:
Thanks. The  registry::instance is already contained in a cpp file here (https://github.com/gabime/spdlog/blob/4b3687f1a61912b778986cf5ac304e287aaebdd4/src/spdlog.cpp#L11) so what am I missing here?

rioki:
sorry I missed this dynamic. The key thing is that the inline needs to be removed, thus no copy is put into the client module.

gabime:
It is already removed in compiled lib configuration (https://github.com/gabime/spdlog/blob/d357a9e0e39263fa4acf634e74428e900ef4e9ae/include/spdlog/common.h#L38)

PR #2205, opened 2021-12-08
Fix warning for gcc, old style cast

Acretock:
In project i work on, we are warning free, and with our warning policy, we have this warning(even when muting 3rdparty):
`(long int)(local_year - gmt_year) * 365);` in include/details/os-inl.h


gabime:
Thanks @Acretock

PR #2206, opened 2021-12-08
Add custom on_stop callback to thread-pool, executed by each thread before joining them.

SBAM:
Hello Gabi,

A while ago, you accepted a pull request https://github.com/gabime/spdlog/pull/208 which was left behind when moving to the thread-pool based async logger.
If this is fine for you, please accept this request and I'll add some lines to the wiki.

Cheers,


gabime:
Thanks @SBAM

PR #2216, opened 2021-12-19
Reduce warnings with pedantic compiler `-Wuseless-cast`

vnepogodin:
on unix systems `getpid` returns `pid_t` which refers to `signed` type same as `int`

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -321,24 +321,15 @@ struct file_event_handlers
 
 namespace details {
 
-// to avoid useless casts (see https://github.com/nlohmann/json/issues/2893#issuecomment-889152324)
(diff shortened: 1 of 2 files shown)


gabime:
how can we be sure it is int in all non win32 platforms?

vnepogodin:
> how can we be sure it is int in all non win32 platforms?

Implement adaptive cast

gabime:
what do you mean?

vnepogodin:
I'll implement that though

vnepogodin:
Well I did implement it

vnepogodin:
also would be nice to add `-Wuseless-cast` for CI/CD with gcc compiler

gabime:
Thanks @vnepogodin

PR #2224, opened 2021-12-25
Exceptions for std::make_shared

AndrzejEndrju:
I would like a way to actually know without abort that logger failed to allocate for whatever reason


gabime:
The exceptions from logger/sink contructors are not cought by design.
You can catch the exception from the outside.

PR #2228, opened 2021-12-30
spdlog: fmt - support `std::span` in `to_hex`

timblechmann:
`std::span` does not have `const_iterator`. this prevents `to_hex` from
being used with `std::span<>`. to fix this, we provide an explicit
overload.

compare: https://cplusplus.github.io/LWG/issue3320


gabime:
Thanks @timblechmann

PR #2234, opened 2022-01-04
Reset current size if rotated files on open

SpriteOvO:
If rotated files on open, `file_helper_` will open an empty file and `current_size_` should be reset to 0.

This PR fixed the bug.


gabime:
Typhanks @SpriteOvO . Merged.

PR #2245, opened 2022-01-11
Allow forward-declaration of level_enum

daverigby:
spdlog::level::level_enum cannot be forward-declared at present, as
the definition does not specify an underlying type.

To allow users to make use of <spdlog/fwd.h> to refer to
level::level_enum without pulling in all of <spdlog/common.h> (which
can be quite costly), specify an underlying type (int) for
level::level_enum, then add a forward-declaration for it to
spdlog/fwd.h.

Note this required explicitly casting level_enum to size_t within ansicolor_sink due to sign-conversion errors:

    implicit conversion changes signedness: 'const level::level_enum' to 'std::__1::array::size_type' (aka 'unsigned long') [-Wsign-conversion]

It would appear that an enum with an unspecified underlying type is in
some kind of superposition - it can be treated as both signed _and_
unsigned - using an underlying type of 'unsigned int' triggers even
more warnings of this kind...

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -208,7 +208,7 @@ using level_t = std::atomic<int>;
 
 // Log level enum
 namespace level {
-enum level_enum
+enum level_enum : int
 {
     trace = SPDLOG_LEVEL_TRACE,
     debug = SPDLOG_LEVEL_DEBUG,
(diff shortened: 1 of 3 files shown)


gabime:
Thanks @daverigby

Issue #1607, opened 2020-07-07
async logger handle unexpected termination


TamasFlorin:
Hey,

First of all, thank your for all the work that you have put into this! 

I was wondering what is the right way of handling the following situation: we are using the async logger and the process suddenly crashes without having the chance to flush the log messages to the file.

My first thought was to install a signal handler for `SIGSEGV `and `SIGABRT` and call `spdlog::shutdown()` from the respective handler. However, the `shutdown `method does not seem to be async-safe since it uses `std::mutex`, meaning that this could cause deadlocks when the process crashes and that might be even worse than losing a few logs.

Are there any other suggestions that could help us get around this issue?

Thanks

tt4g:
There are not many safe operations within the signal handler (see reference: [std::signal \- cppreference\.com](https://en.cppreference.com/w/cpp/utility/program/signal)).
And spdlog does not support its use within signal handlers.

One workaround is to use `std::condition_variable` to call a function that is not signal safe outside the signal function.
Like [this link](https://stackoverflow.com/a/52424515) example code, if you call the notification function of `std::condition_variable` in the signal function and call `spdlog::shutdown()` when the process thread receives the notification,
spdlog will now shutdown when a signal is received.

gabime:
> First of all, thank your for all the work that you have put into this!

Thanks for the feedback :)

I will add it to the todo list to implement in spdlog somehow (probably similar to @tt4g suggestion).

TamasFlorin:
> > First of all, thank your for all the work that you have put into this!
> 
> Thanks for the feedback :)
> 
> I will add it to the todo list to implement in spdlog somehow (probably similar to @tt4g suggestion).

Thank you! Looking forward to it!

TamasFlorin:
> There are not many safe operations within the signal handler (see reference: [std::signal - cppreference.com](https://en.cppreference.com/w/cpp/utility/program/signal)).
> And spdlog does not support its use within signal handlers.
> 
> One workaround is to use `std::condition_variable` to call a function that is not signal safe outside the signal function.
> Like [this link](https://stackoverflow.com/a/52424515) example code, if you call the notification function of `std::condition_variable` in the signal function and call `spdlog::shutdown()` when the process thread receives the notification,
> spdlog will now shutdown when a signal is received.

Thanks for the suggestion! I will look into it.

qaler:
Could you please provide an example code snippet of how to implement logs flushing when segfault occurs?

[...]

PR #2246, opened 2022-01-14
Now only getting time if pattern_formatter needs it

doug1234:
The pattern_formatter class was making unwanted calls to localtime_r or similar functions even when it didn't use the time information.  A flag was added which disables the calls when the pattern formatter is not going to use the information.

### DIFF
--- a/include/spdlog/pattern_formatter-inl.h
+++ b/include/spdlog/pattern_formatter-inl.h
@@ -1021,6 +1021,7 @@ SPDLOG_INLINE pattern_formatter::pattern_formatter(
     , pattern_time_type_(time_type)
     , last_log_secs_(0)
     , custom_handlers_(std::move(custom_user_flags))
+    , needs_time_(false)
 {
     std::memset(&cached_tm_, 0, sizeof(cached_tm_));
     compile_pattern_(pattern_);
@@ -1032,6 +1033,7 @@ SPDLOG_INLINE pattern_formatter::pattern_formatter(pattern_time_type time_type, 
     , eol_(std::move(eol))
     , pattern_time_type_(time_type)
     , last_log_secs_(0)
+    , needs_time_(false)
 {
     std::memset(&cached_tm_, 0, sizeof(cached_tm_));
     formatters_.push_back(details::make_unique<details::full_formatter>(details::padding_info{}));
@@ -1049,11 +1051,13 @@ SPDLOG_INLINE std::unique_ptr<formatter> pattern_formatter::clone() const
 
 SPDLOG_INLINE void pattern_formatter::format(const details::log_msg &msg, memory_buf_t &dest)
 {
-    auto secs = std::chrono::duration_cast<std::chrono::seconds>(msg.time.time_since_epoch());
-    if (secs != last_log_secs_)
-    {
-        cached_tm_ = get_time_(msg);
-        last_log_secs_ = secs;
+    if (needs_time_) {
+        const auto secs = std::chrono::duration_cast<std::chrono::seconds>(msg.time.time_since_epoch());
+        if (secs != last_log_secs_)
+        {
+            cached_tm_ = get_time_(msg);
+            last_log_secs_ = secs;
+        }
     }
 
     for (auto &f : formatters_)
@@ -1097,6 +1101,7 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
     {
     case ('+'): // default formatter
         formatters_.push_back(details::make_unique<details::full_formatter>(padding));
+        needs_time_ = true;
         break;
 
     case 'n': // logger name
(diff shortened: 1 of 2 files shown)


gabime:
Thanks. Not sure if it worth it considering the fact that vast majority of use cases need time and that it is cached each second. Whats the use case that make it worth the extra complxity?
Also this doesn't pass the tests (the second patter_formatter overload, the one without te pattern, needs to set the flag to true since it uses the default formatter).

doug1234:
Fixed the incorrect initialization.  Sorry about that.

Thanks for looking over the pull request.  I actually didn't add it for efficiency reasons.  I was making some spdlog calls between a fork() and execv().  There was a random crash that was in localtime_r().  It seems that localtime_r() needs to do some stuff with environment variables and likely that wasn't a safe time to check environment variables.  So I modified my pattern so as to not have time and was shocked to find it randomly crashing in the same place!  After my change the random crashes went away.

doug1234:
If you take a look at the fork() man page (https://man7.org/linux/man-pages/man2/fork.2.html), calling a lot of functions is unsafe between fork() and execv() which would seem to include localtime_r(). However the fork() man page is not as explicit as I would like (localtime_r isn't signal safe so it is not guaranteed to be safe to call between fork() and execv() but that doesn't mean it will always be a problem either).  You might be able to reproduce the issue by logging between a fork() and execv() but I only saw the crash less than 1 in 10 times that my software ran.  However, the fork() man page is enough to convince me that its not a good idea to call localtime_r() between fork() and execv() and removing the call fixed my issue.

gabime:
> After a fork() in a multithreaded program, the child can
          safely call only async-signal-safe functions (see
          signal-safety(7)) until such time as it calls execve(2).

spdlog is not async-signal-safe (see #1607), so the crash can be from anywhere, not just localtime.

doug1234:
While I agree that spdlog can't make the async-signal-safe guarantee, spdlog is close enough to it (when used carefully) for my purposes minus the spurious call to localtime.  Overall I don't see a down side to this change other than a small increase in complexity and it should improve efficiency in a corner case.

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1021,6 +1021,7 @@ SPDLOG_INLINE pattern_formatter::pattern_formatter(
     , pattern_time_type_(time_type)
     , last_log_secs_(0)
     , custom_handlers_(std::move(custom_user_flags))
+    , needs_time_(false)
```
gabime, line 1024: please rename to “need_localtime_”
Some of the flags needs time but not localtime.

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1121,101 +1127,125 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
...
         break;
 
     case ('S'): // seconds
         formatters_.push_back(details::make_unique<details::S_formatter<Padder>>(padding));
+        needs_time_ = true;
         break;
 
     case ('e'): // milliseconds
         formatters_.push_back(details::make_unique<details::e_formatter<Padder>>(padding));
+        needs_time_ = true;
```
gabime, line 1202: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1121,101 +1127,125 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
...
         break;
 
     case ('F'): // nanoseconds
         formatters_.push_back(details::make_unique<details::F_formatter<Padder>>(padding));
+        needs_time_ = true;
         break;
 
     case ('E'): // seconds since epoch
         formatters_.push_back(details::make_unique<details::E_formatter<Padder>>(padding));
+        needs_time_ = true;
```
gabime, line 1217: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1121,101 +1127,125 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
...
         break;
 
     case ('e'): // milliseconds
         formatters_.push_back(details::make_unique<details::e_formatter<Padder>>(padding));
+        needs_time_ = true;
         break;
 
     case ('f'): // microseconds
         formatters_.push_back(details::make_unique<details::f_formatter<Padder>>(padding));
+        needs_time_ = true;
```
gabime, line 1207: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1121,101 +1127,125 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
...
         break;
 
     case ('F'): // nanoseconds
         formatters_.push_back(details::make_unique<details::F_formatter<Padder>>(padding));
+        needs_time_ = true;
```
gabime, line 1212: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1121,101 +1127,125 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
...
         break;
 
     case ('z'): // timezone
         formatters_.push_back(details::make_unique<details::z_formatter<Padder>>(padding));
+        needs_time_ = true;
         break;
 
     case ('P'): // pid
         formatters_.push_back(details::make_unique<details::pid_formatter<Padder>>(padding));
+        needs_time_ = true;
```
gabime, line 1248: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1252,18 +1282,22 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
 
     case ('u'): // elapsed time since last log message in nanos
         formatters_.push_back(details::make_unique<details::elapsed_formatter<Padder, std::chrono::nanoseconds>>(padding));
+        needs_time_ = true;
```
gabime, line 1285: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1252,18 +1282,22 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
...
         break;
 
     case ('i'): // elapsed time since last log message in micros
         formatters_.push_back(details::make_unique<details::elapsed_formatter<Padder, std::chrono::microseconds>>(padding));
+        needs_time_ = true;
```
gabime, line 1290: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1252,18 +1282,22 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
...
         break;
 
     case ('o'): // elapsed time since last log message in millis
         formatters_.push_back(details::make_unique<details::elapsed_formatter<Padder, std::chrono::milliseconds>>(padding));
+        needs_time_ = true;
```
gabime, line 1295: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1252,18 +1282,22 @@ SPDLOG_INLINE void pattern_formatter::handle_flag_(char flag, details::padding_i
...
         break;
 
     case ('O'): // elapsed time since last log message in seconds
         formatters_.push_back(details::make_unique<details::elapsed_formatter<Padder, std::chrono::seconds>>(padding));
+        needs_time_ = true;
```
gabime, line 1300: not needed

include/spdlog/pattern_formatter-inl.h
```diff
@@ -1049,11 +1051,13 @@ SPDLOG_INLINE std::unique_ptr<formatter> pattern_formatter::clone() const
 
 SPDLOG_INLINE void pattern_formatter::format(const details::log_msg &msg, memory_buf_t &dest)
 {
-    auto secs = std::chrono::duration_cast<std::chrono::seconds>(msg.time.time_since_epoch());
-    if (secs != last_log_secs_)
-    {
-        cached_tm_ = get_time_(msg);
-        last_log_secs_ = secs;
+    if (needs_time_) {
```
gabime, line 1054: Please put the open brace in a new line like rest of the codebase

gabime (changes requested):

gabime:
Merged. Thanks @doug1234

PR #2256, opened 2022-01-24
optimized jetbrains logo SVG

RiedleroD:
idk where you got it from but sheesh those were some horrid paths. Except for the text, that was fine.
I also removed unnecessary groups, shortened the gradient ids and moved the attribute order to be more readable.


gabime:
I downloaded it from the official jetbrain site. I am not sure if it is allowed to modify. In any case I plan to remove this logo from the readme page soon. So thanks, but no need for now.

PR #2258, opened 2022-01-25
Added support for MSBuild native builds

alec500oo:
Including msvc/spdlog.vcxproj allows library users to natively include
spdlog in their Visual Studio projects. This is especially useful on
systems that do not include CMake.


gabime:
Thanks, however I don't want to support another build tree. Note that cmake also offers various spdlog options

PR #2265, opened 2022-02-04
Support conversion from std::source_location to spdlog::source_loc

sylveon:
Enables better interop between spdlog and code using `std::source_location`

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -16,6 +16,17 @@
 #include <functional>
 #include <cstdio>
 
+#ifdef __has_include
+#if __has_include(<version>)
+#include <version>
+#endif
+#endif
+
+#if __cpp_lib_source_location >= 201907L
+#include <source_location>
+#define SPDLOG_HAS_STD_SOURCE_LOC
+#endif
+
 #ifdef SPDLOG_USE_STD_FORMAT
 #    include <string_view>
 #endif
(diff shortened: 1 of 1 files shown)


gabime:
Thanks! I am not sure I want to add yet another include if not really needed. Whats the usecase here ?

sylveon:
It helps interop with `std::source_location`, so you don't have to convert between `spdlog::source_loc` and it all the time.

I tried looking into enabling to swap `spdlog::source_loc` for `std::source_location` entirely but ran into the issue that while `std::source_location` can be default constructed, it initializes to an unspecified value so we have no way to distinguish an empty `std::source_location` from a one that came from `std::source_location::current()`. We would need to wrap it in something like `std::optional<std::source_location>` for it to work with the current code.

The `<source_location>` header is very small in any case: the only thing it does is `#include <cstdint>` and then declare the `std::source_location` class (which is less than 60 lines).

gabime:
source_loc is mainly used by the lib code not by users, so this ability to convert seems to me a rare thing to need.

sylveon:
I have my own win32 error handling helpers wrapping spdlog, so there is definitely user code that would be interested in using source_loc. But that's fair enough, manually doing the conversion isn't the end of the world.

gabime:
Closing for now. Thanks anyway !

PR #2266, opened 2022-02-06
Improved Readability Using Digit Separator

fynnwilliam:
Using C++14 digit separator to enhance code readability.


gabime:
Thanks. However this is c++14 feature while spdlog is c++11.

fynnwilliam:
Welcome and sorry about that.

PR #2269, opened 2022-02-08
Fix issue #2201

### LINKED ISSUE #2201: Pattern width is not applied for missing source information
If the width is set in the format pattern for one of the source information fields (for example `%10s`), this width is only applied for messages which contain the source information. For example, this code
```cpp
#include <spdlog/spdlog.h>

int main(int argc, char *argv[]) {
  spdlog::set_pattern("(%10s:%10#:%10!) %v");

  spdlog::info("Hello World");
  SPDLOG_INFO("Hello World");
}
```
will produce the output
```
$ ./a.out
(::) Hello World
(  main.cpp:         7:      main) Hello World
```
while I would expect it to produce output similar to this, to respect the alignment specified, even if the fields are not available.
```
$ ./a.out
(          :          :          ) Hello World
(  main.cpp:         7:      main) Hello World
```

kyuheon-kr:
Fix #2201  

Add ScopedPadder if the source information is empty while formatting source information fields

### DIFF
--- a/include/spdlog/pattern_formatter-inl.h
+++ b/include/spdlog/pattern_formatter-inl.h
@@ -766,6 +766,7 @@ public:
     {
         if (msg.source.empty())
         {
+            ScopedPadder p(0, padinfo_, dest);
             return;
         }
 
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @kyuheon-kr

PR #2273, opened 2022-02-14
Limit max number of rotating files to 200000, not max size

surfycui:
Should limit max number of rotating files to 200000, not max size


gabime:
Thanks!

PR #2281, opened 2022-02-15
Add get_max_size() and set_max_size() methods to rotating_file_sink class

AenBleidd:
Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

### DIFF
--- a/include/spdlog/sinks/rotating_file_sink-inl.h
+++ b/include/spdlog/sinks/rotating_file_sink-inl.h
@@ -35,7 +35,7 @@ SPDLOG_INLINE rotating_file_sink<Mutex>::rotating_file_sink(
     {
         throw_spdlog_ex("rotating sink constructor: max_size arg cannot be zero");
     }
-    
+
     if (max_files > 200000)
     {
         throw_spdlog_ex("rotating sink constructor: max_files arg cannot exceed 200000");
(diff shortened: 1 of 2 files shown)


gabime:
Thanks for the PR.

This sink is designed to be initialized once on construction. Why is it needed?


Also, those functions need to be locked using the base `mutex_`:

```c++
std::lock_guard<Mutex> lock(base_sink<Mutex>::mutex_);
```

AenBleidd:
> This sink is designed to be initialized once on construction. Why is it needed?

Sometimes it's very convenient to change file max_size 'on the fly', especially after changing 'log_level'.
This is mostly 'nice-to-have' to get rid of the workaround to reinitialize logger every time file 'max_size' need to be changed.

> Also, those functions need to be locked using the base `mutex_`

Fixed

gabime:
This is rare need and I prefer not to add more code to this already complicated sink. I will reconsider if there is more demand for it.
Thanks anyway !

artemyv:
We use this functionality when changing logs from regular to trace which uses a lot more space.
Before adding this feature to our private version we tried replacing the sink in the loggers - but this approach looses the configured global pattern, since it is not applied to new sink (and when applied skips sinks not used by any logger). So we opted to add the set_max_size()

Sofapriester:
Also in my company we would benefit from such a setting. Same problem like @artemyv described. If a user decides to switch to huge debug logs, then we also would have to make the log size bigger. Otherwise the debug logs would be more or lese senseless because always overwritten by log rotation. Currently we workaround this by re-creating the sink. But this only because such a setting was missing.

**We would vote for adding set_max_size()**

PR #2305, opened 2022-03-11
Fix fopen_s error reporting with PREVENT_CHILD_FD

nUl1:
`open()` failure was incorrectly reported as a success under *nix with `SPDLOG_PREVENT_CHILD_FD` enabled, leaving `*fp` untouched.


gabime:
Thanks @nUl1

PR #2315, opened 2022-03-18
Added qt_sink_color class

mguludag:
Hello, I added qt_sink_color_mt/st that uses bundled fmtlib's color and text style syntax, also implemented set_color function for we can able to change color or style later.
set_color function example:
`set_color(spdlog::level::err, fmt::fg(fmt::color::white) | fmt::bg(fmt::color::red) | fmt::emphasis::bold | fmt::emphasis::italic);`


gabime:
Thanks. Since spdlog now can be compiled without fmt ( `#define SPDLOG_USE_STD_FORMAT` ) this is a little problematic to rely on fmt color. Also, is some installations an external fmt is used and the fmt/bundled/color.h doesnt exist at all.

mguludag:
> Thanks. Since spdlog now can be compiled without fmt ( `#define SPDLOG_USE_STD_FORMAT` ) this is a little problematic to rely on fmt color. Also, is some installations an external fmt is used and the fmt/bundled/color.h doesnt exist at all.

I forgot this scenarios, I added guards for this

include/spdlog/sinks/qt_sinks.h
```diff
@@ -4,46 +4,55 @@
...
 // qt_sink class
 //
 namespace spdlog {
 namespace sinks {
-template<typename Mutex>
-class qt_sink : public base_sink<Mutex>
-{
+#if !defined(SPDLOG_USE_STD_FORMAT)
+template <typename Mutex>
+class qt_sink_color;
```
gabime, line 34: why the forward declaration is needed?

include/spdlog/sinks/qt_sinks.h
```diff
@@ -53,50 +62,184 @@ class qt_sink : public base_sink<Mutex>
     std::string meta_method_;
 };
 
+#if !defined(SPDLOG_USE_STD_FORMAT)
+template <typename Mutex>
+class qt_sink_color : public qt_sink<Mutex> {
```
gabime, line 67: I don't see much benefit in inherit from qt_sink here.

include/spdlog/sinks/qt_sinks.h
```diff
@@ -53,50 +62,184 @@ class qt_sink : public base_sink<Mutex>
...
+        auto bg = fmt::detail::make_background_color<fmt::detail::char8_type>(color.get_background());
+        auto fg = fmt::detail::make_foreground_color<fmt::detail::char8_type>(color.get_foreground());
+        auto em = fmt::detail::make_emphasis<fmt::detail::char8_type>(color.get_emphasis());
+
+        str.append(bg.begin(), bg.end());
+        str.append(fg.begin(), fg.end());
+        str.append(em.begin(), em.end());
+
+        colors_[color_level] = str;
+    }
```
gabime, line 99: str can be moved here to avoid the copy

include/spdlog/sinks/qt_sinks.h
```diff
@@ -4,46 +4,55 @@
...
-class qt_sink : public base_sink<Mutex>
-{
+#if !defined(SPDLOG_USE_STD_FORMAT)
+template <typename Mutex>
+class qt_sink_color;
+#endif
+template <typename Mutex>
+class qt_sink : public base_sink<Mutex> {
+#if !defined(SPDLOG_USE_STD_FORMAT)
+    friend class qt_sink_color<Mutex>;
```
gabime, line 39: why friend not needed?

include/spdlog/sinks/qt_sinks.h
```diff
@@ -53,50 +62,184 @@ class qt_sink : public base_sink<Mutex>
...
 
+#if !defined(SPDLOG_USE_STD_FORMAT)
+template <typename Mutex>
+class qt_sink_color : public qt_sink<Mutex> {
+public:
+    qt_sink_color(QObject *qt_object = nullptr, bool color_ = true, const std::string &meta_method = "")
+    {
+        qt_sink<Mutex>::qt_object_ = qt_object;
+        qt_sink<Mutex>::meta_method_ = meta_method;
+        should_do_colors_ = color_;
```
gabime, line 73: Whats the point of shoild_do_colors? If colors not needed this sinks should not be used anyway

include/spdlog/sinks/qt_sinks.h
```diff
@@ -53,50 +62,184 @@ class qt_sink : public base_sink<Mutex>
...
+    {
+        msg.color_range_start = 0;
+        msg.color_range_end = 0;
+        memory_buf_t formatted;
+        base_sink<Mutex>::formatter_->format(msg, formatted);
+
+        if (should_do_colors_ && msg.color_range_end > msg.color_range_start)
+        {
+            // before color range
+            std::string str;
```
gabime, line 123: str could be made a member and reused to avoid allocations

include/spdlog/sinks/qt_sinks.h
```diff
@@ -53,50 +62,184 @@ class qt_sink : public base_sink<Mutex>
...
+            auto ccode = to_ccode_(colors_[msg.level]);
+            auto reset = to_ccode_(reset_);
+            str.reserve(formatted.size() + ccode.size());
```
gabime, line 126: reserve should take into account the reset size as well I think.

include/spdlog/sinks/qt_sinks.h
```diff
@@ -53,50 +62,184 @@ class qt_sink : public base_sink<Mutex>
...
-{
+template <typename Factory = spdlog::synchronous_factory>
+inline std::shared_ptr<logger>
+qt_logger_st(const std::string &logger_name, QObject* qt_object, const std::string &meta_method) {
     return Factory::template create<sinks::qt_sink_st>(logger_name, qt_object, meta_method);
 }
+
+//
+// Factory functions (color)
+//
```
gabime, line 207: I dont like all those factory functions. too much code. They exist only in the most popular sinks

gabime (changes requested):

include/spdlog/sinks/qt_sinks.h
mguludag: yes, not too much now I changed to base base_sink

include/spdlog/sinks/qt_sinks.h
mguludag: Ok, added move semantics

include/spdlog/sinks/qt_sinks.h
mguludag: I want to access qt_sink's priv members from qt_sink_color class

include/spdlog/sinks/qt_sinks.h
mguludag: Ok, I removed this

gabime:
@mguludag How do the colors work with qt? Do the qt widgets understand ansi colors codes?

gabime:
Not sure I understand. In this pr you are using fmt::color. How does it work with qt?

mguludag:
> Not sure I understand. In this pr you are using fmt::color. How does it work with qt?

fmt::color turns into ansi color codes in QString then it goes to slot that includes ansicolor parser, it parse ansicolor codes and convert into QColor then push into FormattedText object with corresponding text part. setting TextCursor's format coming from this FormattedText object and inserting the text from FormattedText with this format into the widgets.

```C++
void AnsiEscapeCodeHandler::appendAnsiText(const QString &text)
{
    auto const formattedText = parseText(FormattedText(text, m_textCursor.charFormat())); // returns QList<FormattedText>

    for (auto const& text_ : qAsConst(formattedText)) {
        m_textCursor.movePosition(QTextCursor::End);
        m_textCursor.setCharFormat(text_.format);
        m_textCursor.insertText(text_.text);
        m_textCursor.movePosition(QTextCursor::End);
    }

}
```

image

gabime:
So it only usable with your FormattedText code? I suggest adding this to your repo instead. There is no point adding to this repo.

mguludag:
> So it only usable with your FormattedText code? I suggest adding this to your repo instead. There is no point adding to this repo.

Unfortunately Qt doesn't show colored text from ansicolors out of the box. Users have to write parser their own or use prewritten codes like qt-creator's source code.

PR #2317, opened 2022-03-22
Fixed compiler error when building on Windows with #define UNICODE

risa2000:
The original `InetPton` expands to `InetPtonW` when building with UNICODE defined and expects the string parameter to be wchar_t. On the other hand macro `TEXT()` just adds prefix `L` to a string literal (just making it wchar_t literal). The proper way here would be converting `host.c_str()` result from UTF-8(?) into wchar_t (UNICODE) string, but this seems to be an overkill since the host is typically an IP address or a host/domain name. So assuming an ASCII input should be reasonably safe.


gabime:
Thanks @risa2000

PR #2324, opened 2022-03-27
Add option to enable formatting of systemd sink

### LINKED ISSUE #2320: Add option to use formatter inside systemd sink
Hi.

We are migrating from `file_sink` to `systemd_sink` and we would like to preserve the `spdlog` format. It's dropped (by design, I understand) during conversion because `journald` implements its own structured logging. That makes sense, but we're losing some important fields in the process.

Basically, I'm looking #1289 but using the alternative approach proposed.
That is, adding a boolean to constructor and re-implement #1107 for `systemd_sink`.

I understand that I can create my own sink to implement it this way, but I having this feature upstream is of course preferable (avoid duplication and guarantees further updates).

I can definitely open a Pull Request if you think this is an acceptable addition.

I have also just one question: `sd_journal_send()` being thread safe, does that mean I can safely use `systemd_sink_st` in a multi-threaded environment with multiple loggers? Is there any reason to prefer `systemd_sink_mt`?

Delgan:
Fixes #2320.

### DIFF
--- a/include/spdlog/sinks/systemd_sink.h
+++ b/include/spdlog/sinks/systemd_sink.h
@@ -18,16 +18,14 @@ namespace sinks {
 
 /**
  * Sink that write to systemd journal using the `sd_journal_send()` library call.
- *
- * Locking is not needed, as `sd_journal_send()` itself is thread-safe.
  */
 template<typename Mutex>
 class systemd_sink : public base_sink<Mutex>
 {
 public:
-    //
-    systemd_sink()
-        : syslog_levels_{{/* spdlog::level::trace      */ LOG_DEBUG,
+    explicit systemd_sink(bool enable_formatting)
+        : enable_formatting_{enable_formatting}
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/systemd_sink.h
```diff
@@ -18,16 +18,14 @@ namespace sinks {
...
- * Locking is not needed, as `sd_journal_send()` itself is thread-safe.
  */
 template<typename Mutex>
 class systemd_sink : public base_sink<Mutex>
 {
 public:
-    //
-    systemd_sink()
-        : syslog_levels_{{/* spdlog::level::trace      */ LOG_DEBUG,
+    explicit systemd_sink(bool enable_formatting)
```
gabime, line 26: This breaks existing API. Please add default value of false or a new ctor

gabime:
@Delgan Any news about this pr requested changes?

Delgan:
@gabime Sorry for the delay, I updated the PR. :+1:

gabime:
Thanks @Delgan

PR #2326, opened 2022-03-29
Add new rotating file sink with different naming formation

msarmento1:
This is a proposal for current rotating file sink.
In this sink, files will be renamed appending the date in which they
were rotated, having so there is no circular file limit.
There is also a protection against multiple files rotating at the same time,
by appending ".{index}"

I haven't come up with a better name for this sink, sorry.

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -11,6 +11,7 @@ void load_levels_example();
 void stdout_logger_example();
 void basic_example();
 void rotating_example();
+void rotating2_example();
 void daily_example();
 void async_example();
 void binary_example();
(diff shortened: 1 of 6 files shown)


gabime:
Thanks. How is it better than creating rotating sink with max files = maxint?

msarmento1:
In some situations I think it's more informative to have the date in which it was rotated than simply a ".{index}". I did it mainly because I need this feature where I work, but if you think it doesn't fit this lib, I'll restrain it to my lib only

gabime:
Thansk for clarifying. Closing for now as it seems a very specific need.

PR #2328, opened 2022-04-01
Add optional "ident" argument to systemd sink constructor

### LINKED ISSUE #2323: Add option to customize/disable "SYSLOG_IDENTIFIER" in systemd sink?
Hi again. :)

I'm still exploring the `systemd_sink` and I'm a little bothered by the use of `logger_name` to complete the `SYSLOG_IDENTIFIER` field.
It was introduced by #1289 to ensure logger's name isn't lost. This is absolutely fine, but this isn't useful when `enable_formatter` (#2320) is enabled.
By default, `SYSLOG_IDENTIFIER` contains the application name which seems more natural. Using the logger's name as `SYSLOG_IDENTIFIER` makes it kind if harder to identify the application logs within `journalctl`.
My main "problem" is that by overriding  `SYSLOG_IDENTIFIER` we can no longer filter logs with `journalctl -t appname`, we have to use `journalctl -f _COMM=appname`. It's a little annoyance, but that makes me feel `SYSLOG_IDENTIFIER` isn't used properly.

For reference, here is a message running `./my_app` without `SYSLOG_IDENTIFIER` being overridden:

```json
{
	"CODE_FILE": "main.cpp",
	"CODE_FUNC": "main",
	"CODE_LINE": "6",
	"MESSAGE": "Hello World!",
	"PRIORITY": "6",
	"SYSLOG_IDENTIFIER": "my_app",
	"_BOOT_ID": "5a72020658b14867a5f011f283647138",
	"_GID": "1000",
	"_HOSTNAME": "delgan-laptop",
	"_MACHINE_ID": "3fcc29c834cf4bdeb22dfd1340a789f6",
	"_PID": "393259",
	"_SOURCE_REALTIME_TIMESTAMP": "1648374364807251",
	"_TRANSPORT": "journal",
	"_UID": "1000",
	"__CURSOR": "s=cf394ceea1a04b3bbff567538c6a8070;i=3c72a;b=5a72020658b14867a5f011f283647138;m=1038984c36;t=5db301078a0a5;x=c9adb40b0832588b",
	"__MONOTONIC_TIMESTAMP": "69668981814",
	"__REALTIME_TIMESTAMP": "1648374364807333"
}
```

And here is the same message with `SYSLOG_IDENTIFIER=my_logger`:

```json
{
	"CODE_FILE": "main.cpp",
	"CODE_FUNC": "main",
	"CODE_LINE": "6",
	"MESSAGE": "Hello World!",
	"PRIORITY": "6",
	"SYSLOG_IDENTIFIER": "my_logger",
	"_BOOT_ID": "5a72020658b14867a5f011f283647138",
	"_COMM": "my_app",
	"_GID": "1000",
	"_HOSTNAME": "delgan-laptop",
	"_MACHINE_ID": "3fcc29c834cf4bdeb22dfd1340a789f6",
	"_PID": "393265",
	"_SOURCE_REALTIME_TIMESTAMP": "1648374366820985",
	"_TRANSPORT": "journal",
	"_UID": "1000",
	"__CURSOR": "s=cf394ceea1a04b3bbff567538c6a8070;i=3c72c;b=5a72020658b14867a5f011f283647138;m=1038b7065a;t=5db3010975aca;x=9e960f3f99fad8a6",
	"__MONOTONIC_TIMESTAMP": "69670995546",
	"__REALTIME_TIMESTAMP": "1648374366821066"
}
```

I have been thinking of the following possibilities:
- Add a new `bool` argument to `systemd_sink` to disable `SYSLOG_IDENTIFIER` overriding. The implementation may require adding nested `if` and repeating `sd_journal_send` calls which isn't very elegant.
- Add a new `std:string ident` argument similar to `syslog_sink`. If it's not empty, use it to override `SYSLOG_IDENTIFIER` otherwise use logger's name.
- Do nothing and keep it as it is. I'm still in an exploratory phase and I'm not sure I'll use `systemd_sink` in the end, so if you want to avoid adding complexity, that's perfectly acceptable and I'll find another solution anyway.

Let me know what's your opinion on this, and if you think it's worth implementing something, I'll open a PR. :+1:

Delgan:
Fixes #2323.

### DIFF
--- a/include/spdlog/sinks/systemd_sink.h
+++ b/include/spdlog/sinks/systemd_sink.h
@@ -23,8 +23,9 @@ template<typename Mutex>
 class systemd_sink : public base_sink<Mutex>
 {
 public:
-    explicit systemd_sink(bool enable_formatting = false)
-        : enable_formatting_{enable_formatting}
+    systemd_sink(std::string ident = "", bool enable_formatting = false)
+        : ident_{std::move(ident)}
+        , enable_formatting_{enable_formatting}
         , syslog_levels_{{/* spdlog::level::trace      */ LOG_DEBUG,
               /* spdlog::level::debug      */ LOG_DEBUG,
               /* spdlog::level::info       */ LOG_INFO,
@@ -40,6 +41,7 @@ public:
     systemd_sink &operator=(const systemd_sink &) = delete;
 
 protected:
+    const std::string ident_;
     bool enable_formatting_ = false;
     using levels_array = std::array<int, 7>;
     levels_array syslog_levels_;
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @Delgan . Merged.

Issue #1751, opened 2020-11-30
Gibberish stdout of UTF-8 characters in non-unicode codepage


yetanothercheer:
https://github.com/gabime/spdlog/blob/233e97c5e46d8f8e83a6ea06b2dc8db804a4aed4/include/spdlog/sinks/wincolor_sink-inl.h#L138

AFAIK, this line will not work as expected on UTF-8 string in locale codepage. A workaround is
```
 SetConsoleOutputCP(65001);
```

So, is this a bug or a feature?

tt4g:
IMO.

There is no specification for character encoding in C++.
Therefore, it is the responsibility of the programmer to coordinate the character encoding inside C++ with the console encoding.

yetanothercheer:
Thanks for your reply!

BTW, I notice fmt use 
https://github.com/gabime/spdlog/blob/233e97c5e46d8f8e83a6ea06b2dc8db804a4aed4/include/spdlog/fmt/bundled/core.h#L321-L326
to determine the encoding of source files (I am not quite sure, though), and use this 
https://github.com/gabime/spdlog/blob/cd701761f9ad9919b9409e519237fb4c90f6375d/include/spdlog/fmt/bundled/format-inl.h#L1419-L1425
to print correctly in a locale codepage.

Can we do the same thing to make programmer's life easier?

tt4g:
For now there is only support for converting UTF-16 (`wchar_t`) to UTF-8 (`char`).

If you need that feature right now, you might want to create a custom sync.
A helpful implementation is `spdlog::sinks::wincolor_sink`. This is the only one that uses the `WriteConsoleA()` API.

https://github.com/gabime/spdlog/blob/cbe9448650176797739dbab13961ef4c07f4290f/include/spdlog/sinks/wincolor_sink-inl.h#L134-L139

Issue #762, opened 2018-07-21
UTF-8 output on Windows?


YukinoHayakawa:
wincolor_sink uses WriteConsoleA to write output, which AFAIK can't handle UTF-8. Should we convert the buffer to UTF-16 and use WriteConsoleW instead?

gabime:
Maybe.. spdlog never tried to deal with  wchars in the windows console.

YukinoHayakawa:
I'm thinking of of whether this change will break existing code. Maybe add another sink implement (like wincolor_utf8_sink) would be better?

gabime:
There is already an optional  ```SPDLOG_WCHAR_TO_UTF8_SUPPORT``` macro.
So maybe a workaround would be:
```c++
#ifdef SPDLOG_WCHAR_TO_UTF8_SUPPORT
WriteConsoleW(..)
#else
WriteConsoleA(..)
#endif

**Comment by YukinoHayakawa** [comment]
I'm afraid they are not dealing with exactly the same thing. I already have UTF-8 `char`-based strings as params & format string. I only hope they can reach the terminal without being treated badly.

**Comment by gabime** [comment]
Yes but still, WriteConsoleW should be used instead.

**Comment by YukinoHayakawa** [comment]
Yes.

**Comment by polesapart** [comment]
WriteConsoleA can be used to handle utf8 just fine. In fact, if your application doesn't use wide strings at all and you decide to use WriteConsoleW, you'd first need to copy-convert all your console-targeted strings to widechar, which can have a huge performance hit, specially if it involves allocation/deallocation.

What you'd need to do to pass utf-8 to WriteConsoleA is set windows console to expect it, before using it.

There are three aspects to consider: 

- Text  that your application will pass to console (windows has to expect it in utf-8),
- How console will try to render it (you need a windows font that has enough glyphs for most of utf8), and: 
- How the rest of your application will handle that (if you have writers to console, other than spdlog).

Something like the following will deal with the first two issues, and I'll leave the third as an exercise to the reader as it fall out of our context:
```
SetConsoleOutputCP(CP_UTF8);
CONSOLE_FONT_INFOEX cfi;
  cfi.cbSize = sizeof cfi;
  cfi.nFont = 0;
  cfi.dwFontSize.X = 0;
  cfi.dwFontSize.Y = 14;
  cfi.FontFamily = FF_DONTCARE;
  cfi.FontWeight = FW_NORMAL;
[...]

PR #2331, opened 2022-04-08
support print utf-8 string to non-unicode windows console.

dhmemi:
There are many feature-request about utf-8-supported wincolor-sink like https://github.com/gabime/spdlog/issues/762, https://github.com/gabime/spdlog/issues/1751. So I fixed the wincolor-sink to support output utf-8 string to non-utf8 code page console, which is consistent with win_eventlog_sink's behavior.


gabime:
The wgole purpose of the SPDLOG_WCHAR_TO_UTF8_SUPPORT define is to convert whcar to utf. I done
t see hte sense of converting it back.

PR #2335, opened 2022-04-10
V1.x

dust-II:
add compress sink

### DIFF
--- a/include/spdlog/sinks/compress_file_predicate.h
+++ b/include/spdlog/sinks/compress_file_predicate.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#include <string>
+#include <set>
+#include <map>
+
+struct binaryPredicate {
+    bool operator() (const std::string& l, const std::string& r) const {
+        if (l.size() != r.size()) {
+            return l.size() < r.size();
+        }
+
+        return l < r;
+    }
+};
+
+using name_set_t = std::set<std::string, binaryPredicate>;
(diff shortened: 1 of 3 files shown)


gabime:
Thanks. Seems bit out of scope for spdlog, especially since it relies on external command line tool.

dust-II:
Thank you for the trouble. Since I'm not very good at Using Github, I didn't actually plan on pulling requests.

BTW:
This compressed sink was developed by me on a Linux machine. It limits the total size of compressed logs. The elimination rules are 
   // Rotate files: max n is 65535，and n is  loop 1~65535,  This design solves the problem of compressing the log when the machine is powered down
    // log.1.gz -> delete 
    // log.2.gz -> delete
    // log.3.gz 
     ... 
     // log.n.txt

dust-II:
Thank you for the trouble. Since I'm not very good at Using Github, I didn't actually plan on pulling requests.

BTW:
This compressed sink was developed by me on a Linux machine. It limits the total size of compressed logs. The elimination rules are 
   // Rotate files: max n is 65535，and n is  loop 1~65535,  This design solves the problem of compressing the log when the machine is powered down
    // log.1.gz -> delete 
    // log.2.gz -> delete
    // log.3.gz 
     ... 
     // log.n.txt

gabime:
On second thought, using an external compressor has advantages.  It allows one to choose whatever compressor he wants without adding dependencies to spdlog.

dust-II:
Thank you for sharing spdlog and your thoughts on compressed logs

PR #2342, opened 2022-04-14
Make file_event_handlers an aggregate

espkk:
Just making file_event_handlers an aggregate to allow designated initialization


gabime:
I think std function is nullptr by default, so even this init is not needed.

espkk:
> I think std function is nullptr by default, so even this init is not needed.

yes indeed, nullptr_t constructor should be exactly the same. I left it so to keep the original explicit manner
But I agree it maybe makes sense to remove it completely

gabime:
Yes. Please remove it and I will merge.

espkk:
> Yes. Please remove it and I will merge.

done already. I think it was a race between my push and my comment here 😆

gabime:
Thanks @espkk

PR #2343, opened 2022-04-16
Add sync to file_helper

espkk:
Since file_helper already works well for managing file sinks and generally can be reused, it seems reasonable to me to implement fsync there which is required in some cases.
This adds `sync()` method to file_helper. It doesn't affect existing sinks and doesn't touch public interfaces.

### DIFF
--- a/include/spdlog/details/file_helper-inl.h
+++ b/include/spdlog/details/file_helper-inl.h
@@ -90,6 +90,14 @@ SPDLOG_INLINE void file_helper::flush()
     }
 }
 
+SPDLOG_INLINE void file_helper::sync()
+{
+    if(!os::fsync(fd_))
+    {
+        throw_spdlog_ex("Failed to fsync file " + os::filename_to_str(filename_), errno);
+    }
+}
+
 SPDLOG_INLINE void file_helper::close()
 {
     if (fd_ != nullptr)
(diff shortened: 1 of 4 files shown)


gabime:
Thanks. However since it is not used by any of spdlog sinks, I prefer not to add unused code,

gabime:
Thanks. However since it is not used by any of spdlog sinks, I prefer not to add unused code,

gabime:
@espkk Reopened due to popular demand for fsync. Could you please fix the code so it passes the tests?

espkk:
Should be good now.  I could add a sink for it but the use-cases can be different. I need it for termination, but someone might need it after each flush

gabime:
Thanks @espkk . I am wondering myself whether to 
1. add `fsync()` to the logger api.
2. or add `fsync()`just to file sinks.
3. or perhaps provide auto sync by using `logger->set_auto_fsync(true/false)`

espkk:
> Thanks @espkk . I am wondering myself whether to 
> 1. add `fsync()` to the logger api.
> 2. or add `fsync()`just to file sinks.
> 3. or perhaps provide auto sync by using `logger->set_auto_fsync(true/false)`

IMO, fsync on file sink should be flexible enough, while I would still prefer implementing custom sink if the desired behavior is sync-after-write

cpp977:
For my use case  (#2605), the third option would be the most convenient.
`logger->set_auto_sync()` would then lead to a call to `fsync` after each flush. One could then just use the existing `spdlog::flush_every()` to also have a regular `fsync` call.

PR #2346, opened 2022-04-22
Switch to vformat_to

sylveon:
Drive-by: reduce the amount of occurences of `#ifdef SPDLOG_USE_STD_FORMAT`

### DIFF
--- a/include/spdlog/details/fmt_helper.h
+++ b/include/spdlog/details/fmt_helper.h
@@ -23,6 +23,20 @@ inline spdlog::string_view_t to_string_view(const memory_buf_t &buf) SPDLOG_NOEX
     return spdlog::string_view_t{buf.data(), buf.size()};
 }
 
+#ifdef SPDLOG_USE_STD_FORMAT
+template<typename T>
+std::basic_string<T> to_string(const std::basic_string<T> &buf)
+{
+    return buf;
+}
+#else
+template<typename T, size_t Size>
+std::basic_string<T> to_string(const fmt::basic_memory_buffer<T, Size> &buf)
+{
+    return fmt::to_string(buf);
+}
+#endif
+
 inline void append_string_view(spdlog::string_view_t view, memory_buf_t &dest)
 {
     auto *buf_ptr = view.data();
--- a/include/spdlog/details/fmt_helper.h
+++ b/include/spdlog/details/fmt_helper.h
@@ -23,20 +23,6 @@ inline spdlog::string_view_t to_string_view(const memory_buf_t &buf) SPDLOG_NOEX
     return spdlog::string_view_t{buf.data(), buf.size()};
 }
 
-#ifdef SPDLOG_USE_STD_FORMAT
-template<typename T>
-std::basic_string<T> to_string(std::basic_string<T> &&buf)
-{
-    return std::move(buf);
-}
-#else
-template<typename T, size_t Size>
-std::basic_string<T> to_string(fmt::basic_memory_buffer<T, Size> &&buf)
-{
-    return fmt::to_string(buf);
-}
-#endif
-
 inline void append_string_view(spdlog::string_view_t view, memory_buf_t &dest)
 {
     auto *buf_ptr = view.data();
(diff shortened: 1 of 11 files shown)


gabime:
That’ s great. Thanks. Makes everything much cleaner.
There is small issue that now the logger.h includes the fmt_helper.h file and this might affect the compile times everywhere (it was on,ly included from “-inl.h” files, meaning included only when compiling the lib rather than from user code). Might be worth to place the new to_string in common.h instead.

sylveon:
I had only included that in there for to_string_view, which isn't really needed to begin with, so I went and removed that include from logger.h to return to manually building the string_view_t

include/spdlog/details/fmt_helper.h
```diff
@@ -23,6 +23,20 @@ inline spdlog::string_view_t to_string_view(const memory_buf_t &buf) SPDLOG_NOEX
     return spdlog::string_view_t{buf.data(), buf.size()};
 }
 
+#ifdef SPDLOG_USE_STD_FORMAT
+template<typename T>
+std::basic_string<T> to_string(const std::basic_string<T> &buf)
```
gabime, line 28: maybe could be optimize by changing the signature to 
```std::basic_string<T> &. to_string(const std::basic_string<T> &buf)```

File: include/spdlog/sinks/ringbuffer_sink.h [hunk 2]
```diff
@@ -50,11 +50,7 @@ class ringbuffer_sink final : public base_sink<Mutex>
         {
             memory_buf_t formatted;
             base_sink<Mutex>::formatter_->format(q_.at(i), formatted);
-#ifdef SPDLOG_USE_STD_FORMAT
-            ret.push_back(std::move(formatted));
-#else
-            ret.push_back(fmt::to_string(formatted));
-#endif
+            ret.push_back(fmt_helper::to_string(formatted));
```
gabime, line 53: Need to add include to fmt_helper.h file

include/spdlog/details/fmt_helper.h
sylveon: We don't actually benefit much from this specific form. All usages of `to_string` are to return by value or push_back, so we'll end up doing a copy anyways. A better form would be `std::basic_string<T> to_string(std::basic_string<T> &&buf)`, as this wouldn't copy. I'll apply that.

include/spdlog/sinks/ringbuffer_sink.h
sylveon: Good catch.

gabime:
Returning with std::move is problematic since it prohibits copy elision optimization and compiler is allowed to move anyway  (see  https://stackoverflow.com/questions/17473753/c11-return-value-optimization-or-move).

So I think your original signature in this pr (just return by value, no references or moves) is best after all. The compiler would optimize this easily anyway and the code is simplest. sorry for the confusion.

sylveon:
Calling `fmt_helper::to_string` in itself already kills NRVO.
Returning by `std::move` is only harmful when we do return a local variable by value, because it kills NRVO.
NRVO does not apply to returning parameters.

Until P1825R0, returning a rvalue reference does not move by default - we need to explicitly call std::move (this is because rvalue references are lvalues, thanks C++). This can be easily tested by trying to compile the following with GCC in C++17 mode or older (Clang backported P1825R0 to older standards). It won't compile:
```cpp
std::unique_ptr<int> test(std::unique_ptr<int> &&ptr)
{
    return ptr;
}
```
Changing the return to use `std::move(test)` allows it to compile.

---

All of the above is also confirmed by looking at [the assembly output](https://godbolt.org/z/3Gx9Y9TKf) of when using different to_string methods:
- test_no_func is the one with the shortest assembly output, because it allows NRVO to kick in
- Closely behind are test_rvalue_ref and test_rvalue_ref_non_moved, which do a few more `mov`s before returning, because they call the move operator.
- test_by_val_moved_std_move and test_by_val_std_move do not copy, but are bigger than the prior two because it does a double move.
- test_const_ref (the one passing `const std::string&` does a copy. So does test_by_val_moved_copy and test_by_val_copy.

So the "most optimal" solution is to use `#ifdef SPDLOG_USE_STD_FORMAT` to only call `fmt::to_string` when needed, but the one that's less intrusive to source code while being almost as efficient is to keep `std::string to_string(std::string&& str)` with an explicit call to `std::move` inside `to_string`.

gabime:
This all leads me to the think that ifdef should be used. Possibly by defining some macro like SPDLOG_BUF_TO_STRING(buf)

sylveon:
Good idea that I didn't think of - done.

gabime:
Merged. Thanks @sylveon

PR #2348, opened 2022-04-26
Added function to get the names of all registered loggers

tk-ka:
It'd like to find out from my application at runtime which loggers are registered, so that I can manipulate them (e.g. set the logging level) dynamically by their name.
For this purpose I have added a new top-level interface that returns a `std::vector` containing the keys of the private `registry::loggers_`.

### DIFF
--- a/include/spdlog/details/registry-inl.h
+++ b/include/spdlog/details/registry-inl.h
@@ -92,6 +92,17 @@ SPDLOG_INLINE std::shared_ptr<logger> registry::get(const std::string &logger_na
     return found == loggers_.end() ? nullptr : found->second;
 }
 
+SPDLOG_INLINE std::vector<std::string> registry::get_logger_names(void)
+{
+    std::lock_guard<std::mutex> lock(logger_map_mutex_);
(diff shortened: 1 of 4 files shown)


gabime:
Thanks. However this could be achieved using the ```spdlog:: apply_all(..)``` (https://github.com/gabime/spdlog/blob/b3ce5ed379ed01d9147b079e605e84dd408a0c19/include/spdlog/spdlog.h#L95)

tk-ka:
Thanks for the hint. I would prefer an API for that, but I understand your decision.
An convenient *external* solution would be:
```cpp
std::vector<std::string> get_logger_names(void) {
    std::vector<std::string> logger_names;
    spdlog::apply_all([&logger_names](std::shared_ptr<spdlog::logger> l) { logger_names.push_back(l->name()); });
    return logger_names;
}
```

gabime:
> I would prefer an API for that, but I understand your decision.

Thanks. I try to wait before adding new API until really needed (e.g. strong demand from the community), so I might still add it in the future.

PR #2352, opened 2022-04-28
Allow modifying the used Android buffer

tiolan:
Please consider this untested change WIP and a basis for discussion.

This would allow setting the Android log buffer to be used from outside.
Would something like this have a change to be merged?

The usage of a macro is not really nice, but simple and does not break the API.
Another approach could be to again template `android_sink` and provide a default template argument for not breaking the API.
Would need to see how this is compatible with C++11.

What are your thoughts?

### DIFF
--- a/include/spdlog/sinks/android_sink.h
+++ b/include/spdlog/sinks/android_sink.h
@@ -15,6 +15,7 @@
 #include <mutex>
 #include <string>
 #include <thread>
+#include <type_traits>
 
 #if !defined(SPDLOG_ANDROID_RETRIES)
 #define SPDLOG_ANDROID_RETRIES 2
@@ -24,13 +25,15 @@ namespace spdlog {
 namespace sinks {
 
 /*
- * Android sink (logging using __android_log_write)
- * __android_log_write is thread-safe. No lock is needed.
+ * Android sink
+ * (logging using __android_log_write or __android_log_buf_write depending on the specified BufferID)
+ * They are thread-safe. No lock is needed.
  */
-class android_sink : public sink
+template<int BufferID>
+class android_sink_buf : public sink
 {
 public:
-    explicit android_sink(const std::string &tag = "spdlog", bool use_raw_msg = false)
+    explicit android_sink_buf(const std::string &tag = "spdlog", bool use_raw_msg = false)
         : _tag(tag)
         , _use_raw_msg(use_raw_msg)
     {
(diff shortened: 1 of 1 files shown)


gabime:
Seems to me that passing the id thru the constructor would be good here.

tiolan:
The default is `LOG_ID_MAIN`.
I could probably just do a
`android_sink(int bufferId = LOG_ID_MAIN)`
and 
`__android_log_buf_write(bufferId, ...)`
just as they do here https://android.googlesource.com/platform/system/core/+/0ee8de3/liblog/logd_write.c#503.

I was not sure about different versions of liblog.
It seems not all of them offer `__android_log_buf_write`

https://android.googlesource.com/platform/system/core/+/0ee8de3/liblog/logd_write.c#501

https://android.googlesource.com/platform/system/core/+/android-2.1_r1/liblog/logd_write.c#135

So I wanted to keep the original behavior calling / linking the non-buf version `__android_log_write` if the user wants the default buffer.

gabime:
I guess another option is for the sink code to check the android version with ifdef and act accordingly.

tiolan:
yes or instead of a macro one could achieve this by a C++11 version of this:
```
android_sink<TBufferId = LOG_ID_MAIN>()
...
if constexpr(TBufferId == LOG_ID_MAIN)
    __android_log_write(...);
else
    __android_log_buf_write(bufferId, ...);

```
I would leave the decision and the "check" if logd supports it to the user and result in a compile error if `__android_log_buf_write` is not available.

gabime:
I think “ if constexpr” is c++17..

tiolan:
What do you think about this?

I used return value based SFINAE, in order to either call `android_log` with `__android_log_write` or `android_log` with `__android_log_buf_write` based on the `BufferID`, the user provided.

Since both methods have the same signature, it is guaranteed never both of them are compiled in at the same time.

So the user decides to use the currently existing mechanism via `__android_log_write` by just continue using `android_sink` or to use `__android_log_buf_write` by using `android_sink_buf<int>`.

I intentionally used `int` as template parameter and not `log_id` so that users can have non-default and custom log buffers.

This change should not break the API.

I tested it on a device.

For unittests, I did not find something to adapt - likely because the code is flagged under `__ANDROID__` and hence cross-compiled.

Would there be something else to adapt or test?

gabime:
Thanks. Looks good.

gabime:
Thanks. Looks good.

tiolan:
I just realized I requested the merge into master.
But master is behind of v1.x I guess.
How is your release strategy?
Will you merge from master to v1.x or should I have requested this to be merged to v1.x directly?

gabime:
Please open a new pr for v.x. Those branches are too diverge for merge.

PR #2365, opened 2022-05-07
Allow overriding need_localtime for custom formatter

conr2d:
If user wants to use the pattern with no builtin flags related to time, but custom formatter that needs time information, there is no way to make pattern_formatter pass the updated `const std::tm&` to `format()` of custom formatter. This PR allows user to override `need_localtime` so that custom formatter can utilize cached_tm with the pattern not enabling `need_localtime` by default.

### DIFF
--- a/tests/test_pattern_formatter.cpp
+++ b/tests/test_pattern_formatter.cpp
@@ -475,7 +475,7 @@ TEST_CASE("custom flags-exception", "[pattern_formatter]")
 
 TEST_CASE("override need_localtime", "[pattern_formatter]")
 {
-    auto formatter = std::make_shared<spdlog::pattern_formatter>();
+    auto formatter = std::make_shared<spdlog::pattern_formatter>(spdlog::pattern_time_type::local, "\n");
     formatter->add_flag<custom_test_flag>('t', "time").set_pattern("%t> %v");
 
     {
--- a/tests/test_pattern_formatter.cpp
+++ b/tests/test_pattern_formatter.cpp
@@ -330,11 +330,17 @@ public:
         : some_txt{std::move(txt)}
     {}
 
-    void format(const spdlog::details::log_msg &, const std::tm &, spdlog::memory_buf_t &dest) override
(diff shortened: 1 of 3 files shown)


gabime:
Thanks. Need to fid the clone() function ad well. Any idea why tests fail?

tests/test_pattern_formatter.cpp
```diff
@@ -466,3 +472,30 @@ TEST_CASE("custom flags-exception", "[pattern_formatter]")
...
+
+        auto now_tm = spdlog::details::os::localtime();
+        std::stringstream oss;
+        oss << (now_tm.tm_hour % 12) << ":" << std::setfill('0') << std::setw(2) << now_tm.tm_min << (now_tm.tm_hour / 12 ? "PM" : "AM")
+            << "> some message\n";
+
+        memory_buf_t formatted;
+        spdlog::details::log_msg msg(spdlog::source_loc{}, "logger-name", spdlog::level::info, "some message");
+        formatter->format(msg, formatted);
+        REQUIRE(to_string_view(formatted) == oss.str());
```
gabime, line 499: This can fail if the  time changes when getting here

gabime (changes requested):

conr2d:
Thank you for fast feedback.

- I added copying the value of `need_localtime` to `pattern_formatter::clone()`.
- Test failed due to eol symbol. I guess that automated test for this repository is processed on WIN32 environment, and eol would be `\r\n` instead of `\n`. I set eol for the pattern formatter in test to `\n` explicitly.
- As you mentioned, test can fail with low probability if the expected string and actually formatted string are generated in different seconds. I followed the one case of original tests, [date MM/DD/YY](https://github.com/gabime/spdlog/blob/bd0198de2de4bc24d13806fc03da7555deda9293/tests/test_pattern_formatter.cpp#L54-L61), but if you think the tests should be fixed in both cases, I will investigate how to fix the time to refer.

gabime:
Thanks @conr2d . Merged. 
If you could investigate on fixing the formatter date tests it would be great.

Issue #2333, opened 2022-04-09
build example error on gcc 4.8 .


aengusjiang:
in file: bin_to_hex.h

#if defined(__has_include) && __has_include(<version>)
#    include <version>
#endif

error: missing binary operator before token "("
 #if defined(__has_include) && __has_include(<version>)

gabime:
Thanks. which compiler version?

aengusjiang:
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

and the spdlog version:1.10

aengusjiang:
step: 
cmake -B build .
cd build 
make 

then the error as below:
![image](https://user-images.githubusercontent.com/22406984/162558381-8c527066-41d9-421f-90ac-1c36001d7263.png)

gabime:
I dont have gcc 4.8 to reproduce. Please open a pr to fix

aengusjiang:
open a pr?
What's mean?

What can to do for you?




发自我的手机


-------- 原始邮件 --------
发件人： Gabi Melman ***@***.***>
日期： 2022年4月9日周六 19:40
收件人： gabime/spdlog ***@***.***>
抄送： "Aengus.Jiang" ***@***.***>, Author ***@***.***>
主 题： Re: [gabime/spdlog] build example error on gcc 4.8 . (Issue #2333)

I dont have gcc 4.8 to reproduce. Please open a pr to fix

gabime:
Create a pull request with the fix.

aengusjiang:
According to the gcc support doc, the __has_include macro support required gcc 5.0 and up.

发自我的手机


-------- 原始邮件 --------
发件人： ***@***.***
日期： 2022年4月9日周六 19:49
收件人： gabime/spdlog ***@***.***>
主 题： 回复：[gabime/spdlog] build example error on gcc 4.8 . (Issue #2333)
open a pr?
What's mean?

What can to do for you?




发自我的手机


-------- 原始邮件 --------
发件人： Gabi Melman ***@***.***>
日期： 2022年4月9日周六 19:40
收件人： gabime/spdlog ***@***.***>
抄送： "Aengus.Jiang" ***@***.***>, Author ***@***.***>
[...]

PR #2372, opened 2022-05-11
C++14 build fixes for older gcc #2333

kslattery:
@gabime  I'm unable to test that the changes for `__has_include` in bin_to_hex.h work for @aengusjiang, but the approach I've followed matches the recommended pattern in the [gcc documentation for __has_include](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html).

I have tested the rest of the changes on my machine with gcc version `gcc (Ubuntu 5.5.0-12ubuntu1~14.04) 5.5.0 20171010`. I get the overloaded formatter used as seen in the output from the example program:

> [2022-05-11 18:22:25.008] [info] user defined type: [my_type i=14]

The problem seems to be caused by this clause in the [namespace specification](https://en.cppreference.com/w/cpp/language/namespace) (emphasis added):
>To reopen an existing namespace (formally, to be an extension-namespace-definition), the lookup for the identifier used in the namespace definition must resolve to a namespace name (**_not a namespace alias_**), that was declared as a member of the enclosing namespace or of an inline namespace within an enclosing namespace.

The approach I've taken here is the same approach used within `fmt` for the `FMT_BEGIN_NAMESPACE` macro defined in include/spdlog/fmt/bundled/core.h.

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -262,15 +262,16 @@ struct my_type
         : i(i){};
 };
 
-namespace fmt_lib = spdlog::fmt_lib;
+FMTLIB_BEGIN_NAMESPACE
 template<>
-struct fmt_lib::formatter<my_type> : fmt_lib::formatter<std::string>
+struct formatter<my_type> : formatter<std::string>
 {
     auto format(my_type my, format_context &ctx) -> decltype(ctx.out())
     {
-        return fmt_lib::format_to(ctx.out(), "[my_type i={}]", my.i);
+        return format_to(ctx.out(), "[my_type i={}]", my.i);
     }
 };
+FMTLIB_END_NAMESPACE
 
 void user_defined_example()
 {
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -262,7 +262,14 @@ struct my_type
         : i(i){};
 };
 
-FMTLIB_BEGIN_NAMESPACE
+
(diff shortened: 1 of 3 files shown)


gabime:
Thanks a lot. I think the has_include fix is ok, but the namespace define can be replaced with an example instead. I prefer not to add yet another ifdef to the code if possible.

kslattery:
Makes sense. I wasn't sure what you'd want to do. The problem with just using `namespace fmt` in the example code is that the example will then not work if using the std formatter. I'd suggest putting something like this in the example to make it work either way:
```
#ifdef SPDLOG_USE_STD_FORMAT
    namespace std {
#else 
    namespace fmt {
#endif
template<>
struct formatter<my_type> : formatter<std::string>
...
}. // end of namespace

```

gabime:
> I'd suggest putting something like this in the example to make it work either way

Yes. Thank you.

kslattery:
Ok, do you want me to create a new commit with those changes? Or do you want to just take it from here?

gabime:
Please do.

kslattery:
@gabime If you don't like the comment I added, let me know what, if anything, you'd like in its place.

gabime:
Thanks @kslattery

anhns26693:
Thanks @kslattery very much

Issue #2374, opened 2022-05-12
undefined reference for std::function when not using file_event_handlers in daily_file_sink constructor call


kslattery:
I have the following code to create a daily_file_sink with an overridden calc_filename():
```
    spdlog::init_thread_pool(4096,1);
    _thread_pool = spdlog::thread_pool();
    std::vector<spdlog::sink_ptr> sinks;
    sinks.push_back(std::make_shared<spdlog::sinks::daily_file_sink<std::mutex, custom_file_name_calculator>>(_output_dir, 0, 0));
    auto logger = std::make_shared<spdlog::async_logger>("mydailylogger", begin(sinks), end(sinks), std::move(_thread_pool), spdlog::async_overflow_policy::block);
    spdlog::register_logger(logger);
```

When this is linked, I get the following errors:
```
CMakeFiles/my_project.dir/src/RuntimeFileLogger.cpp.o: In function `void __gnu_cxx::new_allocator<spdlog::sinks::daily_file_sink<std::mutex, custom_file_name_calculator> >::construct<spdlog::sinks::daily_file_sink<std::mutex, custom_file_name_calculator>, std::string&, int, int>(spdlog::sinks::daily_file_sink<std::mutex, custom_file_name_calculator>*, std::string&, int&&, int&&)':
/usr/include/c++/5/ext/new_allocator.h:120: undefined reference to `std::function<void (std::string const&, _IO_FILE*)>::function()'
/usr/include/c++/5/ext/new_allocator.h:120: undefined reference to `std::function<void (std::string const&, _IO_FILE*)>::function()'
collect2: error: ld returned 1 exit status
```

Note that I'm not specifying every parameter for the daily_file_sink constructor, but relying on the default values specified here:
```
    daily_file_sink(filename_t base_filename, int rotation_hour, int rotation_minute, bool truncate = false, uint16_t max_files = 0,
        const file_event_handlers &event_handlers = {})
```

The problem seems to come from the fact that in places like file_helper::open(), we have code like this:
```
    if (event_handlers_.before_open)
    {
        event_handlers_.before_open(filename_);
    }
```

So although the before_open() call would never be made when using the default `event_handlers = {}`, the linker is still looking for the symbol which simply does not exist.

I worked around this by changing my code to be this:
```
    spdlog::init_thread_pool(4096,1);
    _thread_pool = spdlog::thread_pool();
    spdlog::file_event_handlers handlers;
[...]

PR #2375, opened 2022-05-13
Add default file-event_handler callbacks. #2374

kslattery:
Prevents undefined reference errors seen in issue #2374.


gabime:
Thanks @kslattery

gabime:
@kslattery I just realized that initialized functions in the pr would cause spdlog calling them (as opposed to nullptr), so I modified it. Please check that latest commit still solves the undefined error.

PR #2376, opened 2022-05-13
build: expand SOVERSION to not give false illusion of compatibility

### LINKED ISSUE #2369: Unresolved symbols when moving to spdlog 1.10 without recompile - maybe full symbol
The cryfs package uses spdlog 1.8.x. openSUSE moved to spdlog 1.10. This resulted in unresolved symbols, because of a ABI break in one of the later commits: c47ae3b15dedc35cc62cc34ec1e411b040373776

Would it be possible to include version symbols in the releases? Or somehow maintain backward compatibility?

See our bugreport: https://bugzilla.opensuse.org/show_bug.cgi?id=1199306#c0

jengelh:
Fixes #2369


gabime:
Thanks @jengelh

Issue #2379, opened 2022-05-18
Compile error when using spdlog with FMT_ENFORCE_COMPILE_STRING


john4744:
I'm building an exe that uses spdlog and fmt. I'm building with FMT_ENFORCE_COMPILE_STRING defined. This is causing the following compile error:
```
c:\apps\3rdparty\spdlog\include\spdlog\fmt\bundled\core.h(603,39): error C2338: FMT_ENFORCE_COMPILE_STRING requires all format strings to use FMT_STRING.
c:\apps\3rdparty\spdlog\include\spdlog\fmt\bundled\core.h(3057): message : see reference to function template instantiation 'void fmt::v8::detail::check_format_string<,char[21],0>(const S (&))' being compiled
        with
        [
            S=char [21]
        ]
c:\apps\3rdparty\spdlog\include\spdlog\fmt\bundled\format-inl.h(96): message : see reference to function template instantiation 'fmt::v8::basic_format_string<char>::basic_format_string<char[21],0>(const S (&))' being compiled
        with
        [
            S=char [21]
        ]
c:\apps\3rdparty\spdlog\include\spdlog\fmt\bundled\format-inl.h(96): message : see reference to function template instantiation 'fmt::v8::basic_format_string<char>::basic_format_string<char[21],0>(const S (&))' being compiled
        with
        [
            S=char [21]
        ]
```
Not sure what I'm doing wrong, any help would be appreciated. Thanks!

tt4g:
Please wrap the format string with the `FMT_STRING` macro if define `FMT_ENFORCE_COMPILE_STRING` macro (this is fmt library feature.).

```cpp
spdlog::info(FMT_STRING("Log with {}."), "macro"); // OK.

spdlog::info("Log without {}.", "macro"); // Assertion error.
```

john4744:
Thanks for the feedback! In this case, I only have `#include "spdlog/spdlog.h"`, no actual calls into the spdlog API. I think I may have tracked down the problem to a misconfiguration on my part 🤦‍♂️ . I noticed that the `example` project has `SPDLOG_COMPILED_LIB` defined and my project didn't. Adding this define to my project seems to resolve the compile error.

john4744:
Actually, there still may be a problem here building for Android... When I include `android_sink.h` I get this compile error:
```
[...]

PR #2381, opened 2022-05-18
Add FMT_STRING to allow compilation with FMT_ENFORCE_COMPILE_STRING

john4744:
See https://github.com/gabime/spdlog/issues/2379


gabime:
Thanks @john4744

PR #2383, opened 2022-05-19
Fix pkg-config generation with unconventional `CMAKE_INSTALL_*DIR`

### LINKED ISSUE #2380: pkg-config file is broken when CMAKE_INSTALL_LIBDIR is absolute
As per title: `cmake/spdlog.pc.in` has (since https://github.com/gabime/spdlog/pull/1238)
```
prefix=@CMAKE_INSTALL_PREFIX@
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
```
and so can’t handle absolute paths in `CMAKE_INSTALL_LIBDIR`. This leads to broken .pc files on NixOS in particular.

See “[Concatenating paths when building pkg-config files](https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files)” for a thorough discussion of the problem and a suggested fix, or [KDE’s extra-cmake-modules](https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/modules/ECMGeneratePkgConfigFile.cmake#L166) for a simpler approach.

alexshpilkin:
Fixes #2380 and a further issue where the pkg-config template blindly assumed that CMAKE_INSTALL_INCLUDEDIR was equal to `include` even though it’s externally configurable and the actual installation process respected it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -298,6 +298,11 @@ if(SPDLOG_INSTALL)
     # ---------------------------------------------------------------------------------------
     # Install pkg-config file
     # ---------------------------------------------------------------------------------------
+    if (IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
+        set(PKG_CONFIG_LIBDIR "${CMAKE_INSTALL_LIBDIR}")
+    else()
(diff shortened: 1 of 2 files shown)


gabime:
Thanks @alexshpilkin

PR #2384, opened 2022-05-19
fix error: cannot bind lvalue to right reference

aengusjiang:
before fix error:
image

this commit is verify with gcc4.8.5 and gcc8.3.1


gabime:
Thanks @aengusjiang

aengusjiang:
Before the try to fix,  it like below:
return cloned;

when I use gcc 4.8 and gcc8.3， it will has the error belore：
image

aengusjiang:
I not test it with gcc 11.2.0.

PR #2385, opened 2022-05-20
Remove `try_lock` from null_mutex.

panzhongxian:
Only `std::unique_lock` and `std::lock_guard` are used on member variable `mutex_`. Both of them should meet the [BasicLockable](https://en.cppreference.com/w/cpp/named_req/BasicLockable) requirements, which means `lock()` and `unlock` are enough.


gabime:
Thanks @panzhongxian

PR #2386, opened 2022-05-24
Romove the empty file if no log in first period in hourly logger

panzhongxian:
If there is no logger sinked in first period, there will be an empty log file, which is not consistent with what it behaves in the consequent period(no file if no log).


gabime:
Thanks @panzhongxian

PR #2396, opened 2022-06-02
Remove redundant std::move at return 

polesapart:
The statement triggers -Wredundant-move in GCC. Return value optimization already moves/constructs in place if target allows it.


gabime:
Thanks @polesapart

Issue #2397, opened 2022-06-06
how to know if async queue is full?


bergen4:
overrun_counter() function  can get the number of logs that have been overwritten since the service was started,
but I need to know periodically if the spdlog queue if full. so how to handle this situation

tt4g:
`spdlog::details::thread_pool::queue_size()` is available.

bergen4:
> `spdlog::details::thread_pool::queue_size()` is available.

My scenario is as follows. When I deal with production problems with spd logs, first i want to confirm whether the logs has been dropped during this period. Then I need to regularly check whether the queue is full, but the queue_size() can only Get the size of the current queue without knowing whether it is full during this period

tt4g:
spdlog does not provide an API to report when the message queue is full.
And unless you specify `async_overflow_policy::overrun_oldest`, log messages that have not been sent to sink will not be dropped from the message queue.

https://github.com/gabime/spdlog/blob/76fb40d95455f249bd70824ecfcae7a8f0930fa3/include/spdlog/async_logger.h#L21-L27

bergen4:
block mode is too dangerous for service.Since there is no api to get the full status of the queue , I will add one. Thank you for your reply

PR #2399, opened 2022-06-09
add overrun_counter reset function

bergen4:
function for monitor queue status.
detail info as bellow:
https://github.com/gabime/spdlog/issues/2397


gabime:
Thanks @bergen4

PR #2407, opened 2022-06-16
stopwatch: change to system_clock

### LINKED ISSUE #2404: test_stopwatch2 fails on UCRT64 (MSYS2)
```
  REQUIRE( val >= wait_duration.count() )^M
with expansion:^M
  0.2496895 >= 0.25^M
```

neheb:
test_stopwatch fails on MinGW, most likely because its steady_clock implementation is faulty.

Fixes: https://github.com/gabime/spdlog/issues/2404

### DIFF
--- a/include/spdlog/stopwatch.h
+++ b/include/spdlog/stopwatch.h
@@ -27,7 +27,7 @@
 namespace spdlog {
 class stopwatch
 {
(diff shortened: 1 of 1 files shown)


neheb:
nvm. just failed.

PR #2417, opened 2022-06-28
[FEAT] add api: set_thread_name, used in periodic_worker and thread_pool

izlyforever:
We use spdlog in our LogService and will see same thread name when debuging
It may be nice to give a name for spdlog thread rather than using it's parent thread name

I only test this PR in Linux/Ubuntu21.10 and MacOS M1
> I have no device for testing other platforms, thanks ci/cd

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -60,6 +60,32 @@
 
 #endif // unix
 
+// set_thread_name
+#if defined(_WIN32) && defined(_MSC_VER)
+#    include <windows.h> // RaiseException
+#endif
+
+#if defined(__linux__)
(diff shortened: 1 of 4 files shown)


gabime:
Thanks @izlyforever . Seems it can be achieved by the user using the `on_thread_start` callback in the init thread pool function. https://github.com/gabime/spdlog/blob/d7690d8e7eed721d78b52e032e996a5d1ef47d6f/include/spdlog/async.h#L77
 In this callback the user can implement whatever naming needed.

izlyforever:
Thanks @gabime, is perfect to have `on_thread_start ` . However we use `flush_every` and `rotating_logger_mt`, both can't use `on_thread_start` simply

Issue #2419, opened 2022-06-30
Logger set_pattern has unexpected behavior


jb-gcx:
I would expect it to call `set_pattern` on each sink. Instead it calls `set_formatter`, which overwrites the previous formatter completely, including the custom eol string. This is causing an issue for me that I'm currently working around, as I do require a custom eol string in one of my sinks.

tt4g:
You can use `spdlog::sinks::set_formatter()` instead of `spdlog::logger::set_formatter()`.
And override EOL by constructing `spdlog::pattern_formatter`.

https://github.com/gabime/spdlog/blob/9b4b37312157d30de928bea8347eb4382f89db07/include/spdlog/pattern_formatter.h#L77-L87

jb-gcx:
Thanks for the quick response @tt4g ! 👍 
I know I can solve my problem, it's not blocking me. I opened this issue because it seems to me that the intended (and I would argue intuitive) codepath is `logger::set_pattern` -> `sink::set_pattern`. Not `logger::set_pattern` -> `logger::set_formatter` -> `sink::set_formatter`.
If I call `logger::set_pattern`, I expect that to arrive at `base_sink::set_pattern_` eventually, which it doesn't, making it unusable in my case.

*Edit:* I've created a PR, in case you agree.

tt4g:
Indeed, the comments in the source code do not explain this behavior.

https://github.com/gabime/spdlog/blob/d7690d8e7eed721d78b52e032e996a5d1ef47d6f/include/spdlog/logger.h#L318-L320

If you think someone will encounter problems with this behavior, please submit a PR. However, I am not sure if the PR will be accepted since it is not API compatible.

EDIT: IMO. `spdlog::pattern_formatter` does pattern parsing in its constructor. This process may worsen performance as the number of patterns increases, so the current implementation is probably made to reduce the number of pattern parsing.

jb-gcx:
> EDIT: IMO. spdlog::pattern_formatter does pattern parsing in its constructor. This process may worsen performance as the number of patterns increases, so the current implementation is probably made to reduce the number of pattern parsing.

That makes sense. But I would argue for logger:set_pattern consistency is more important than performance, since I can hardly imagine a use-case that requires regular calls. If there is such an unusual use-case it can still be done by using logger:set_formatter directly.

gabime:
I added comment to clarify this behavior. I see no way to fix this without breaking API and current behavior.

PR #2420, opened 2022-06-30
Have logger::set_pattern call sink::set_pattern #2419

jb-gcx:
Previously it instantiated a new pattern_formatter and called sink::set_formatter
with it, overwriting the old formatter and any custom settings.
Now, if a custom sink inherits base_sink, a call from logger::set_pattern will arrive
at custom_sink::set_pattern_ as expected, instead of custom_sink::set_formatter_.

This PR addresses an actual issue I am having, where I instantiate a pattern_formatter in the constructor of my custom sink, because I need a custom eol string. If, afterwards, `logger::set_pattern` is called, this pattern_formatter is overwritten completely. The best I can do to prevent this subtle error is overwrite `base_sink::set_formatter_` to make it throw, which makes `logger::set_pattern` unusable.

### DIFF
--- a/include/spdlog/logger-inl.h
+++ b/include/spdlog/logger-inl.h
@@ -100,8 +100,10 @@ SPDLOG_INLINE void logger::set_formatter(std::unique_ptr<formatter> f)
 
 SPDLOG_INLINE void logger::set_pattern(std::string pattern, pattern_time_type time_type)
 {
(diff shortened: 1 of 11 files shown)


gabime:
Thanks. However this breaks the api and can be easily workaround using the set_formatter() function.

PR #2426, opened 2022-07-08
[FEAT] add api:overrun_counter

izlyforever:
make it easy to know how many log lost when using  spdlog async_logger


gabime:
Since this is a rare requirement and since the thread pool can be queried directly for this info, I prefer no to add it to the API. Thanks anyway.

PR #2439, opened 2022-07-17
Expend support for any std::chrono::duration in spdlog::flush_every

LucasChollet:
This allows things like:

```cpp
spdlog::flush_every(std::chrono::minutes(10));
spdlog::flush_every(std::chrono::milliseconds(100));
```

### DIFF
--- a/include/spdlog/details/periodic_worker-inl.h
+++ b/include/spdlog/details/periodic_worker-inl.h
@@ -10,27 +10,6 @@
 namespace spdlog {
 namespace details {
 
-SPDLOG_INLINE periodic_worker::periodic_worker(const std::function<void()> &callback_fun, std::chrono::seconds interval)
-{
-    active_ = (interval > std::chrono::seconds::zero());
-    if (!active_)
-    {
-        return;
-    }
-
-    worker_thread_ = std::thread([this, callback_fun, interval]() {
-        for (;;)
-        {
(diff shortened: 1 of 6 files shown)


tests/test_async.cpp
```diff
@@ -88,8 +88,8 @@ TEST_CASE("async periodic flush", "[async]")
     auto logger = spdlog::create_async<spdlog::sinks::test_sink_mt>("as");
     auto test_sink = std::static_pointer_cast<spdlog::sinks::test_sink_mt>(logger->sinks()[0]);
 
-    spdlog::flush_every(std::chrono::seconds(1));
-    std::this_thread::sleep_for(std::chrono::milliseconds(1700));
+    spdlog::flush_every(std::chrono::milliseconds (100));
+    std::this_thread::sleep_for(std::chrono::milliseconds(170));
```
gabime: We are experiencing problems with sleep accuracy when the tests are running in vm. Please revert this test to the original 1 sec + 1700 ms

gabime (changes requested):
Thanks. Looks good. Please revet the changes in the test

LucasChollet:
I removed `SPDLOG_API` from `spdlog::flush_every` declaration and replace it with `inline`.
It solves the CI issue.

gabime:
Thanks @LucasChollet

PR #2443, opened 2022-07-21
Explicitly casting level_enum to size_t.

ibmibmibm:
These windows-only code need to be fix due to commit 2a4c34b8785137eba9da7eb4cbb28b4162218272


gabime:
Thanks @ibmibmibm

PR #2445, opened 2022-07-25
Removed doc "(shared not supported in windows yet)"

Hish15:
Since SHARED library are handled bellow this documentation, I assume it's no longer true.


gabime:
Thanks @Hish15

Hish15:
You're welcome. I noticed that this is also on the v2.x branch.
Cya

PR #2449, opened 2022-07-27
Re-introduce std::move at return for old GCC (before version 5) inside an ifdef for e.g. CentOS 7

Simon-Janos:
Re-introduce std::move at return for old GCC (before version 5) inside an ifdef for e.g. CentOS 7


gabime:
Thanks @Simon-Janos

PR #2460, opened 2022-08-04
color_sinks use fmtlib styled colors and emphasis instead of predefined string color ansicodes and support ansicode on windows

mguludag:
Hello contributors,
I want to introduce fmtlib's color implementation into spdlog color_sinks and no need to rely on fmtlib source code and copied into spdlog's source. Converted necessary macros to spdlog's macro names and copied/edited needed functions and definitions. After that, we got more color and styling options with familiar syntax from fmtlib. 

## Usage
```C++
#define SPDLOG_WINDOWS_USE_ANSICOLOR // use ansicolors on windows (windows terminal etc.) instead of wincolor sink
#include "spdlog/spdlog.h"
#include "spdlog/sinks/stdout_color_sinks.h"
void stdout_example()
{
    // create color multi threaded logger
    auto console = spdlog::stdout_color_mt("console");
    auto console_sink = static_cast<spdlog::sinks::stdout_color_sink_mt*>(console->sinks().at(0).get());
    console_sink->set_style(spdlog::level::info, spdlog::fg(spdlog::color::cornflower_blue) |  spdlog::bg(spdlog::color::white) |
    spdlog::emphasis::bold | spdlog::emphasis::italic); // or custom rgb color -> spdlog::fg(spdlog::rgb(0, 0, 255))
    spdlog::get("console")->info("loggers can be retrieved from a global registry using the spdlog::get(logger_name)");
 }
 ```

### DIFF
--- a/include/spdlog/common-inl.h
+++ b/include/spdlog/common-inl.h
@@ -1,4 +1,4 @@
-// Copyright(c) 2015-present, Gabi Melman & spdlog contributors.
+// Copyright(c) 2015-present, Gabi Melman & spdlog and fmtlib contributors.
 // Distributed under the MIT License (http://opensource.org/licenses/MIT)
 
 #pragma once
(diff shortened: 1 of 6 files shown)


gabime:
Thanks for the pr. Is there a way to use fmt colors without copy its code? Id rather not add duplicate fmt’s code.

mguludag:
> Thanks for the pr. Is there a way to use fmt colors without copy its code? Id rather not add duplicate fmt’s code.

Another way is using these ansicodes as literals. I can write another implementation but it'll be similar to this.

gabime:
I will close this for now. Will reconsider if there is more demand for this. Thanks anyway!

PR #2475, opened 2022-09-02
cmake: set(CMAKE_POSITION_INDEPENDENT_CODE ON) 

nigels-com:
We're using spdlog as a static library.  On Linux it's necessary to use `-fPIC` flag to link spdlog into shared or executable targets.


gabime:
This depends on usage.. For example if bulding spdlog as static lib it might be not desirable.

nigels-com:
> This depends on usage.. For example if bulding spdlog as static lib it might be not desirable.

As an opt-in option, then?

gabime:
Yes. Thanks

gabime:
Thanks @nigels-com

PR #2481, opened 2022-09-08
cmake: prevent leak of libfmt dependency into spdlog users

mroethke:
When linking dynamically against spdlog there is no need for the user of the shared library to itself link against the dependencies of that spdlog. Doing it anyway causes problems when one of the dependencies gets a soname bump, because now the library user needs to be rebuild despite not using a single symbol of the updated dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,7 +213,9 @@ if(SPDLOG_FMT_EXTERNAL OR SPDLOG_FMT_EXTERNAL_HO)
         target_link_libraries(spdlog_header_only INTERFACE fmt::fmt)
     endif()
 
-    set(PKG_CONFIG_REQUIRES fmt) # add dependency to pkg-config
+    if(NOT SPDLOG_BUILD_SHARED)
+        set(PKG_CONFIG_REQUIRES fmt) # add dependency to pkg-config
+    endif()
 endif()
 
 # ---------------------------------------------------------------------------------------


gabime:
I am mot sure exactly how PKG_CONFIG_REQUIRES works. Would spdlog (when used with pkg config) still be able to find fmt’s header files if PKG_CONFIG_REQUIRES is omitted ?

mroethke:
Yes, in the cases where it is necessary, this only omits it for the shared library build.

gabime:
Not sure I follow. The fmt include location is always needed, even in shared build.

mroethke:
Yes, for building spdlog, but not for linking against its shared library.

Consider this simple program:
```c++
#include "spdlog/spdlog.h"

int main() {
	spdlog::info("test");
}
```

When I query the pc file for the required compiler flags using pkg-config with this command `pkg-config --clfags --libs` I get:
```
-DSPDLOG_SHARED_LIB -DSPDLOG_COMPILED_LIB -DSPDLOG_FMT_EXTERNAL -lspdlog -pthread -lfmt 
```
Now compiling the program with the flags pkg-config told me to use (`g++ -o test test.cpp $(pkg-config --cflags --libs)`) results in the following shared library dependencies (queried with `objdump -p test`):
```
…
Dynamic Section:
  NEEDED               libspdlog.so.1
  NEEDED               libfmt.so.9
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
…
```

According to this my little program now directly depends on libfmt, despite not using a single function from the library.

However if I remove `-lfmt` from the compiler arguments (`g++ -o test test.cpp -DSPDLOG_SHARED_LIB -DSPDLOG_COMPILED_LIB -DSPDLOG_FMT_EXTERNAL -lspdlog -pthread`)  it looks like this:
```
…
Dynamic Section:
  NEEDED               libspdlog.so.1
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
…
```
Which is what I would want, because my little program does not use libfmt and should therefore not depend on it directly.

And if I'm not mistaken this is the only thing my patch does, tell pkg-config not to link against libfmt directly for a shared library, or rather the reverse stop telling it to link against libfmt when linking against spdlog. It does not affect building of spdlog itself.

gabime:
fmt is being used by spdlog in headers (e.g, https://github.com/gabime/spdlog/blob/81de01c02c2b49aa723806b0c1d557dd05721635/include/spdlog/logger.h#L402) which means your program depend on fmt directly  when logging anything with format string.

mroethke:
ah, I see.
Is changing that on the table? I think it is generally preferable to not leak dependencies into dependents.

gabime:
Stop using fmt from the header? I don't think its technically possible since those are variadic template functions

mroethke:
hmm, you are probably right about that. Sorry for the noise.

PR #2483, opened 2022-09-08
Add log about dropped messages

ewok261:
Tracking dropped messages as the drops happen would be very useful in cases where someone is inspecting the logs and wonders if the program just skipped some parts that were supposed to be logged or if the logging was just slow.

The API was already there (overrun_counter), but in this PR it's used from the same thread that does the actual logging to the sink. Otherwise, the user would have to poll the thread_pool's overrun_counter from a different thread, which I think is not as nice.

### DIFF
--- a/include/spdlog/async_logger-inl.h
+++ b/include/spdlog/async_logger-inl.h
@@ -7,6 +7,7 @@
 #    include <spdlog/async_logger.h>
 #endif
 
+#include <spdlog/spdlog.h>
 #include <spdlog/sinks/sink.h>
 #include <spdlog/details/thread_pool.h>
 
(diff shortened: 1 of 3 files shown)


gabime:
Thanks. This could be problematic. When overrun counter >0 it means that the worker thread cant keep up the pace. And under those circumstances the last thing you want is giving it extra logging work.

ewok261:
Thank you for the fast response! That's a valid point, however,  I saw the log about the dropped messages more like a signal to the user to increase the number of worker threads next time they run the program, or to reduce the number of times they are calling the logger, if possible. In any case, the feedback info could be useful for future runs.

PR #2495, opened 2022-09-26
Use lower-case "windows.h" for case-sensitive file systems

panicgh:
The "windows.h" in MinGW-W64 is lower-case. When cross-compiling for Windows on Linux with a case-sensitive file system, the upper-case "Windows.h" file is not found and compilation fails.

Always use lower-case "windows.h" to fix cross-compilation.


gabime:
Thanks @panicgh

Issue #987, opened 2019-02-04
windows: rotating_file_sink: failed renaming


htonus:
Hello! 
Created logger by:
`spd::rotating_logger_mt<spd::async_factory>("MyLogger", "log\\mylog.txt", 5000, 1);`

In some time always get following error:
`[*** LOG ERROR ***] [2019-02-05 00:11:32] [MyLogger] rotating_file_sink: failed renaming logs\mylog.txt to logs\mylog.1.txt: Permission denied`

The file itself has all permissions required for the action.

gabime:
Do you have antivirus turned on? I saw this happening under windows with AV.

gabime:
as a workaround try to increase the 5000to something bigger. e.g. to 1024*500

htonus:
That was 2M when first happened and takes too long to re-produce. So the size was reduced to speed up the issue to happen.

htonus:
Tried to add my app to exclusions and disabled AV totally - doesn't solve the issue.

gabime:
And the setup vbox host with win10 guest? spdlog compiled with msvc?

htonus:
I compile my app in VS2017CE, Windows 1809, Build 17763.253. Hosting OS. But the issue happens with app in VM, will try to reproduce on hosting system.

gabime:
spdlobeg v1.3?

htonus:
```
c:\work\c++\vcpkg>vcpkg.exe list
fmt:x64-windows                                    5.3.0            Formatting library for C++. It can be used as a ...
fmt:x86-windows                                    5.3.0            Formatting library for C++. It can be used as a ...
openssl-windows:x64-windows                        1.0.2q           OpenSSL is an open source project that provides ...
openssl-windows:x86-windows                        1.0.2q           OpenSSL is an open source project that provides ...
openssl:x64-windows                                0                OpenSSL is an open source project that provides ...
openssl:x86-windows                                0                OpenSSL is an open source project that provides ...
spdlog:x64-windows                                 1.3.1            Very fast, header only, C++ logging library
spdlog:x86-windows                                 1.3.1            Very fast, header only, C++ logging library
```

gabime:
Can't reproduce

artemyv:
I can describe one of the scenarios where it happend for me. I'm using log from service. So log file is created as system. But some function use user impersonation. So this parts write to log as user. If the rotation is happening while such function tries to write to log rotate function will fail to rename file.

[...]

PR #2503, opened 2022-10-07
Log file rotation improvements

epronk:
### Motivation

When a service is restarted it is possible renaming a log file fails because of Windows file-locking or receive an "access denied". (see Issue #987) I have observed losing a log file, because the `rename_file_` removes the target first.

The implementation of `rename_file_` always removes the target before renaming source to target. Removing the target is only necessary to replace the oldest file. 

The goal of this change it to:
1. eliminate the remove from `rename_file_`
2. make removing the oldest file an explicit action.

### Description

Proposed solution to make rotation more robust
This change deals with a failed rename, but does not prevent it.

1. Changed rename_file_ so it doesn't delete the target.

2. Changed rotate_ so it removes the last file before starting the rename loop.

3. When rename fails the rotation stops and leaves gaps between the files.

- Added `recover_` function to deal with gaps caused by a failed rename.
- In case of gaps, move files in a single pass. Don't move the oldest file.
- If gaps still exist after first pass, then those will be dealt with next time `recover_` runs

- [x] I have added tests to cover my changes if appropriate.
- [x] All new and existing tests passed.

### DIFF
--- a/include/spdlog/sinks/rotating_file_sink-inl.h
+++ b/include/spdlog/sinks/rotating_file_sink-inl.h
@@ -112,40 +112,86 @@ SPDLOG_INLINE void rotating_file_sink<Mutex>::rotate_()
     using details::os::path_exists;
 
     file_helper_.close();
+    recover_();
+    filename_t first = calc_filename(base_filename_, 0);
+    if (!path_exists(first))
+    {
+        // if recovery found a gap and moved file(s)
+        // then there is space and no rotation
+        // is required.
+        file_helper_.reopen(true);
+        return;
+    }
+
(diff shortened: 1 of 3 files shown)


gabime:
Thanks for the PR.
> Removing the target is only necessary to replace the oldest file.

The rename_file_ function deletes the target file because according to the cpp standard `std::rename(..)` might fail if the target file already exists (“If new_filename exists, the behavior is implementation-defined”).

Possible solution would be to fix the `rename_file_()` to first rename the target file to some temp unique filename and delete it after the rename from src to target succeeded. If rename fails, restore to the orig filename and return false.

PR #2505, opened 2022-10-11
Flushblock

### LINKED ISSUE #1696: Blocking flush on async_logger
`spdlog::async_logger::flush()` is non-blocking operation. Flush is scheduled for later. 
I would like to have blocking variant that will wait untill everything that was scheduled for logging untill now is flushed by all sinks. 
My usecase is that I want to be be sure that everything I wanted to log is safe on disk and can be passed for further processing outside program. 
Frankly I don't see value in flush operation that doesn't flush NOW and only promises it will be done later. It will be done later anyway, at shutdown at latest.

chetansinghsajwan:
This pull request fixes #1696 

I have resolved the issue, async_logger now supports wait(), wait_for() and wait_until() to wait until all async dispatched logs are processed.
It also works with async_overflow_policy::overrun_oldest.
I used a counter to keep track of number of async logs remaining.
I also updated async example to show the use case.

I could not update the tests as I was not build them on my system.


gabime:
Thanks @shifu-dev . I will have to think about it. Seems non trivial and I am not sure if  it worth it the complexity. Closing for now.

PR #2509, opened 2022-10-13
Replace iterator difference with std::distance(revert #2030)

kin4stat:
motivation can be found here

https://github.com/gabime/spdlog/pull/2030#issuecomment-1276889627


gabime:
I don’t want to switch go back and forth between PRs without good reason. 
@neheb what do you think about reverting your PR due to this [reasoning](https://github.com/gabime/spdlog/pull/2030#issuecomment-1276889627)?

neheb:
No real opinion on it.

gabime:
@kin4stat You forgot the r in the return. I would appreciate if you check your pr before posting.

Issue #2512, opened 2022-10-15
compile error on Windows with external fmtlib and MSVC 17.3.6


YunchengLiu:
Thanks a lot for this excellent library, helped me a lot.
But I encountered a sudden problem. 
On Windows platform, after enabling wchar_t and external fmtlib support in spdlog, when compiling with C++20 option in MSVC 17.3.6, fmt throws a compilation error. 

spdlog ( 1.10.0#1) and fmtlib (9.1.0) are all installed via vcpkg(x64-windows) with lasted port comfigurations.

Message as follows:

```
Error C7595: "fmt:: v9:: basic_format_string<wchar_t, const tm&>:: basic_format_string": The call to an instant function is not a constant expression.
```
The wrong location is:
https://github.com/gabime/spdlog/blob/d011332616464bd0c84817d6255bf5910b7803e9/include/spdlog/sinks/daily_file_sink.h#L81

I saw the comments here, but I still tried to modify it to `fmt::runtime` , it compiles successfully at this time, it seems that the comments here are a bit outdated? Has the new version of MSVC resolved the contradiction between wchar_t and fmt?

I tried to reproduce the above problem externally with the following minimal code. On my local PC, the following code using MSVC 17.3.6 will throw the same error (not modified to `fmt::runtime` ).

```c++

#define SPDLOG_WCHAR_FILENAMES
#define SPDLOG_FMT_EXTERNAL
#include <fmt/format.h>
#include <spdlog/spdlog.h>
#include <spdlog/sinks/daily_file_sink.h>

int main(){

    auto logger = spdlog::daily_logger_mt("daily_logger", L"logs/daily.txt", 2, 30);

    logger->info("{} world", "hello");

}

```

Since I don't know about other platforms and the specific requirements for compiler versions, I would like to ask if it is possible to modify this to `fmt::runtime` to solve the problem?

tt4g:
fmt recently added support for `wcart_t *` compilation time checks (https://github.com/fmtlib/fmt/issues/3107#issuecomment-1252716537).

The fmt bundled with spdlog is version 8.x, so it does not take into account the compile check feature of the latest version.

Can you add the `SPDLOG_FMT_RUNTIME` macro to avoid the problem?
Also, if it works, could you please create a PR?

https://github.com/gabime/spdlog/blob/d011332616464bd0c84817d6255bf5910b7803e9/include/spdlog/sinks/daily_file_sink.h#L80-L84

PR #2517, opened 2022-10-17
Use 'SPDLOG_FMT_RUNTIME' to fix compilation error throwed MSVC and fmt 9.1.x

YunchengLiu:
This is the first time I submit a PR to the open source community. I try to submit code and instructions according to the rules I have learned, but there may be inappropriate places to do it. Please also point out my problems actively, I will try to do it as soon as possible. Learn and correct.

fmt 9.1.x added compile-time checks for wide strings, so when compiled with MSVC and this or later version of fmt(9.1.x), codes belows will throw a compilation error  #(https://github.com/gabime/spdlog/issues/2512).
https://github.com/gabime/spdlog/blob/d011332616464bd0c84817d6255bf5910b7803e9/include/spdlog/sinks/daily_file_sink.h#L80-L84

When compiled with `SPDLOG_WCHAR_FILENAMES` with a newer verision of fmt(9.1.x) using MSVC,  `SPDLOG_FMT_RUNTIME` will solve the problem.
```c++
#    if defined(_MSC_VER) && defined(SPDLOG_WCHAR_FILENAMES) // for some reason msvc doesn't allow fmt::runtime(..) with wchar here
#       if FMT_VERSION >= 90101
            // fmt 9.1.x added compile-time checks for wide strings
            // so when compiled with MSVC and this or later version of fmt
            // here should use 'SPDLOG_FMT_RUNTIME' to enable 'fmt::runtime(...)' for wchar to avoid compile error
            return fmt::format(SPDLOG_FMT_RUNTIME(fmt_filename), now_tm);
#       else
            return fmt::format(fmt_filename, now_tm);
#       endif
#    else
        return fmt::format(SPDLOG_FMT_RUNTIME(fmt_filename), now_tm);
#    endif
```

### DIFF
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -78,7 +78,14 @@ struct daily_filename_format_calculator
         // generate fmt datetime format string, e.g. {:%Y-%m-%d}.
         filename_t fmt_filename = fmt::format(SPDLOG_FMT_STRING(SPDLOG_FILENAME_T("{{:{}}}")), filename);
 #    if defined(_MSC_VER) && defined(SPDLOG_WCHAR_FILENAMES) // for some reason msvc doesn't allow fmt::runtime(..) with wchar here
-        return fmt::format(fmt_filename, now_tm);
+#       if FMT_VERSION >= 90101
+            // fmt 9.1.x added compile-time checks for wide strings
+            // so when compiled with MSVC and this or later version of fmt
+            // here should use 'SPDLOG_FMT_RUNTIME' to enable 'fmt::runtime(...)' for wchar to avoid compile error
+            return fmt::format(SPDLOG_FMT_RUNTIME(fmt_filename), now_tm);
+#       else
+            return fmt::format(fmt_filename, now_tm);
+#       endif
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. Seems this all can be reduced to

 ```c++
// msvc doesn't allow fmt::runtime(..) with wchar, with fmtlib versions < 9.1.0
#if defined(SPDLOG_WCHAR_FILENAMES) && defined(_MSC_VER) && FMT_VERSION < 90101
    return fmt::format(fmt_filename, now_tm);
#else 
   return fmt::format(SPDLOG_FMT_RUNTIME(fmt_filename), now_tm);
#endif


```

YunchengLiu:
Thanks. I have simplified the implement code.

YunchengLiu:
@gabime Hi, I found the ci test failed at the code below. 
https://github.com/gabime/spdlog/blob/77429b2e2e70feb5d2619374f2bc99326f6570dd/tests/test_eventlog.cpp#L44

It seems that this is not due to my PR, probably.

I want to know what should I do next, thanks.

gabime:
Thanks @YunchengLiu . Merged.

PR #2539, opened 2022-11-07
Bundle fmt 9.1.0's std.h too

eli-b:
And provide a header to include either it or the external fmt's version.

This file was added to fmt in version 9.0.0.


gabime:
Thanks. Since it is a bundled, should be the exact copy of the original, including the code formatting, which seems to be different.

eli-b:
Sorry. I copy and pasted within my IDE, which might have mangled the
formatting. I'll fix this.

eli-b:
Please review again. I've amended the commit in my branch.

gabime:
Merged. Thanks @eli-b

Issue #2485, opened 2022-09-10
current spdlog v1.x branch (b6051273) fails to build with fmt (b79ed41) or later


tycho:
Originally submitted to fmt via https://github.com/fmtlib/fmt/issues/3090 but they [think](https://github.com/fmtlib/fmt/issues/3090#issuecomment-1242753176) it's an spdlog problem. Including the text of the bug report here for convenience:

---

I just did a rebase of my project, including latest upstream fmt and spdlog. It appears that fmt broke compatibility with spdlog somehow, but my C++ template-fu is weak and I'm not sure what the error message is trying to tell me exactly:

```
In file included from code/app.cpp:1:
In file included from code/lib/universal_include.h:450:
In file included from /Users/steven/Development/game/contrib/spdlog/include/spdlog/spdlog.h:14:
/Users/steven/Development/game/contrib/spdlog/include/spdlog/logger.h:375:13: error: no matching function for call to 'vformat_to'
            fmt::detail::vformat_to(buf, fmt, fmt::make_format_args(std::forward<Args>(args)...));
            ^~~~~~~~~~~~~~~~~~~~~~~
/Users/steven/Development/game/contrib/spdlog/include/spdlog/logger.h:90:9: note: in instantiation of function template specialization 'spdlog::logger::log_<const char *>' requested here
        log_(loc, lvl, fmt, std::forward<Args>(args)...);
        ^
code/app.cpp:114:2: note: in instantiation of function template specialization 'spdlog::logger::log<const char *>' requested here
        SPDLOG_LOGGER_INFO(s_logApp, "Version: {}", Game::Version::AnnotatedVersion());
        ^
/Users/steven/Development/game/contrib/spdlog/include/spdlog/spdlog.h:318:45: note: expanded from macro 'SPDLOG_LOGGER_INFO'
#    define SPDLOG_LOGGER_INFO(logger, ...) SPDLOG_LOGGER_CALL(logger, spdlog::level::info, __VA_ARGS__)
                                            ^
/Users/steven/Development/game/contrib/spdlog/include/spdlog/spdlog.h:296:58: note: expanded from macro 'SPDLOG_LOGGER_CALL'
#define SPDLOG_LOGGER_CALL(logger, level, ...) (logger)->log(spdlog::source_loc{__FILE__, __LINE__, SPDLOG_FUNCTION}, level, __VA_ARGS__)
                                                         ^
[...]

PR #2544, opened 2022-11-11
Support compile-time format string checking with std::format

sylveon:
Now std::format and fmtlib are equivalent when using spdlog with either.


include/spdlog/details/fmt_helper.h
```diff
@@ -23,6 +27,32 @@ inline spdlog::string_view_t to_string_view(const memory_buf_t &buf) SPDLOG_NOEX
...
+    return fmt;
+}
+#elif __cpp_lib_format >= 202207L
+template<typename T, typename... Args>
+SPDLOG_CONSTEXPR_FUNC std::basic_string_view<T> to_string_view(std::basic_format_string<T, Args...> fmt) SPDLOG_NOEXCEPT
+{
+    return fmt.get();
+}
+#endif
+
```
gabime: else what if not __cpp_lib_format >= 202207L ?

include/spdlog/logger.h
```diff
@@ -17,6 +17,7 @@
 #include <spdlog/common.h>
 #include <spdlog/details/log_msg.h>
 #include <spdlog/details/backtracer.h>
+#include <spdlog/details/fmt_helper.h>
```
gabime: Lets try to avoid including this internal header . See https://github.com/gabime/spdlog/pull/2346#issuecomment-1106085031. 
Perhaps moving the “to_string_view()” functions to common.h

gabime (changes requested):

include/spdlog/logger.h
```diff
@@ -368,11 +369,7 @@ class SPDLOG_API logger
         SPDLOG_TRY
         {
             memory_buf_t buf;
-#ifdef SPDLOG_USE_STD_FORMAT
             fmt_lib::vformat_to(std::back_inserter(buf), fmt, fmt_lib::make_format_args(std::forward<Args>(args)...));
```
gabime: I think fmt lib’s back_inserter optimization is not yet released (https://github.com/gabime/spdlog/issues/2485#issuecomment-1245783205)

include/spdlog/details/fmt_helper.h
sylveon: In that case, spdlog::format_string_t resolves to std::string_view (because std::format_string didn't exist), which is handled by the other overloads I added.

include/spdlog/logger.h
sylveon: Good idea, I'll find another place for them.

include/spdlog/logger.h
sylveon: Will add back. I initially removed because the wide character version had the same code when using fmtlib and when not using it.

include/spdlog/common.h
```diff
@@ -323,6 +336,44 @@ struct file_event_handlers
...
+
+SPDLOG_CONSTEXPR_FUNC spdlog::wstring_view_t to_string_view(spdlog::wstring_view_t str) SPDLOG_NOEXCEPT
+{
+    return str;
+}
+#endif
+
+#ifndef SPDLOG_USE_STD_FORMAT
+template<typename T, typename... Args>
+inline fmt::basic_string_view<T> to_string_view(fmt::basic_format_string<T, Args...> fmt)
```
gabime, line 365: shouldn’t this be SPDLOG_CONSTEXPR_FUNC as well?

gabime (changes requested):

include/spdlog/common.h
sylveon: No because the conversion operator for fmt::basic_format_string to fmt::basic_string_view is not constexpr

gabime:
Merged.Thanks @sylveon

PR #2563, opened 2022-12-01
dup_filter_sink adds parameters to enable setting the level of skipped logs

zydxhs:
By default, dup_filter_sink uses the info level to output duplicate messages.In some scenarios, users don't care about the messages and want to keep them at a lower level so they can ignore them.

### DIFF
--- a/include/spdlog/sinks/dup_filter_sink.h
+++ b/include/spdlog/sinks/dup_filter_sink.h
@@ -41,9 +41,9 @@ class dup_filter_sink : public dist_sink<Mutex>
 {
 public:
     template<class Rep, class Period>
-    explicit dup_filter_sink(std::chrono::duration<Rep, Period> max_skip_duration, level::level_enum level = level::info)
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/dup_filter_sink.h
```diff
@@ -41,15 +41,17 @@ class dup_filter_sink : public dist_sink<Mutex>
 {
 public:
     template<class Rep, class Period>
-    explicit dup_filter_sink(std::chrono::duration<Rep, Period> max_skip_duration)
+    explicit dup_filter_sink(std::chrono::duration<Rep, Period> max_skip_duration, level::level_enum level = level::info)
```
gabime: Please provide more descriptive name to this param. something like `level::level_enum notification_level)`

gabime (changes requested):

include/spdlog/sinks/dup_filter_sink.h
zydxhs: ok

zydxhs:
Hi, param name is changed.

gabime:
Merged. Thanks @zydxhs

PR #2573, opened 2022-12-09
Fixed variable name

naddu77:
Changed the variable name from `check_ebugger_present` to `check_debugger_present`.


gabime:
Thanks, could you also fix the typo in check_debbugger_present_ ?

naddu77:
Of course it is possible.
But, it is a attribute of the class.
In many places, members of classes are appended with `_`.
Shall I fix it though?

gabime:
I mean to fix the double bb in debbuger_, not the underscore

naddu77:
Ah!! OK! I got it.

PR #2575, opened 2022-12-11
Feature/add system includes option

ghost:
Adds support for marking include directories and system headers. This stops clang tidy (and compiler warnings) from throwing warnings when including spdlog header.


gabime:
Thanks @dpayne

ghost:
that was fast, thank you!

stertingen:
I believe this should be the default if spdlog is included via add_subdirectory or FetchContent.

PR #2588, opened 2022-12-29
feat(mpmc_blocking_q): add blocking dequeue without timeout

### LINKED ISSUE #2587: tsan false warnings for mpmc_blocking_queue::dequeue_for
gcc 11.3.0 on Ubuntu 22.04 reports "data race" and "double lock of a mutex" for `mpmc_blocking_queue::dequeue_for`

This is most probably due to this gcc bug - https://gcc.gnu.org/bugzilla//show_bug.cgi?id=101978

Replacing timed [wait_for](https://github.com/gabime/spdlog/blob/edc51df1bdad8667b628999394a1e7c4dc6f3658/include/spdlog/details/mpmc_blocking_q.h#L58) with `std::condition_variable::wait` avoids the tsan warnings.

This of course should be fixed in gcc, but I am curious what is the reason for wait_for in the first place? Do you see any issues with replacing `wait_for` with `wait`?

alokpr:
Use the new blocking dequeue to avoid unnecessarily waking up the thread pool every 10s.

Fixes #2587 by replacing std::condition_variable::wait_for with std::condition_variable::wait as a workaroung for gcc 11.3 issue 101978.

### DIFF
--- a/include/spdlog/details/mpmc_blocking_q.h
+++ b/include/spdlog/details/mpmc_blocking_q.h
@@ -49,7 +49,7 @@ public:
         push_cv_.notify_one();
     }
 
-    // try to dequeue item. if no item found. wait up to timeout and try again
+    // dequeue with a timeout.
     // Return true, if succeeded dequeue item, false otherwise
     bool dequeue_for(T &popped_item, std::chrono::milliseconds wait_duration)
     {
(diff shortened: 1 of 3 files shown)


gabime:
Thanks @alokpr

PR #2589, opened 2022-12-30
Do not use LFS64 functions on linux/musl

kraj:
On musl, off_t is 64bit always ( even on 32bit platforms ), therefore using LFS64 funcitons is not needed on such platforms. Moreover, musl has stopped providing aliases for these functions [1] which means it wont compile on newer musl systems. Therefore only use it on 32bit glibc/linux platforms and exclude musl like cygwin or OSX

[1] https://git.musl-libc.org/cgit/musl/commit/?id=246f1c811448f37a44b41cd8df8d0ef9736d95f4
Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -236,8 +236,8 @@ SPDLOG_INLINE size_t filesize(FILE *f)
 #    else
     int fd = ::fileno(f);
 #    endif
-// 64 bits(but not in osx or cygwin, where fstat64 is deprecated)
-#    if (defined(__linux__) || defined(__sun) || defined(_AIX)) && (defined(__LP64__) || defined(_LP64))
+// 64 bits(but not in osx, linux/musl or cygwin, where fstat64 is deprecated)
+#    if ((defined(__linux__) && defined(__GLIBC__)) || defined(__sun) || defined(_AIX)) && (defined(__LP64__) || defined(_LP64))
     struct stat64 st;
     if (::fstat64(fd, &st) == 0)
     {


include/spdlog/details/os-inl.h
```diff
@@ -236,8 +236,8 @@ SPDLOG_INLINE size_t filesize(FILE *f)
 #    else
     int fd = ::fileno(f);
 #    endif
-// 64 bits(but not in osx or cygwin, where fstat64 is deprecated)
-#    if (defined(__linux__) || defined(__sun) || defined(_AIX)) && (defined(__LP64__) || defined(_LP64))
+// 64 bits(but not in osx, linux/musl or cygwin, where fstat64 is deprecated)
+#    if ((defined(__linux__) && defined(__GLIBC__)) || defined(__sun) || defined(_AIX)) && (defined(__LP64__) || defined(_LP64))
```
gabime, line 240: Better check for some kind of musl specifc macro to detect it. Who knows what it may break checking __GLIBC__ for this.

kraj: Glibc defined the macro long time ago and it won’t go away ever however musl does not define any libc identifying macros intentionally

gabime: This would affect any other libc vendors that dont define __GLIBC__ .

kraj: its already checking for linux, which limits the C  library choices. Glibc and uclibc already define __GLIBC__, third option is musl and another option is bionic/android which would be similar to musl, if needed it can be controlled  via android define. Do you have any other combination that spdlog supports on linux ?

gabime: There is no specific combination,  but this is bit dangerous. Who knows what libc impl this would break. According to [this](https://wiki.osdev.org/C_Library#Implementations), there are at least 11. Makes much more sense to check for musl if this is a musl fix

kraj: Let me put it differently. What this code does currently is glibc specific, LFS64 functions are not posix defined. It’s not Linux specific. So perhaps making the check to also look for glibc is the right thing here

gabime: That’s probably right. I am even thinking to remove the use of `fstat64()` entirely, since it seems that in 64 bit platforms the regular fstat() already returns the 64 version of `off_t` (including android bionic). Could you please check this assumption @kraj ?

kraj: yes. On 64bit linux and AFAICT BSDs and OSX `off_t` is 64bit

gabime:
Merged. Thanks @kraj

PR #2590, opened 2022-12-31
fix build: fix for freebsd

segoon:
The build error was:
  include/spdlog/details/tcp_client.h:106:31: error: use of undeclared identifier 'IPPROTO_TCP'


gabime:
Thanks @segoon

Issue #2469, opened 2022-08-18
Interface for notifying that spdlog drops a log?


mudream4869:
Blocking mode is dangerous for services, but when selecting overrun mode, there is a risk of losing logs.

Not sure if there is an interface (ex: pass a callback function?) when spdlog is dropping a log.

Example callback interface?:

```cpp
void func(const Log& log) noexcept {
}
```

tt4g:
There is no API to notify asynchronous log drops.
However, `spdlog::details::thread_pool::overrun_counter()` lets you know how many logs were dropped due to overruns.

https://github.com/gabime/spdlog/blob/f44fa31f5110331af196d0ed7f00ae1c7b8ef3cc/include/spdlog/details/thread_pool.h#L99-L100

dawagner:
@tt4g : I found out about `overrun_counter` and I'm using it but I'm wondering, since it is in a `details` namespace, whether I can rely on the existence of this API? Would it make sense to make it part of the "official" public API and advertise it as such?

tt4g:
@dawagner I am not the maintainer of this repository, so I cannot give an official answer.
You need to ask @gabime for the answer.

gabime:
@dawagner I don’t think it worth it to add “official” api for this, but there are no plans on changing this in version 1.x.

dawagner:
ok, thanks for the answer!

Issue #1328, opened 2019-11-26
Callback on error level log


fourier:
Hi,

I want to add some special behavior to my application which uses spdlog: I want to have a callback after the message of log level "error" has been written.
Most of the time I want to gracefully shutdown the application and do not want to add special lines of code all the places where the ->error could be called.
Is there an easy way to achieve this? Important is that the error entry still has to be written (and flushed) to the file.

gabime:
You could develop a custom sink that does this when an error is logged, and add this to the end of the sinks vector used by your logger

fourier:
Thanks! Sounds like a plan.

Issue #1016, opened 2019-03-06
Question: can i create a callback function for when something is logged?


GeestWagen:
basically i am trying to get something like this
`
void callbackFunction(string message)
{
/*code*/
}
` 
and that function should be called when i do this: `spdlog::info("message");`

i haven't figured a way to do this since quite unfamiliar with this library. thanks in advance

gabime:
There is no obvious way for this. You could imlement a custom sink that do this but I doubt it will worth the effort.

Issue #863, opened 2018-10-05
logging to a remote server like node JS


naveenpinto:
Hi 
   this is second time i am asking this question, do you have any examples on writing to a remote server, example a node JS server.

Regards
Naveen

vvhh2002:
hi, you can write a sub class of base_sink with a callback  of NodeJS js function.

naveenpinto:
Hi @vvhh2002 thanks a lot, do you have any example implementation of the same????

PR #2610, opened 2023-01-16
Add a trivial callback sink

maghorbani:
I needed such a functionality, to receive logs in a callback function, so I added this sink to the library.
The new sink, will call a custom callback, each time something is logged. 
The code was initially cloned from `basic_file_sink.h`. 
Required test, example and readme sample, added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,34 @@ void multi_sink_example()
 ```
 
 ---
+#### Logger with a custom callback function that receives the logs
+```c++
+
+// create logger with a lambda function callback, the callback will be called
+// each time something is logged to the logger
+void callback_example()
+{
+    spdlog::custom_log_callbacks callbacks;
+    callbacks.on_log_formatted = [](std::string msg){
+        // for example you can be notified by sending an email to yourself
+    };
+    // or you can catch the log_msg object
+    // callbacks.on_log = [](spdlog::details::log_msg msg){
+    // };
+    
+    auto callback_sink = std::make_shared<spdlog::sinks::callback_sink_mt>(callbacks);
+    callback_sink->set_level(spdlog::level::err);
+
+    auto console_sink = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
+    spdlog::logger logger("custom_callback_logger", {console_sink, callback_sink});
+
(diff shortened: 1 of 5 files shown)


gabime:
Thanks @maghorbani . I am not sure. Whats exactly the use case here? This can be achieved already by implementing  custom sink just as easily.

maghorbani:
Well I needed it to be like this, so my custom sink would not be an additional patch to spdlog in every single project of mine. I just thought that, it is not harmful to have a general purpose  sink. 
Anyway merging it or not, that's your decision, but I will keep my fork sync with the main repo and use it in my projects that use spdlog.  I saw an issue or two, asking for the same kind of sink.

gabime:
If there a popular demand for such think, I don't mind  adding this. Which issues mention this?
As a side note, I would make it simpler for use by getting rid of the struct and just providing 2 constructor overloads that accept the lambdas. 
Also note that the conversion to string in the second lambda is not cheap since string allocates. Better to reuse a private member string.

maghorbani:
here are some that I found, searching `callback` keyword in issues:
[2469](https://github.com/gabime/spdlog/issues/2469), [1328](https://github.com/gabime/spdlog/issues/1328), [1016](https://github.com/gabime/spdlog/issues/1016)
or [863](https://github.com/gabime/spdlog/issues/863), they did not know how to implement a new sink, but such a callback sink would meet their need
and an old one (i just saw it) [205](https://github.com/gabime/spdlog/pull/205), they implemented such a callback, but it was not general enough to be merged.

I'm not sure if it worth it to be merged, if so, please let me know, and I will improve the new sink to meet your reasonable advice: 
> As a side note, I would make it simpler for use by getting rid of the struct and just providing 2 constructor overloads that accept the lambdas.<br>Also note that the conversion to string in the second lambda is not cheap since string allocates. Better to reuse a private member string.

gabime:
> I'm not sure if it worth it to be merged

Lets start with a very basic, single lambda callback (the one with the the log_msg arg).
This I think should cover most of the cases. 
If more functionality needed, it can be added later.

Thanks!

gabime:
Thanks @maghorbani . Merged.

PR #2619, opened 2023-01-24
Add optional TID definition to the systemd sink send

chardin-cpi:
From the systemd.journal-fields the TID is a user defined field passed directly from the clients and stored in the journal. Adding the arguement in the journal send to support that storage option in the journal.

### DIFF
--- a/include/spdlog/sinks/systemd_sink.h
+++ b/include/spdlog/sinks/systemd_sink.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <spdlog/sinks/base_sink.h>
+#include <spdlog/details/os.h>
 #include <spdlog/details/null_mutex.h>
 #include <spdlog/details/synchronous_factory.h>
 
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/systemd_sink.h
```diff
@@ -75,11 +77,17 @@ class systemd_sink : public base_sink<Mutex>
         {
             // Note: function call inside '()' to avoid macro expansion
             err = (sd_journal_send)("MESSAGE=%.*s", static_cast<int>(length), payload.data(), "PRIORITY=%d", syslog_level(msg.level),
+#ifndef SPDLOG_NO_THREAD_ID
+                "TID=%jd", static_cast<intmax_t>(gettid()),
```
gabime: please use os::thread_id() to get the thread id

gabime (changes requested):

gabime:
Thanks @chardin-cpi

PR #2631, opened 2023-02-06
implements minutely_file_sink

seker:
my app create a very big size log file(400M+) while using hourly_file_sink, so it's helpfull if we can rotate log file according to minutes (for example 10 minutes).


gabime:
It is recommended to use rotating sink to control the log file size.Closing since I don't want or have  much duplicated code with hourly/daily sinks.

PR #2636, opened 2023-02-09
Allow basic_file_sink to open a file in text mode

N-Dekker:
Added an optional `bool` parameter to the explicit `basic_file_sink`
constructor, allowing to open the file in text mode, instead of in binary
mode.

Enables automatic conversion of '\n' characters _inside_ a message string to a
platform specific eol representation. (`\r\n` on Windows.)

### DIFF
--- a/include/spdlog/sinks/base_sink-inl.h
+++ b/include/spdlog/sinks/base_sink-inl.h
@@ -53,7 +53,16 @@ void SPDLOG_INLINE spdlog::sinks::base_sink<Mutex>::set_formatter(std::unique_pt
 template<typename Mutex>
 void SPDLOG_INLINE spdlog::sinks::base_sink<Mutex>::set_pattern_(const std::string &pattern)
 {
-    set_formatter_(details::make_unique<spdlog::pattern_formatter>(pattern));
+    const auto current_pattern_formatter = dynamic_cast<spdlog::pattern_formatter *>(formatter_.get());
+
+    if (current_pattern_formatter)
+    {
+        current_pattern_formatter->set_pattern(pattern);
+    }
+    else
+    {
+        set_formatter_(details::make_unique<spdlog::pattern_formatter>(pattern));
+    }
 }
 
 template<typename Mutex>
--- a/include/spdlog/details/file_helper-inl.h
+++ b/include/spdlog/details/file_helper-inl.h
@@ -29,13 +29,13 @@ SPDLOG_INLINE file_helper::~file_helper()
     close();
 }
 
-SPDLOG_INLINE void file_helper::open(const filename_t &fname, bool truncate)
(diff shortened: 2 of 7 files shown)


gabime:
Thanks for the pull. However this can be fixed  using custom formatter flag or custom sink, so I’d rather not to add it (and also break the ABI of the basic file sink ctor).

include/spdlog/sinks/base_sink-inl.h
```diff
@@ -53,7 +53,16 @@ void SPDLOG_INLINE spdlog::sinks::base_sink<Mutex>::set_formatter(std::unique_pt
...
+    const auto current_pattern_formatter = dynamic_cast<spdlog::pattern_formatter *>(formatter_.get());
+
+    if (current_pattern_formatter)
+    {
+        current_pattern_formatter->set_pattern(pattern);
+    }
+    else
+    {
+        set_formatter_(details::make_unique<spdlog::pattern_formatter>(pattern));
+    }
```
N-Dekker, line 65: @gabime I see you rejected my pull request, unfortunately, but could you possibly still consider this change? When a sink already has a pattern_formatter, why not call `set_pattern(pattern)` on its current formatter, instead of creating a new one?

gabime: Why is it needed?

N-Dekker: It preserves the original pattern_formatter properties of the sink, when a new pattern is set. Specifically, it preserves the original pattern_formatter::eol_ string: 
https://github.com/gabime/spdlog/blob/927cc29444a294d76e83dfb898e797dc431ce094/include/spdlog/pattern_formatter.h#L104-L111

gabime: This dynamic cast can be problematic(I think rtti would be required on some setups). 
You can use  `set_formatter(new_formatter)` instead.

N-Dekker:
> also break the ABI of the basic file sink ctor

This could be avoided by leaving the original basic file sink ctors unchanged, and adding a new ctor overload with the proposed extra `text_mode` parameter, right?

gabime:
Perhaps, but still, I prefer to to make such a change.

include/spdlog/sinks/base_sink-inl.h
N-Dekker: > This dynamic cast can be problematic

OK, I understand, on platforms that don't have rtti, `dynamic_cast` could be a problem. Still it seems better to me if `base_sink<Mutex>::set_pattern_` could only just set the pattern string of its `formatter_`, while leaving the other `pattern_formatter` data unchanged. Anyway, that's just my impression.

Issue #2638, opened 2023-02-10
MinGw: Fail to build example with udp sink 


zEuS0390:
I'm using MinGW GCC on Windows. When I'm building the examples, I'm getting a linker error:
```
[ 50%] Building CXX object CMakeFiles/example.dir/example.cpp.obj
[100%] Linking CXX executable example.exe
f:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles\example.dir/objects.a(example.cpp.obj):example.cpp:(.text$_ZN6spdlog7details10udp_client13init_winsock_Ev[_ZN6spdlog7details10udp_client13init_winsock_Ev]+0x20): undefined reference to `__imp_WSAStartup'
f:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles\example.dir/objects.a(example.cpp.obj):example.cpp:(.text$_ZN6spdlog7details10udp_client13init_winsock_Ev[_ZN6spdlog7details10udp_client13init_winsock_Ev]+0x3c): undefined reference to `__imp_WSAGetLastError'
f:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles\example.dir/objects.a(example.cpp.obj):example.cpp:(.text$_ZN6spdlog7details10udp_client8cleanup_Ev[_ZN6spdlog7details10udp_client8cleanup_Ev]+0x26): undefined reference to `__imp_closesocket'
f:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles\example.dir/objects.a(example.cpp.obj):example.cpp:(.text$_ZN6spdlog7details10udp_client8cleanup_Ev[_ZN6spdlog7details10udp_client8cleanup_Ev]+0x3a): undefined reference to `__imp_WSACleanup'
[...]

PR #2642, opened 2023-02-11
Fix MinGW build issue on example

zEuS0390:
Fix build issue #2638 on example by adding ```ws2_32``` in these two target_include_directories if the type of compiler is from MinGW. 
```cmake
target_link_libraries(spdlog PUBLIC Threads::Threads $<$<BOOL:${MINGW}>: ws2_32>)
target_link_libraries(spdlog_header_only INTERFACE Threads::Threads $<$<BOOL:${MINGW}>: ws2_32>)
```
I also made some little changes to "udp_client-windows.h" which uses pragma comment macros only when the Visual C++ compiler is used.
```c++
#if defined(_MSC_VER)
#	pragma comment(lib, "Ws2_32.lib")
#	pragma comment(lib, "Mswsock.lib")
#	pragma comment(lib, "AdvApi32.lib")
#endif
```

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -12,7 +12,7 @@ endif()
 # Example of using pre-compiled library
 # ---------------------------------------------------------------------------------------
 add_executable(example example.cpp)
-target_link_libraries(example PRIVATE spdlog::spdlog $<$<BOOL:MINGW>:ws2_32>)
+target_link_libraries(example PRIVATE spdlog::spdlog $<$<BOOL:${MINGW}>:ws2_32>)
 
 # ---------------------------------------------------------------------------------------
 # Example of using header-only library
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -12,7 +12,7 @@ endif()
 # Example of using pre-compiled library
 # ---------------------------------------------------------------------------------------
 add_executable(example example.cpp)
-target_link_libraries(example PRIVATE spdlog::spdlog)
+target_link_libraries(example PRIVATE spdlog::spdlog $<$<BOOL:MINGW>:ws2_32>)
 
 # ---------------------------------------------------------------------------------------
 # Example of using header-only library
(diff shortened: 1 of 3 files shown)


gabime:
Thanks. Please move the cmake change to the example/CMakeLists.txt. The main cmake doesn't deal with the udp sink.

zEuS0390:
Ok, got it.

example/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ endif()
 # Example of using pre-compiled library
 # ---------------------------------------------------------------------------------------
 add_executable(example example.cpp)
-target_link_libraries(example PRIVATE spdlog::spdlog)
+target_link_libraries(example PRIVATE spdlog::spdlog $<$<BOOL:${MINGW}>:ws2_32>)
```
tt4g, line 15:
```suggestion
target_link_libraries(example PRIVATE spdlog::spdlog $<$<PLATFORM_ID:MinGW>:ws2_32>)
```

You can also use CMake generator-expression `$<PLATFORM_ID:MinGW>`. This is just a suggestion since the result is the same.

zEuS0390: Thanks! I tried it, but it didn't work. Maybe because of the names mentioned in [CMake](https://github.com/Kitware/CMake/blob/323f5feb93a764bf46698958b871adc3e76b2270/Modules/CMakeDetermineSystem.cmake#L5-L31) and in the [documentation](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#platform).

I found out that there's a shorter version, but I think ```$<BOOL:...>``` is much safer.
```cmake
target_link_libraries(example PRIVATE spdlog::spdlog $<${MINGW}:ws2_32>)
```

tt4g: It is odd that it does not work since it has to be defined in the CMake v3 source.

https://github.com/Kitware/CMake/blob/01a25d72c69b0a9a60c1a62939d6a2c543448118/Modules/CMakePlatformId.h.in#L14-L15

Well, the result is the same, so it is fine as it is now.
I just thought it would be better to unify with CMake generator-expression.

zEuS0390: Yeah, strange. I'm using cmake version 3.24.2.

gabime:
Thanks @zEuS0390

Issue #1675, opened 2020-09-18
stdout sink produces extra carriage returns on Windows


chris-t-w:
Given the following example:

```
#include "spdlog/spdlog.h"
#include "spdlog/sinks/stdout_sinks.h"
#include "spdlog/sinks/stdout_color_sinks.h"

int main()
{
    std::shared_ptr<spdlog::logger> logger1 = spdlog::stdout_logger_st("one");
    logger1->set_pattern("%v");
    logger1->info("hello");

    std::shared_ptr<spdlog::logger> logger2 = spdlog::stdout_color_st("two");
    logger2->set_pattern("%v");
    logger2->info("world");

    return 0;
}
```

If we run this program on Windows and redirect the output to a file (`example.exe > output.txt`) and have a look at this output in a hex editor, we see the following:

```
68 65 6C 6C 6F 0D 0D 0A 77 6F 72 6C 64 0D 0A
h  e  l  l  o  \r \r \n w  o  r  l  d  \r \n
```

i.e. `"hello\r\r\nworld\r\n"` - notice the additional carriage return after "hello".

I believe the first example can be explained because:

1. `stdout_logger_st` uses `fwrite` to write to the `stdout` handle
2. The formatter automatically appends `\r\n` by default
3. `stdout` is opened in text mode according to: https://docs.microsoft.com/en-us/cpp/c-runtime-library/text-and-binary-mode-file-i-o ("_The stdin, stdout, and stderr streams always open in text mode by default_")
4. Streams opened in text mode will automatically replace `\n` with `\r\n` according to: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fwrite ("_If stream is opened in text mode, each line feed is replaced with a carriage return-line feed pair_")

This is a problem because some programs interpret the `\r\r\n` as two lines, resulting in all of the logs appearing with extra blank lines between them.

The only thing I can't explain is why the problem doesn't seem to happen with the color sink, which you can see from the output above produces the correct `"world\r\n"`.

gabime:
Thanks for the detailed analysis. Good to know.

> The only thing I can't explain is why the problem doesn't seem to happen with the color sink..

That't because it the windows color sink uses ```::WriteConsoleA``` instead of fwrite ([see here](https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/sinks/wincolor_sink-inl.h#L138)).

I guess the way to fix this is to use ```WriteFile``` instead under Windows in this sink..

[...]

Issue #63, opened 2015-03-18
travis-ci unit tests


ido:
@gabime:
Would you consider writing some unit/regression tests and a `.travis.yml` file so that they run on each git push?  I had to fix a few minor bugs last time I did a `git subtree pull`.  I'd especially like to see the `bench` program run on each `git push` on travis-ci so we can easily see if a commit introduces any performance regressions.

Here is an example C++11 project `.travis.yml` file you could base yours off of, if you've never used travis-ci before, it's super straightforward to get started:
https://github.com/zaphoyd/websocketpp/blob/master/.travis.yml

I do not have time to do this myself right now or I'd submit this as a pull request.  If I have time this summer, I'll add some performance/unit tests as well, but maybe you could start with lightly modified versions of the example or bench programs that output the max message rate/sec, time to log 1,000,000 entries with each of the various logging methods (sync st file, async mt file, async st file, etc.)?  Then move on to unit tests of the public API?

If you don't have time, thanks again for taking your time to write this library in the first place. :-)

gabime:
Why modify the output? It already provides most of needed info..

ido:
No need to modify the output.  

Just set it up to build and run with gcc, clang, and any other compilers you want.  Don't forget to add any relevant compile/linker flags to your build matrix.  Good luck! :-)

gnzlbg:
Here is range-v3's travis-ci script: https://github.com/ericniebler/range-v3/blob/master/.travis.yml

It is really useful to have travis ci because it tests pull-requests automatically with all the compilers you want and makes sure that at least whats tested doesn't break on any platform.

By default you get linux + clang-3.4 ... clang-3.7, and gcc 4.9, with both libc++ and libstdc++. 
You can also apply for macos build bots to test latest AppleClang + libc++.

gnzlbg:
Setting this up takes half an hour and would have prevented issue #94 .

gabime:
Right. I never had the time to get to it, Would really appreciate a pull

gnzlbg:
A pull-request is coming, it would be helpful if you registered on https://travis-ci.org and enabled testing for your project :)

gabime:
OK. Will do

gnzlbg:
How do you choose in the examples between Debug and Release builds?

gabime:
```make debug``` vs ```make```

gabime:
I registered. Please tell me if you need anything else

gnzlbg:
[...]

PR #2646, opened 2023-02-20
Fix stdout_sink_base::log's behavior inconsistency

scc-tw:
It will flush every time when it if not defined _WIN32, but not in Windows family.
We viewed the commit https://github.com/gabime/spdlog/commit/48d4ed9bc0179792d24621572a9fe40fe9080ebe for fixing issue #1675 . It seems missing this flushing operation in mistake.

### DIFF
--- a/include/spdlog/sinks/stdout_sinks-inl.h
+++ b/include/spdlog/sinks/stdout_sinks-inl.h
@@ -68,6 +68,7 @@ SPDLOG_INLINE void stdout_sink_base<ConsoleMutex>::log(const details::log_msg &m
     {
         throw_spdlog_ex("stdout_sink_base: WriteFile() failed. GetLastError(): " + std::to_string(::GetLastError()));
     }
+    FlushFileBuffers(handle_); // flush every line to terminal
 #else
     std::lock_guard<mutex_t> lock(mutex_);
     memory_buf_t formatted;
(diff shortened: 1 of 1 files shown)


gabime:
please use fflush in both cases.

scc-tw:
@gabime I moved the ffush out of the ifdefine.

gabime:
@25077667 After closer examination I think the original behavior is consistent in both platforms since the windows impl use WriteFile which means no libc memory buffer to be fflushed. Note that fflush only flush the libc buffer. It doesnt call fsync. So I think this PR is not needed.

scc-tw:
@gabime Well, I found this issue from our team's CI.

We used gtest framework to capture the stdout stream. We always could capture the output from **SPDLOG_ERROR** in  Linux and MacOS, but not Windows "sometime".

After I traced code to here, I add `std::endl` to our code and it fixed the flushing issue.

Honestly speaking, I'm not quite sure what's the behavior after invoke the Window API.
But we knew that the `fflush` would send data to the kernel.

gabime:
If you are ***absolutely*** sure that fflush fixes the issue I don’t mind to add it (and also remove the call to  fflush in line #63)

scc-tw:
@gabime I believe we can keep an eye on things for a week and report back.

scc-tw:
Hello @gabime, 

It was monitored for a week, and over 40 runs were completed successfully (never trigger any flushing bug issue).
It is a solid remedy, in my opinion.

scc-tw:
Hello @gabime ,

After a month, we tested over 200 times via our CI.
I think this result could provide a solid remedy.

gabime:
Ok. I reopen the pr. Please remove the the redundant  `fflush(..)` in line 63 since it won't be needed anymore.

scc-tw:
@gabime ,
Is it possible for you to review my changes?

gabime:
Merged. Thanks @25077667

Issue #2648, opened 2023-02-22
spdlog::initialize_logger undesired effect on patterns


afshinpir:
Hi all

I have a small code with 1 logger and 2 sinkers with different pattern on sinkers:
```
auto create_logger() {
    spdlog::cfg::load_env_levels();

    std::vector<spdlog::sink_ptr> sinks;

    auto console1_sink = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
    console1_sink->set_level(spdlog::level::info);
    console1_sink->set_pattern("[%n] %v");
    sinks.push_back(console1_sink);

    auto console2_sink = std::make_shared<spdlog::sinks::stderr_color_sink_mt>();
    console2_sink->set_level(spdlog::level::trace);
    sinks.push_back(console2_sink);

    auto logger = std::make_shared<spdlog::logger>(GGL_PROVIDER_LOG_NAME, std::begin(sinks), std::end(sinks));
    spdlog::initialize_logger(logger);
    return logger;
}
```
now if you log anything with `logger` like this:
```
logger->info("test");
```
You will see default pattern on output of both sinkers. Why? Because `spdlog::initialize_logger` applies pattern on loggers(which is default) and these patterns will be applied to sinkers overriding their patterns. I used `spdlog::initialize_logger` only in order to apply `SPDLOG_LEVEL` values from command line, but this undesired behavior is very bad. I think I have a workaround by setting patterns after `spdlog::initialize_logger(logger)`, but I don't think this is a desired approach.

tt4g:
The description of `spdlog::initialize_logger()` clearly states that it changes the formatter. Its behavior cannot be easily changed.

https://github.com/gabime/spdlog/blob/da14258533cb951ce85087ceb45556e0b8253660/include/spdlog/spdlog.h#L39-L47

If you are calling `spdlog::initialize_logger()` to apply the log levels loaded by `spdlog::cfg::load_env_levels()`, the formatter can only be changed after `spdlog::initialize_logger()`.

afshinpir:
@tt4g Since you normally create `spdlog::logger` after creating sinkers and setting up their configuration, I think this behavior is somehow not great. 
I think it is better to have another method that only applies levels to loggers directly based on `SPDLOG_LEVEL` environment variable too.

tt4g:
If you think there is demand, please open a PR to add an API that allows only log levels to be set.
[...]

PR #2649, opened 2023-02-23
Added `apply_logger_env_levels`

afshinpir:
This method applies levels which is set by environment variable `SPDLOG_LEVEL` to the a single logger. Useful for loading configuration into manually created loggers.

### DIFF
--- a/include/spdlog/spdlog.h
+++ b/include/spdlog/spdlog.h
@@ -131,6 +131,15 @@ SPDLOG_API spdlog::logger *default_logger_raw();
 
 SPDLOG_API void set_default_logger(std::shared_ptr<spdlog::logger> default_logger);
 
+// Initialize logger level based on environment configs.
+//
+// Useful for applying SPDLOG_LEVEL to manually created loggers.
+//
+// Example:
+//   auto mylogger = std::make_shared<spdlog::logger>("mylogger", ...);
+//   spdlog::apply_logger_env_levels(mylogger);
+SPDLOG_API void apply_logger_env_levels(std::shared_ptr<logger> logger);
+
 template<typename... Args>
 inline void log(source_loc source, level::level_enum lvl, format_string_t<Args...> fmt, Args &&... args)
 {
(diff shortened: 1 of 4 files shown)


tt4g:
Fix #2648

include/spdlog/spdlog.h
```diff
@@ -46,6 +46,15 @@ inline std::shared_ptr<spdlog::logger> create(std::string logger_name, SinkArgs
 //   spdlog::initialize_logger(mylogger);
 SPDLOG_API void initialize_logger(std::shared_ptr<logger> logger);
```
gabime: Should  be declared somewhere at the bottom

gabime (changes requested):

include/spdlog/spdlog.h
afshinpir: I moved to after `set_default_logger`. I also did the same in other files too.

gabime:
Thanks @afshinpir

tt4g:
@afshinpir Sorry about this after merged. I forgot to submit my review.
Could you please open PR to remove `std::lock_guard<std::mutex> lock(logger_map_mutex_);` in `registry::apply_logger_env_levels()`?
This is because `logger_map_mutex_`  is `std::mutex` to protect `loggers_` variables.

https://github.com/gabime/spdlog/blob/51bcff820e40a6d26c684284509a9fcf27b10c0a/include/spdlog/details/registry-inl.h#L290-L296

afshinpir:
@tt4g I can open a new PR, but since `log_levels_` is a map, so we may need to have excusive access to it for thread safety. Since `registry::set_levels` and `registry::initialize_logger` both use this mutex, I used the same one. So this mutex currently protects access to both `log_levels_` and `loggers_`. I think it is better to have it like this rather than adding new critical section. or I can just remove it and assume that `set_levels` and `apply_logger_env_levels` will never be called at same time. Please let me know what to do.

tt4g:
@afshinpir I agree with your thoughts.
Because of the variable names I was unaware that `logger_map_mutex_` protected `loggers_` and `log_levels_`.
There would be no need to open a PR.

Issue #2012, opened 2021-07-25
MSVC sink will output wrong result when string contain non-ASCII character


Demonese:
Some project detail:
* all source files using utf8 encoding and compiled with /utf-8 option
* some string contain Chinese or Japenese character
* spdlog disable wchar support (because all in utf8)
In VS2019 debugger output window, Chinese or Japenese character will display in wrong way. I found a OutputDebugStringA calling in msvc_sink.h but this method seems doesn't support unicode encoding (instead, we should using OutputDebugStringW, but it recive wchar string, converting might be expensive?).

tt4g:
I think the reason is that the character encoding of the debugger console is not UTF-8 (or does not support it).
It should not be garbled in a console that supports UTF-8.

> but it recive wchar string, converting might be expensive?

Probably yes.

tt4g:
Do you have the compile option "/utf8" set (can use `/source-charset:utf-8` and `/execution-charset:utf-8` instead)?
If these options are not set, the Visual Studio compiler will not print UTF-8 characters.

https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-160

As described on the `/execution-charset` option page, the Visual Studio compiler will not detect character codes other than UTF-16 and UTF-8 with BOM in the source code.

https://docs.microsoft.com/en-us/cpp/build/reference/execution-charset-set-execution-character-set?view=msvc-160

Demonese:
> I think the reason is that the character encoding of the debugger console is not UTF-8 (or does not support it).
> It should not be garbled in a console that supports UTF-8.

Yeah, VS2019 debug output window only support OutputDebugStringW (UTF16-LE).

![image](https://user-images.githubusercontent.com/57629614/126889832-fd4567d0-1311-4256-b921-ff8f8759aa81.png)
![image](https://user-images.githubusercontent.com/57629614/126889849-2cdf9567-578f-44c0-aca4-5ddd4fbb86bc.png)
![image](https://user-images.githubusercontent.com/57629614/126889844-8847cbb8-48b0-45c4-8482-bb964a32d3a6.png)

tt4g:
`OutputDebugStringW` converts the UTF-16 string to the encoding of the system locale, and then passes it to `OutputDebugStringA`.

https://docs.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-outputdebugstringa

[...]

PR #2651, opened 2023-02-23
msvc_sink: support utf8

Demonese:
Here's a workaround I'm using in my own project, about this issue: #2012.

This time I added an option for developers to decide whether to enable this feature.

I think this is very helpful for users who use Visual Studio but cannot modify the built-in debug output window encoding. I believe this improvement has reason to be merged.


gabime:
Thanks. I don't think new cmake variable is needed. We can use the SPDLOG_WCHAR_TO_UTF8_SUPPORT  variable, like in https://github.com/gabime/spdlog/blob/45b83a776d495dfed73b6e0950ce50f4e7f60d18/include/spdlog/sinks/win_eventlog_sink.h#L239

gabime:
Thanks @Demonese

PR #2653, opened 2023-02-25
Fixed issue where msvc_sink did not correctly output Unicode when defining SPDLOG_WCHAR_TO_UTF8_SUPPORT macro.

naddu77:
We should use OutputDebugStringW, instead of OutputDebugStringA.


gabime:
So is it instead of pr #2651?

naddu77:
I didn't realize there was already a PR to solve the same issue.
I don't think this PR needs to be merged.

PR #2655, opened 2023-02-28
feat(kafka_skin.h): kafka log support

vbirds:
add kafka log support

### DIFF
--- a/include/spdlog/sinks/kafka_skin.h
+++ b/include/spdlog/sinks/kafka_skin.h
@@ -1,0 +1,139 @@
+// Copyright(c) 2015-present, Gabi Melman & spdlog contributors.
+// Distributed under the MIT License (http://opensource.org/licenses/MIT)
+
+#pragma once
+
+//
+// Custom sink for kafka
+// Building and using requires librdkafka library.
+// For building librdkafka library check the url below
+// https://github.com/confluentinc/librdkafka
+//
+
+#include <spdlog/common.h>
+#include "spdlog/details/log_msg.h"
+#include "spdlog/sinks/base_sink.h"
+#include "spdlog/details/synchronous_factory.h"
+#include "spdlog/details/null_mutex.h"
+#include "spdlog/async.h"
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/kafka_skin.h
```diff
@@ -0,0 +1,139 @@
...
+        if (producer_ != nullptr)
+        {
+            producer_->flush(config_.flush_timeout_ms);
+        }
+    }
+
+protected:
+    void sink_it_(const details::log_msg &msg) override
+    {
+        if (producer_ != nullptr)
```
gabime: seems producer can never be null according to the ctor, so no need  check if null

gabime (changes requested):

include/spdlog/sinks/kafka_skin.h
```diff
@@ -0,0 +1,139 @@
...
+        }
+        catch (const std::exception &e)
+        {
+            throw_spdlog_ex(fmt_lib::format("error create kafka instance: {}", e.what()));
+        }
+      }
+
+    ~kafka_skin()
+    {
+        if (producer_ != nullptr)
```
gabime: seems producer can never be null according to the ctor, so no need check if null

gabime (changes requested):

include/spdlog/sinks/kafka_skin.h
vbirds: Thank you for the corrections，I will fix it

include/spdlog/sinks/kafka_skin.h
vbirds: Thank you for the corrections，I will fix it

gabime:
Thanks @vbirds

SpriteOvO:
@vbirds Is the word `skin` intended? Or just a typo of `sink`?

gabime:
Thanks for noticing. Fixed in 2d57e3b

PR #2656, opened 2023-02-28
Added Catch v3 support

xvitaly:
Added Catch v3 support with backward compatibility with Catch v2.


gabime:
Thanks. I prefer to keep it simpler by just upgrading the bundled version to catch3. If this works with current tests, we can avoid any cmake changes and ifdefs.

xvitaly:
> If this works with current tests, we can avoid any cmake changes and ifdefs.

It builds, but the tests fails:

```
Test project /builddir/build/BUILD/spdlog-1.11.0/redhat-linux-build
    Start 1: spdlog-utests
1/1 Test #1: spdlog-utests ....................Subprocess aborted***Exception:   0.07 sec
spdlog-utests: src/catch2/catch_test_case_info.cpp:147: Catch::TestCaseInfo::TestCaseInfo(Catch::StringRef, const Catch::NameAndTags&, const Catch::SourceLineInfo&): Assertion `inTag' failed.


0% tests passed, 1 tests failed out of 1
```

Also catch3 is no longer a header-only library. It need to be compiled first.

xvitaly:
https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md

gabime:
So I think we need to use their recommended approach #2

xvitaly:
> So I think we need to use their recommended approach 2

I will update my PR.

xvitaly:
Done. It will use FetchContent() for fetching catch.

tests/CMakeLists.txt
```diff
@@ -13,6 +13,20 @@ if(PkgConfig_FOUND)
     pkg_check_modules(systemd libsystemd)
 endif()
 
+find_package(Catch2 3.0.0 QUIET)
```
tt4g:
```suggestion
find_package(Catch2 3 QUIET)
```

If you are only interested in the major version, you do not need to specify a minor version.

See: https://cmake.org/cmake/help/v3.26/command/find_package.html#config-mode-version-selection

xvitaly:
Missing Git on CI. I will add it.

gabime:
@xvitaly Sorry, I had to revert this pr, it doens't pass the CI. I realized this only after the merge. Feel free to fix and open new PR.

xvitaly:
> I had to revert this pr, it doens't pass the CI. I realized this only after the merge

#2659 will prevent issues like this from happening in the future.

PR #2661, opened 2023-03-01
Added Catch v3 support

xvitaly:
Added Catch v3 support. Second attempt.


xvitaly:
Release configurations builds fine, but Debug fails.

xvitaly:
It fails on this line: https://github.com/catchorg/Catch2/blob/devel/src/catch2/catch_test_case_info.cpp#L147

gabime:
It seems that some tests has bad tags with extra closing ]]

see https://github.com/search?q=repo%3Agabime%2Fspdlog+%5D%5D%22&type=code

xvitaly:
> It seems that some tests has bad tags with extra closing ]]

Yep. I will try to fix that.

xvitaly:
@gabime It works now. Thanks for the pointer.

gabime:
Thanks @xvitaly . Merged.

gabime:
@xvitaly Unfortunately the windows appveyor CI fails now. Could you please take a look?

xvitaly:
> Unfortunately the windows appveyor CI fails now. Could you please take a look?

Typical Windows problem - missing Catch2Main.dll near spdlog-utests.exe.

xvitaly:
We need to copy compiled `Catch2Main.dll` to `spdlog-utests.exe` if `BUILD_SHARED` is enabled.

xvitaly:
I will try to fix that in a separate PR.

gabime:
Thanks

PR #2662, opened 2023-03-01
[WIP] Fixed AppVeyor build

xvitaly:
Copy all compiled DLLs to correct destinations.


xvitaly:
@gabime Visual Studio 14 2015 can't build catchv3, so these workers should be disabled. All other works as expected.

PR #2668, opened 2023-03-08
Fix os namespace in systemd_sink.h

MyNameIsLame:
#2619  added `os::thread_id()` function calls to `spdlog/sinks/systemd_sink.h`.

But access to `os::thread_id()` is not valid, as the full namespace is `spdlog::details::os`, while the current code is in the namespace `spdlog::sinks` and thus the namespace `os::` is not directly accessible.

As a side note: The [test_systemd.cpp](https://github.com/gabime/spdlog/blob/v1.x/tests/test_systemd.cpp) is currently not built in the pipeline, because it fails to find systemd in the cmake configuration step. Mabye it is sufficient to install the `libsystemd-dev` package in the container?


gabime:
Thanks. I think `details::os::` is enough and the `spdlog::` can be omitted.
Regarding the fix to the CI, PR is welcome.

MyNameIsLame:
Removed the `spdlog::` prefix and also created #2669

gabime:
Thanks @MyNameIsLame

PR #2680, opened 2023-03-21
setting the cmake standard to 20 when using std format

bachittle:
when SPDLOG_USE_STD_FORMAT is turned on the cmake standard is still set to 11 and the code does not compile. This fixes it.


gabime:
Thanks @bachittle

PR #2686, opened 2023-03-24
Seperate std string view

bachittle:
I need to use std::string_view but currently it is coupled to std::format, which isn't as supported by compilers yet (GCC 13 has std::format but most people don't have GCC 13. std::string_view is much more supported, since GCC 7). 

So I decoupled the two and got all permutations to compile. I make std::format require std::string_view option in order to work correctly.


bachittle:
build failed because appveyor doesn't have the new flag. For compatibility I think I will just automatically turn on the flag when std_format is set to on.

gabime:
Not sure I understand why is it needed. Could you please provide a code example of the problem?

bachittle:
Sure. Let's say, for whatever reason, the user is using std::string_view and not the built in string_view_t, and is trying to print it with spdlog. Or if say they are using an external string view such as boost::string_view. boost::string_view is implicitly convertible to std::string_view but not to fmt::basic_string_view, which is what string_view_t can be sometimes. 

example:
image

error:
image

This code fixes that. And std::format is not needed and can therefore be run on C++17 compilers such as GCC 7.

bachittle:
Yea sure. In my rare use case, I need to convert std::string_view to a string_view_t. I'm also doing things with boost::string_view and want to keep things as string views rather than as const char *, and am getting issues with doubly implicit conversions. It has an implicit conversion to std::string_view but obviously not to the fmt::basic_string_view<char>. std::string_view cannot convert to spdlog::string_view_t as it would be a double implicit conversion (spdlog::string_view_t <= std::string_view <= boost::string_view)

So this was my solution. There may be other instances where the user wants to use std::string_view but not std::format (fmt only provides a subset of std::string_views features). So it would be nice to have it decoupled for all users, which is why I made this PR.

example/example.cpp
```diff
@@ -279,7 +279,7 @@ struct fmt::formatter<my_type> : fmt::formatter<std::string>
 {
     auto format(my_type my, format_context &ctx) -> decltype(ctx.out())
     {
-        return format_to(ctx.out(), "[my_type i={}]", my.i);
+        return fmt::format_to(ctx.out(), "[my_type i={}]", my.i);
```
bachittle, line 282: this is also an error that arises when compiling spdlog using the standard set to C++20. Very simple fix, just add fmt namespace to the front and the code can compile using C++20. If this PR doesn't get merged this should at least be fixed.

gabime:
Thanks. The thing is, that spdlog already have too much ifdefs to my liking, and increasing the possible combinations of flags would complicate things even more. 

> this is also an error that arises when compiling spdlog using the standard set to C++20. Very simple fix, just add fmt namespace to the front and the code can compile using C++20. If this PR doesn't get merged this should at least be fixed.

Sure. PR is welcome

bachittle:
ok, will fix the other issue with a separate pr for the time being. If there is some alternate way I can go about replacing fmt::basic_string_view with std::string_view I will attempt that route instead of deviating too far with my fork. I think there may be flags in fmt to do this as well, and since there is a flag for external fmt I might do something like this instead.

gabime:
> ok, will fix the other issue with a separate pr for the time being

Thanks !

PR #2688, opened 2023-03-24
fix small issue when compiling with C++20 without std::format

bachittle:
make sure to specify that fmt namespace is in use when calling format_to. It's resolution was ambiguous when compiling with C++20.


gabime:
Thanks @bachittle

PR #2690, opened 2023-03-26
Added source location feature

mguludag:
Hello everyone,
I added source location feature and it does not require C++20 (only requirements are compiler versions). Also for adding this feature I wrapped format_string_t and added source_location::current() into its ctor.
image


gian21391:
That's a very nice idea! My suggestions are: 

- use std::source_location if available
- line and column should be unsigned to match std::source_location and to avoid conversion warnings

I've created a [patch](https://github.com/gabime/spdlog/files/11100364/patch.txt) with the modifications.

mguludag:
> That's a very nice idea! My suggestions are:
> 
> * use std::source_location if available
> * line and column should be unsigned to match std::source_location and to avoid conversion warnings
> 
> I've created a [patch](https://github.com/gabime/spdlog/files/11100364/patch.txt) with the modifications.

Thank you! I added your suggestions into pr 👍

mguludag:
Hello @gabime ,
Could you please review the pr at your available time?

gabime:
Thanks @mguludag . I am not sure yet wether to merge since it breaks the API (the new extra loc param at the start of the log function).

mguludag:
> Thanks @mguludag . I am not sure yet wether to merge since it breaks the API (the new extra loc param at the start of the log function).

You're welcome. Yes I called line 88 `void log(source_loc loc, level::level_enum lvl, format_string_t<Args...> fmt, Args &&... args)` instead of line 94 `void log(level::level_enum lvl, format_string_t<Args...> fmt, Args &&... args)`. I don't know maybe we will find a better solution. Tried hard to keep compatibility with every possible config (fmt/external_fmt/std::format...) 🤔

mguludag:
> Thanks @mguludag . I am not sure yet wether to merge since it breaks the API (the new extra loc param at the start of the log function).

Now I removed this parameter and used fmt.loc() instead of source_loc{} into this function: 
```C++
template<typename... Args>
void log(level::level_enum lvl, format_string_t<Args...> fmt, Args &&... args)
{
    log(fmt.loc(), lvl, fmt, std::forward<Args>(args)...);  // old: log(source_loc{}, lvl, fmt, std::forward<Args>(args)...);
}
```

mguludag:
Hi @gabime Can it breaks the API again?

HaroldCC:
I test this codes use MSVC, this is the output:
image
It is not correct file name and line number.
My Visual Studio version :17.6.6
C++ standard: /std:c++20

bebuch:
@mguludag Great patch!

As already stated by @HaroldCC there is unfortunately an asymmetry for single argument logs and format argument logs now.

```cpp
spdlog::info("single argument");
spdlog::info("format argument {}", 42);
```

When I run this from the `example.cpp` I get:

```
[2023-10-06 19:12:28.380] [info] single argument
[2023-10-06 19:12:28.380] [info] [example.cpp:48] format argument 42
```

This can simple be fixed by an additional `source_location` argument for single argument `info` and friends.

```diff
diff --git a/include/spdlog/logger.h b/include/spdlog/logger.h
index 9fe80dcb..10073486 100644
--- a/include/spdlog/logger.h
+++ b/include/spdlog/logger.h
@@ -275,9 +275,9 @@ public:
     }

     template<typename T>
-    void info(const T &msg)
+    void info(const T &msg, source_loc source = source_loc::current())
     {
-        log(level::info, msg);
+        log(source, level::info, msg);
     }

     template<typename T>
diff --git a/include/spdlog/spdlog.h b/include/spdlog/spdlog.h
index 695775b1..6ef1defb 100644
--- a/include/spdlog/spdlog.h
+++ b/include/spdlog/spdlog.h
@@ -264,9 +264,9 @@ inline void debug(const T &msg)
 }

 template<typename T>
-inline void info(const T &msg)
+inline void info(const T &msg, source_loc source = source_loc::current())
 {
-    default_logger_raw()->info(msg);
+    default_logger_raw()->info(msg, source);
 }

 template<typename T>
```

Of course that would be an API break again. @gabime:

> Thanks @mguludag . I am not sure yet wether to merge since it breaks the API (the new extra loc param at the start of the log function).

My opinion is that the new functionality is more valuable than full API backward compatibility. Only references and pointers to the functions are affected by the break:

```cpp
static_cast<void(*)(const int&)>(&spdlog::info); // fails with current PR
```

There is probably such a usage in real code somewhere, but it would be easy to fix by adding the new parameter to the cast or using a lambda wrapper. The latter fix would even compile with new and old versions of the lib.

```cpp
static_cast<void(*)(const int&)>([](const int& v){spdlog::info(v);}); // works always
```

In the documentation of spdlog it is also not ranted anywhere that pointers or references to these functions work. However, the lambda fix should be mentioned in the release notes to provide users with a simple solution.

We gain tremendously when source location suddenly just works all the time. Simple should be the default.

For the format argument overloads I find the solution to make the source location a part of the format string very good! Thus, there are never conflicts with the format arguments. In addition, an explicitly created source location can be submitted:

```cpp
auto const source = spdlog::source_loc::current();
spdlog::info("single argument", source);
spdlog::info({"format argument {}", source}, 42);
spdlog::info({"enforced format argument?", source}); // not sure what gets called
```

```
[2023-10-06 19:55:17.484] [info] [example.cpp:51] single argument
[2023-10-06 19:55:17.484] [info] [example.cpp:51] format argument 42
[2023-10-06 19:55:17.484] [info] [example.cpp:51] enforced format argument?
```

With the last line I am not sure which overload is called at which position. But in any case the result is convincing. Both the single argument syntax and the format argument syntax work for `source` with MSVC 2019. (Of course, it still needs to be checked whether this goes through for decent compilers. Unfortunately I don't have clang or GCC for testing at the moment.)

@mguludag What do you think? My coworkers and I would be very happy if the PR would be accepted.

gabime:
v2.x branch already supports source location. you need to define `#define SPDLOG_SOURCE_LOCATION ` before  including spdlog.h.

Please try it out and feedback. It requires to compile with cpp 20(-DCMAKE_CXX_STANDARD=20) although it seems to even work with some compilers with cpp 17 (which is the default standard in spdlog v2.x).

mguludag:
> @mguludag Great patch!
> 
> As already stated by @HaroldCC there is unfortunately an asymmetry for single argument logs and format argument logs now.
> 
> ```c++
> spdlog::info("single argument");
> spdlog::info("format argument {}", 42);
> ```
> 
> When I run this from the `example.cpp` I get:
> 
> ```
> [2023-10-06 19:12:28.380] [info] single argument
> [2023-10-06 19:12:28.380] [info] [example.cpp:48] format argument 42
> ```
> 
> This can simple be fixed by an additional `source_location` argument for single argument `info` and friends.
> 
> ```diff
> diff --git a/include/spdlog/logger.h b/include/spdlog/logger.h
> index 9fe80dcb..10073486 100644
> --- a/include/spdlog/logger.h
> +++ b/include/spdlog/logger.h
> @@ -275,9 +275,9 @@ public:
>      }
> 
>      template<typename T>
> -    void info(const T &msg)
> +    void info(const T &msg, source_loc source = source_loc::current())
>      {
> -        log(level::info, msg);
> +        log(source, level::info, msg);
>      }
> 
>      template<typename T>
> diff --git a/include/spdlog/spdlog.h b/include/spdlog/spdlog.h
> index 695775b1..6ef1defb 100644
> --- a/include/spdlog/spdlog.h
> +++ b/include/spdlog/spdlog.h
> @@ -264,9 +264,9 @@ inline void debug(const T &msg)
>  }
> 
>  template<typename T>
> -inline void info(const T &msg)
> +inline void info(const T &msg, source_loc source = source_loc::current())
>  {
> -    default_logger_raw()->info(msg);
> +    default_logger_raw()->info(msg, source);
>  }
> 
>  template<typename T>
> ```
> 
> Of course that would be an API break again. @gabime:
> 
> > Thanks @mguludag . I am not sure yet wether to merge since it breaks the API (the new extra loc param at the start of the log function).
> 
> My opinion is that the new functionality is more valuable than full API backward compatibility. Only references and pointers to the functions are affected by the break:
> 
> ```c++
> static_cast<void(*)(const int&)>(&spdlog::info); // fails with current PR
> ```
> 
> There is probably such a usage in real code somewhere, but it would be easy to fix by adding the new parameter to the cast or using a lambda wrapper. The latter fix would even compile with new and old versions of the lib.
> 
> ```c++
> static_cast<void(*)(const int&)>([](const int& v){spdlog::info(v);}); // works always
> ```
> 
> In the documentation of spdlog it is also not ranted anywhere that pointers or references to these functions work. However, the lambda fix should be mentioned in the release notes to provide users with a simple solution.
> 
> We gain tremendously when source location suddenly just works all the time. Simple should be the default.
> 
> For the format argument overloads I find the solution to make the source location a part of the format string very good! Thus, there are never conflicts with the format arguments. In addition, an explicitly created source location can be submitted:
> 
> ```c++
> auto const source = spdlog::source_loc::current();
> spdlog::info("single argument", source);
> spdlog::info({"format argument {}", source}, 42);
> spdlog::info({"enforced format argument?", source}); // not sure what gets called
> ```
> 
> ```
> [2023-10-06 19:55:17.484] [info] [example.cpp:51] single argument
> [2023-10-06 19:55:17.484] [info] [example.cpp:51] format argument 42
> [2023-10-06 19:55:17.484] [info] [example.cpp:51] enforced format argument?
> ```
> 
> With the last line I am not sure which overload is called at which position. But in any case the result is convincing. Both the single argument syntax and the format argument syntax work for `source` with MSVC 2019. (Of course, it still needs to be checked whether this goes through for decent compilers. Unfortunately I don't have clang or GCC for testing at the moment.)
> 
> @mguludag What do you think? My coworkers and I would be very happy if the PR would be accepted.

I removed the `source_location` parameter from single arg log functions for compatibility. In my previous commits you can see the argument in those functions.  As @gabime mentioned the `v2.x` supports the C++20 `source_location`, maybe I'll create PR to `v2.x` branch that adds `source_location` feature for older C++ standards 🤔

gabime:
>ranch that adds source_location feature for older C++ standards 🤔

spdlog v2.x requires c++17 or above, so there is no much benefit to add it to v2.x (especially since it works even with c++17 when `exeprimental/source_location.h` is available).

PR #2694, opened 2023-03-31
fix build for master fmt (non-bundled)

### LINKED ISSUE #2693: spdlog fails to build with latest fmt
Using latest fmt, the spdlog will fail to build.

You must change the **basic_runtime** to **runtime_format_string** to fix the issue, however bundled spdlog uses **basic_runtime.** Line: https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/common.h#L181

H1X4Dev:
fixes #2693

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -173,10 +173,11 @@ using format_string_t = fmt::format_string<Args...>;
 template<class T>
 using remove_cvref_t = typename std::remove_cv<typename std::remove_reference<T>::type>::type;
 
+template <typename Char>
 #if FMT_VERSION >= 90101
-# define FMT_STRING_RUNTIME fmt::runtime_format_string<Char>
+using fmt_runtime_string = fmt::runtime_format_string<Char>;
 #else
-# define FMT_STRING_RUNTIME fmt::basic_runtime<Char>
+using fmt_runtime_string = fmt::basic_runtime<Char>;
 #endif
 
 // clang doesn't like SFINAE disabled constructor in std::is_convertible<> so have to repeat the condition from basic_format_string here,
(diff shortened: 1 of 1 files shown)


include/spdlog/common.h
```diff
@@ -173,12 +173,18 @@ using format_string_t = fmt::format_string<Args...>;
 template<class T>
 using remove_cvref_t = typename std::remove_cv<typename std::remove_reference<T>::type>::type;
 
+#if FMT_VERSION >= 90101
+# define FMT_STRING_RUNTIME fmt::runtime_format_string<Char>
+#else
+# define FMT_STRING_RUNTIME fmt::basic_runtime<Char>
+#endif
```
mguludag: I think its better than macro
```C++
template <typename Char>
#if FMT_VERSION >= 90101
using fmt_runtime_string = fmt::runtime_format_string<Char>;
#else
using fmt_runtime_string = fmt::basic_runtime<Char>;
#endif
```

include/spdlog/common.h
```diff
@@ -173,12 +173,18 @@ using format_string_t = fmt::format_string<Args...>;
...
+
 // clang doesn't like SFINAE disabled constructor in std::is_convertible<> so have to repeat the condition from basic_format_string here,
 // in addition, fmt::basic_runtime<Char> is only convertible to basic_format_string<Char> but not basic_string_view<Char>
 template<class T, class Char = char>
 struct is_convertible_to_basic_format_string
     : std::integral_constant<bool,
-          std::is_convertible<T, fmt::basic_string_view<Char>>::value || std::is_same<remove_cvref_t<T>, fmt::basic_runtime<Char>>::value>
+          std::is_convertible<T, fmt::basic_string_view<Char>>::value || std::is_same<remove_cvref_t<T>, FMT_STRING_RUNTIME>::value>
```
mguludag: After using declarations it have to be changed like this
```C++
std::is_convertible<T, fmt::basic_string_view<Char>>::value || std::is_same<remove_cvref_t<T>, fmt_runtime_string<Char>>::value>
```

include/spdlog/common.h
gabime: I agree. Seems cleaner using `using`. @H1X4Dev Please update the pr

H1X4Dev:
Updated, thank you for pointing out the better way.

gabime:
Merged. Thanks @H1X4Dev and @mguludag

PR #2707, opened 2023-04-15
Override BUILD_SHARED_LIBS with SPDLOG_BUILD_SHARED

rafalmaziejuk:
SPDLOG_BUILD_SHARED option should override BUILD_SHARED_LIBS value
in case user wants to force building spdlog as static library.

Signed-off-by: rafalmaziejuk <rafalmaziejuk@gmail.com>


gabime:
Not sure I understand why it should override it? perhaps provide an example why its needed.

rafalmaziejuk:
> Not sure I understand why it should override it? perhaps provide an example why its needed.

In my case, I would like to compile spdlog into my library so I don't need to include spdlog in distributed package. Of course the same can be achieved with linking against already compiled static spdlog but this approach is more convenient when adding spdlog through `add_subdirectory`.

```
add_subdirectory(third_party/spdlog spdlog)
set(SPDLOG_BUILD_SHARED OFF CACHE INTERNAL "") #build static spdlog

target_link_libraries(
    lib_name
    PRIVATE
    spdlog
)
```

gabime:
This might break existing builds, so i prefer not to unless there is no other way.

PR #2717, opened 2023-04-23
Update fmtlib fixed issues

L-Super:
Because fmtlib fixes this problem, it hasn't been released yet. The spdlog lib binds it. In order to better use sdplog, I want to fix the error.

To fix [fmtlib/fmt#3137](https://github.com/fmtlib/fmt/pull/3137)

I know it is not a wise choice, but I can't think of a better way for people to avoid this mistake


gabime:
I’d rather wait for an official fmt release. @vitaut Is there an estimate for the next fmt release ?

vitaut:
The new release should be in the next couple of weeks.

gabime:
Thanks @vitaut . Closing.

Issue #2727, opened 2023-05-03
Adding minute_file_sink_mt


mmanoj:
I created basic sink implementation as attached and driver code. However getting below errors. Please advice to fix it and if all fine we can merge with project.

===================
Errors
`logtest.cpp: In function ‘void setupLogging()’:
logtest.cpp:17:101: error: no matching function for call to ‘spdlog::sinks::minute_file_sink<std::mutex>::minute_file_sink(const char [13], const char [17], bool, int)’
   17 |         auto logger = spdlog::sinks::minute_file_sink_mt("basic_logger", "logs/min-log.txt",false,72);
      |                                                                                                     ^
In file included from logtest.cpp:10:
../include/spdlog/sinks/minute_file_sink.h:49:5: note: candidate: ‘spdlog::sinks::minute_file_sink<Mutex, FileNameCalc>::minute_file_sink(spdlog::filename_t, bool, uint16_t, const spdlog::file_event_handlers&) [with Mutex = std::mutex; FileNameCalc = spdlog::sinks::minute_filename_calculator; spdlog::filename_t = std::__cxx11::basic_string<char>; uint16_t = short unsigned int]’
   49 |     minute_file_sink(
      |     ^~~~~~~~~~~~~~~~
../include/spdlog/sinks/minute_file_sink.h:50:109: note:   no known conversion for argument 4 from ‘int’ to ‘const spdlog::file_event_handlers&’
   50 |         filename_t base_filename, bool truncate = false, uint16_t max_files = 0, const file_event_handlers &event_handlers = {})
      |                                                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~

`

mmanoj:
[logtest.cpp.txt](https://github.com/gabime/spdlog/files/11383632/logtest.cpp.txt)
[minute_file_sink.h.txt](https://github.com/gabime/spdlog/files/11383641/minute_file_sink.h.txt)

Driver code and sink files attached for reference.Appreciate your help

tt4g:
In `setupLogging()`, passing `74` 4th arguments.

```
auto logger = spdlog::sinks::minute_file_sink_mt("basic_logger", "logs/min-log.txt",false,72);
                                                                                                                                             ^^
```

But declared `file_event_handlers` in constructor.

```
minute_file_sink(
        filename_t base_filename, bool truncate = false, uint16_t max_files = 0, const file_event_handlers &event_handlers = {})
[...]

PR #2729, opened 2023-05-04
Adding minute_file_sink_mt #2727

mmanoj:
Adding minute_file_sink_mt #2727 to support minute_file_sink to support every 1 min file rotation.
[minute_file_sink.h.txt](https://github.com/gabime/spdlog/files/11395573/minute_file_sink.h.txt)

https://github.com/gabime/spdlog/commit/5306b1fd58b84a8b8471c1e0efb6750614a66002


include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,208 @@
...
+ * Generator of Minute log file names in format basename.YYYY-MM-DD-HH-MM.ext
+ */
+struct minute_filename_calculator
+{
+    // Create filename for the form basename.YYYY-MM-DD-H-M
+    static filename_t calc_filename(const filename_t &filename, const tm &now_tm)
+    {
+        filename_t basename, ext;
+        std::tie(basename, ext) = details::file_helper::split_by_extension(filename);
+        return fmt_lib::format(SPDLOG_FILENAME_T("{}_{:04d}-{:02d}-{:02d}-{:02d}_{:02d}{}"), basename, now_tm.tm_year + 1900, now_tm.tm_mon + 1,
```
tt4g:
```suggestion
        return fmt_lib::format(SPDLOG_FMT_STRING(SPDLOG_FILENAME_T("{}_{:04d}-{:02d}-{:02d}-{:02d}_{:02d}{}")), basename, now_tm.tm_year + 1900, now_tm.tm_mon + 1,
```

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,208 @@
...
+            now_tm.tm_mday, now_tm.tm_hour,now_tm.tm_min,ext);//colock details define in os-inl.h
```
tt4g:
```suggestion
            now_tm.tm_mday, now_tm.tm_hour,now_tm.tm_min,ext);
```

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,208 @@
...
+/*
+ * Rotating file sink based on time(minutes).
+ * If truncate != false , the created file will be truncated.
+ * If max_files > 0, retain only the last max_files and delete previous.
+ */
+template<typename Mutex, typename FileNameCalc = minute_filename_calculator>
+class minute_file_sink final : public base_sink<Mutex>
+{
+public:
+    // create hourly file sink which rotates on given time
```
tt4g:
```suggestion
    // create every minute file sink which rotates on given time
```

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+protected:
+    void sink_it_(const details::log_msg &msg) override
+    {
+        auto time = msg.time;
+        bool should_rotate = time >= rotation_tp_;
+        if (should_rotate)
+        {
+            if (remove_init_file_)
```
tt4g: I don't understand why you would delete an empty log file.
Why is this necessary?

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,208 @@
...
+    }
+};
+
+/*
+ * Rotating file sink based on time(minutes).
+ * If truncate != false , the created file will be truncated.
+ * If max_files > 0, retain only the last max_files and delete previous.
+ */
+template<typename Mutex, typename FileNameCalc = minute_filename_calculator>
+class minute_file_sink final : public base_sink<Mutex>
```
tt4g: daily_sink and hourly_sink can change the rotation time, but `minute_file_sink` seems to have a fixed rotation time. I suggest renaming it to `every_minute_file_sink` to avoid confusion.

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,208 @@
...
+    {
+        time_t tnow = log_clock::to_time_t(tp);
+        return spdlog::details::os::localtime(tnow);
+    }
+
+    log_clock::time_point next_rotation_tp_()
+    {
+        auto now = log_clock::now();
+        tm date = now_tm(now);
+        date.tm_min = 0;
```
tt4g:
```suggestion

```

The most important rotation feature seems to be broken.
The minute of the next rotation time is always 1 (1 minute every hour).

tt4g (commented):
I commented on the points of concern.
It must also pass @gabime review after modification.

include/spdlog/sinks/minute_file_sink.h
mmanoj: @tt4g 
Thanks for suggestions and review,I will add configuration to make X MIN interval to do the rotation as configurable number of minutes.Hope it more flexible.I also go through other changes as well

include/spdlog/sinks/minute_file_sink.h
mmanoj: Updated, under testing

include/spdlog/sinks/minute_file_sink.h
mmanoj: This part kept as it is from hourly log sink.

`bool should_rotate = time >= rotation_tp_;
        if (should_rotate)
        {
            if (remove_init_file_)
            {
                file_helper_.close();
                details::os::remove(file_helper_.filename());`

tt4g: Somehow I figured out why this code is there.
Perhaps, the author of sink wanted to clean up the log file created in the constructor, if it was empty, removing it at the time of rotation.

The following scenario is assumed.

1. Create log file `log.2023-01-01-12-01.txt` in the constructor.
2. More than one minute passes without a log being recorded.
3. The contents of the file to be rotated are empty. Delete it as unnecessary and record log messages in a new log file.

Can you add a comment to this code as I am having difficulty understanding the purpose of this code?

include/spdlog/sinks/minute_file_sink.h
mmanoj: Updated with configurable rotation interval minutes

auto logger = spdlog::minute_logger_mt("basic_logger", "logs/min-log.txt",false,60,2);

mmanoj (commented):
fixed suggested changes and configurable minutes

tt4g:
I think the current rotation implementation is wrong.
`minute_file_sink` rotates at intervals (minutes).
However, the daily sink and hourly sink rotate at a specified time (hour or minute).
If you are adding sinks that work differently, I think you should change the name of the sinks to make it easier to understand.

mmanoj:
@tt4g 

Only sink name change required? hope the other implemented logic and suggestions are fixed?

tt4g:
> Only sink name change required? hope the other implemented logic and suggestions are fixed?

Did you not develop this sink because you wanted it?
If you wanted this sink that rotates on an interval, change the name.
If not, please fix it.

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,217 @@
...
+        , file_helper_{event_handlers}
+        , truncate_(truncate)
+        , max_files_(max_files)
+        ,rotation_m_(rotation_minute)
+        , filenames_q_()
+    {
+        
+        if (rotation_minute < 0 || rotation_minute > 59)
+        {
+            throw_spdlog_ex("daily_file_sink: Invalid rotation time in ctor");
```
gabime: 1. Why can't it be greater than 59? what would happen if someone wants 90 minutes for example?
2. should be "minute_file_sink: Invalid rotation_minutes value in ctor"

mmanoj: @gabime 

Thanks for the feedback and got your point, remove the validation considering the concern.

mmanoj:
@tt4g 

Yes, This is required as per my project requirements and hope many people like the feature.I change the name to "minute_interval_logger_mt" to make it meaningful.

include/spdlog/sinks/minute_file_sink.h
tt4g: @mmanoj Note: In the current rotation implementation, `rotation_m_` should not be allowed to be `0`.

mmanoj: @tt4g 
Implement the check for 0 min.

gabime:
@mmanoj  Please add a unit test for this sink. Something very basic just to verify it compiles right (and perhaps test to validate it throws on invalid input).

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,217 @@
...
+template<typename Factory = spdlog::synchronous_factory>
+inline std::shared_ptr<logger> minute_interval_logger_mt(const std::string &logger_name, const filename_t &filename, bool truncate = false,
+    uint16_t max_files = 0, int minute = 0, const file_event_handlers &event_handlers = {})
+{
+    return Factory::template create<sinks::minute_interval_file_sink_mt>(logger_name, filename, truncate, max_files, minute ,event_handlers);
+}
+
+template<typename Factory = spdlog::synchronous_factory>
+inline std::shared_ptr<logger> minute_interval_logger_st(const std::string &logger_name, const filename_t &filename, bool truncate = false,
+    uint16_t max_files = 0,int minute = 0, const file_event_handlers &event_handlers = {})
```
tt4g:
```suggestion
    uint16_t max_files = 0, int minute = 1, const file_event_handlers &event_handlers = {})
```

`minute = 0` will thrown.

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,217 @@
...
+using minute_interval_file_sink_st = minute_interval_file_sink<details::null_mutex>;
+
+} // namespace sinks
+
+//
+// factory functions
+//
+template<typename Factory = spdlog::synchronous_factory>
+inline std::shared_ptr<logger> minute_interval_logger_mt(const std::string &logger_name, const filename_t &filename, bool truncate = false,
+    uint16_t max_files = 0, int minute = 0, const file_event_handlers &event_handlers = {})
```
tt4g:
```suggestion
    uint16_t max_files = 0, int minute = 1, const file_event_handlers &event_handlers = {})
```

`minute = 0` will thrown.

tt4g (changes requested):

mmanoj:
> @mmanoj Please add a unit test for this sink. Something very basic just to verify it compiles right (and perhaps test to validate it throws on invalid input).

@gabime Sure,can you refer some example, I'm checking under test/test_daily_logger

gabime:
1. Just look at any test in the tests for example. please create a new file in tests with onr or two tests.
2. Also, I think a better name for the sink would be “periodic_file_sink”. This way in the future it could be extended to support any chrono units, not just minutes.
3. Please remove the lime with the email address from the top comments. I try to have same header format for all files

mmanoj:
> 
1. Working On the test cases files
2. Rename the file and update the sink names as per the advice.Also we can make it  seconds as base unit, so we can represent minutes and hours from that. What do you think?
3. Done.

mmanoj:
https://github.com/mmanoj/spdlog/blob/v1.x/include/spdlog/sinks/periodic_file_sink.h

gabime:
1. I meant using `std::chrono::duration<Rep, Period> interval` as for the interval. see https://github.com/gabime/spdlog/blob/57a9fd0841f00e92b478a07fef62636d7be612a8/include/spdlog/spdlog.h#LL85C12-L85C12 for an example.

However, for this to work the generated filename should probably also contain seconds in the name and the units restricted to seconds or higher,(e.g. not accept milliseconds or micros  using  `std::enable_if`)
so I suggest leaving this for a later stage and start with minutes for this initial version

2. The name should be periodic_file_sink instead of periodic_interval_file_sink.

mmanoj:
> 1. I meant using `std::chrono::duration<Rep, Period> interval` as for the interval. see https://github.com/gabime/spdlog/blob/57a9fd0841f00e92b478a07fef62636d7be612a8/include/spdlog/spdlog.h#LL85C12-L85C12 for an example.
> 
> However, for this to work the generated filename should probably also contain seconds in the name and the units restricted to seconds or higher,(e.g. not accept milliseconds or micros using `std::enable_if`) so I suggest leaving this for a later stage and start with minutes for this initial version
> 
> 2. The name should be periodic_file_sink instead of periodic_interval_file_sink.

Updated as per advice.Please check and update.Once this clear I will work on the unit test

gabime:
I see the pr contains 2 sink files? which one is it?

mmanoj:
> I see the pr contains 2 sink files? which one is it?

https://github.com/mmanoj/spdlog/blob/v1.x/include/spdlog/sinks/periodic_file_sink.h

mmanoj:
> > I see the pr contains 2 sink files? which one is it?
> 
> https://github.com/mmanoj/spdlog/blob/v1.x/include/spdlog/sinks/periodic_file_sink.h

@gabime 

Any feedback regarding the PR?

include/spdlog/sinks/minute_file_sink.h
```diff
@@ -0,0 +1,217 @@
+// Copyright(c) 2015-present, Gabi Melman & spdlog contributors.
```
gabime: Please remove this file as we use the other one

include/spdlog/sinks/periodic_file_sink.h
```diff
@@ -0,0 +1,217 @@
...
+ * If max_files > 0, retain only the last max_files and delete previous.
+ */
+template<typename Mutex, typename FileNameCalc = periodic_filename_calculator>
+class periodic_file_sink final : public base_sink<Mutex>
+{
+public:
+   // create every periodic file sink which rotates on given time
+    periodic_file_sink(
+        filename_t base_filename, bool truncate = false, uint16_t max_files = 0,
+        std::chrono::duration<int> rotation_period = 0, const file_event_handlers &event_handlers = {})
```
gabime, line 55: Should be 

```c++
  periodic_file_sink(
       filename_t base_filename, 
       std::chrono::minutes rotation_interval,
       uint16_t max_files max_files,
       bool truncate_first = false,       
       const file_event_handlers &event_handlers = {})

```

include/spdlog/sinks/periodic_file_sink.h
```diff
@@ -0,0 +1,217 @@
...
+        , max_files_(max_files)
+        ,rotation_p_(rotation_period)
+        , filenames_q_()
+    {
+        
+          if (rotation_period <= std::chrono::minutes(0))
+        {
+            throw_spdlog_ex("periodic_file_sink: Invalid rotation time in ctor");
+        }
+
```
gabime, line 76: I think we should add some sanity restriction on max_files, especially since we store all the names in a buffer

```c++
  if (max_files > 2000)  
   {
         throw_spdlog_ex("periodic_file_sink: max_files param exceeds max of 2000");
   }
```

gabime (changes requested):

include/spdlog/sinks/minute_file_sink.h
mmanoj: done

include/spdlog/sinks/periodic_file_sink.h
mmanoj: Done

include/spdlog/sinks/periodic_file_sink.h
mmanoj: Done

gabime:
Thanks. Would you add a test?

mmanoj:
> Thanks. Would you add a test?

If current implementation is fine,I will work on it.Please confirm.

gabime:
The thing is, I cannot know if it even complies without a minimal test. Just add minimal test that includes it and use it in the most basic way.

mmanoj:
> The thing is, I cannot know if it even complies without a minimal test. Just add minimal test that includes it and use it in the most basic way.

sure, will work on this and update you.

mmanoj:
> > The thing is, I cannot know if it even complies without a minimal test. Just add minimal test that includes it and use it in the most basic way.
> 
> sure, will work on this and update you.

Can you help to understand how spdlog test framework work.I'm not able to run the test cases except example application under spdlog/tests/example directory.

for example how to run test_daily_logger.cpp

include/spdlog/sinks/periodic_file_sink.h
```diff
@@ -0,0 +1,231 @@
...
+} // namespace sinks
+
+//
+// factory functions
+//
+template<typename Factory = spdlog::synchronous_factory>
+inline std::shared_ptr<logger> periodic_logger_mt(const std::string &logger_name, const filename_t &filename,std::chrono::minutes period = 1,uint16_t max_files = 0, bool truncate = false,
+     const file_event_handlers &event_handlers = {})
+{
+    return Factory::template create<sinks::periodic_file_sink_mt>(logger_name, filename,period,max_files, truncate, event_handlers);
```
kegechen, line 222:
```suggestion
    return Factory::template create<sinks::periodic_file_sink_mt>(logger_name, filename, period,max_files, truncate, event_handlers);
```

include/spdlog/sinks/periodic_file_sink.h
```diff
@@ -0,0 +1,231 @@
...
+}
+
+template<typename Factory = spdlog::synchronous_factory>
+inline std::shared_ptr<logger> periodic_logger_st(const std::string &logger_name, const filename_t &filename,std::chrono::minutes period = 1,uint16_t max_files = 0, bool truncate = false,
+     const file_event_handlers &event_handlers = {})
+{
+    return Factory::template create<sinks::periodic_file_sink_st>(logger_name, filename,period, max_files,truncate,   event_handlers);
```
kegechen, line 229:
```suggestion
    return Factory::template create<sinks::periodic_file_sink_st>(logger_name, filename, period, max_files,truncate, event_handlers);
```

include/spdlog/sinks/periodic_file_sink.h
```diff
@@ -0,0 +1,231 @@
...
+    periodic_file_sink(
+       filename_t base_filename, 
+       std::chrono::minutes rotation_interval,
+       uint16_t max_files,
+       bool truncate_first = false,       
+       const file_event_handlers &event_handlers = {})
+        : base_filename_(std::move(base_filename))
+        , file_helper_{event_handlers}
+        , truncate_(truncate_first)
+        , max_files_(max_files)
```
kegechen, line 67: It would be great if the max file size was also configurable
like 20 * 1024 *1024 （20MB）
like [rotating_file_sink](https://github.com/gabime/spdlog/blob/2312489bdc304886b4a2eebc8f7f8dc0c4de688f/include/spdlog/sinks/rotating_file_sink.h#L47C14-L47C14)

mmanoj: @kegechen 

Thanks for the suggestion,I will look in to that.I have some issues on running the test case as I mention above.If someone can help to fix it I can quickly close the PR.

kegechen: Can't help, unfortunately I'm not Not familiar with the test case either
[Catch2 tutorial](https://github.com/catchorg/Catch2/blob/devel/docs/tutorial.md#top) may help

kegechen: by the way, to many commits in your pr. should have rebase first
the title of pr also should change

mmanoj: @kegechen 

Sorry for the delay due to some urgent tasks,I'm going to start working on this again.Will update the progress

PR #2736, opened 2023-05-17
Removes special format handling for fmt. 

### LINKED ISSUE #2735: spdlog 1.11.0 with fmt 1.10.0 has test_daily_logger:126 failing
I am using stable spdlog 1.11.0 with stable fmt 10.0.0. To make it work I've added patch 0ca574ae168820da0268b3ec7607ca7b33024d05. The compilation now works, but the test unit [test_daily_logger](https://github.com/gabime/spdlog/blob/v1.x/tests/test_daily_logger.cpp#L126) now fails on me.

```
[   23s] daily_file_sink::daily_filename_format_calculator
[   23s] -------------------------------------------------------------------------------
[   23s] /home/abuild/rpmbuild/BUILD/spdlog-1.11.0/tests/test_daily_logger.cpp:126
[   23s] ...............................................................................
[   23s] 
[   23s] /home/abuild/rpmbuild/BUILD/spdlog-1.11.0/tests/test_daily_logger.cpp:126: FAILED:
[   23s] due to unexpected exception with message:
[   23s]   invalid format
```

Has anyone any insight, what might have changed here due to the patch? If I build spdlog with the bundled fmt it works.

commel:
Regains test compatibility with fmt 1.10.0.

Most of the patch was provided by @tt4g.

fixes #2735

### DIFF
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -18,7 +18,6 @@
 #include <ctime>
 #include <mutex>
 #include <string>
-#include <vector>
 
 namespace spdlog {
 namespace sinks {
@@ -58,12 +57,12 @@ struct daily_filename_format_calculator
         // https://github.com/fmtlib/fmt/issues/2238
         tm_format.push_back(' ');
 
-        const size_t MIN_SIZE = 10;
-        std::vector<char> buf;
+        const size_t MIN_SIZE = 64;
+        filename_t buf;
         buf.resize(MIN_SIZE);
         for (;;)
         {
-            size_t count = strftime(buf.data(), buf.size(), tm_format.c_str(), &now_tm);
+            size_t count = strftime(&buf[0], buf.size(), tm_format.c_str(), &now_tm);
             if (count != 0)
             {
                 // Remove the extra space.
@@ -73,7 +72,7 @@ struct daily_filename_format_calculator
             buf.resize(buf.size() * 2);
         }
 
-        return std::string(buf.cbegin(), buf.cend());
+	return buf;
     }
 
 private:
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -13,6 +13,9 @@
 #include <spdlog/details/circular_q.h>
 #include <spdlog/details/synchronous_factory.h>
 
+#include <iostream>
+#include <sstream>
+#include <iomanip>
 #include <chrono>
 #include <cstdio>
 #include <ctime>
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/daily_file_sink.h
```diff
@@ -59,7 +59,7 @@ struct daily_filename_format_calculator
         tm_format.push_back(' ');
 
         const size_t MIN_SIZE = 10;
-        filename_t buf;
+        std::vector<char> buf;
```
tt4g:
```suggestion
        filename_t buf;
```

This line should not be changed.
`filename_t` can be an alias for `std::wstring`.

https://github.com/gabime/spdlog/blob/57a9fd0841f00e92b478a07fef62636d7be612a8/include/spdlog/common.h#L121-L129

If you care that `std::string::data()` returns `const char *`, replace `buf.data()` with `&buf[0]`.
See also: [Can I get a non\-const C string back from a C\+\+ string? \- Stack Overflow](https://stackoverflow.com/questions/1919626/can-i-get-a-non-const-c-string-back-from-a-c-string)

tt4g (changes requested):

include/spdlog/sinks/daily_file_sink.h
commel: It does not look nice, but unter the premise of wstring-support and that fmt's implementation uses &buf[0] as well, i'm okay with that.

tt4g (approved):

include/spdlog/sinks/daily_file_sink.h
```diff
@@ -73,19 +72,7 @@ struct daily_filename_format_calculator
             buf.resize(buf.size() * 2);
```
gabime: How about adding a check that the resulting buf is not too big ?
Not sure if this can actually happen, but in order to deal with situations when `strftime` keeps returning zero due to some error in formatting (or any other reason):
```c++
if(buf.size() > 4096)) 
{
  throw spdlog_ex("daily_file_sink: calc_filename() - strftime error or file name too big")
}

Comment (commel) [reply include/spdlog/sinks/daily_file_sink.h]: A filename (the single file name not the file path itself) is usually limited to 255 bytes, if [wikipedia](https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits) is correct. If would resize it to that size and check if the result would not overflow. How about that?

Comment (gabime) [reply include/spdlog/sinks/daily_file_sink.h]: `calc_filename()` accepts a full path name, including the path afair. The param name `base_filename` is confusing

**Comment by gabime** [comment]
Since `filename_t` might be `wstring`in windows, I think we should use `wcsftime` in such case:

```cpp

#if defined(_WIN32) && defined(SPDLOG_WCHAR_FILENAMES)
    size_t count = std::wcsftime(&buf[0], buf.size(), tm_format.c_str(), &now_tm);
#else
    size_t count = std::strftime(&buf[0], buf.size(), tm_format.c_cstr(), &now_tm);
#endif
```

gabime:
I just found out the we could just simply use[ std::put_time](https://en.cppreference.com/w/cpp/io/manip/put_time) and avoid all this complexity :)

```c++
#include <iostream>
#include <iomanip>
 
static filename_t calc_filename(const filename_t &filepath, const tm &now_tm)
{
#if defined(_WIN32) && defined(SPDLOG_WCHAR_FILENAMES)
    std::wstringstream stream;  
else
  std::stringstream stream;
#endif
    stream << put_time(now_tm, file_path); 
    return stream.str();
}
```

The only downside is that it might not be supported on very old compilers like gcc < 5, which I can live with..

commel:
> I just found out the we could just simply use[ std::put_time](https://en.cppreference.com/w/cpp/io/manip/put_time) and avoid all this complexity :)

True! Is wasn't sure if there is something else needed. I've updated the PR on that issue. Btw, there are two calc_filename methods in ```include/spdlog/sinks/daily_file_sink.h```, is that right?

> The only downside is that it might not be supported on very old compilers like gcc < 5, which I can live with.

The README says you need a C++11 compiler :-)

gabime:
> Btw, there are two calc_filename methods in include/spdlog/sinks/daily_file_sink.h, is that right?

Yes. In `daily_filename_calculator`class (default, no custom formatting) and in `daily_filename_format_calculator` which do the strftime.

gabime:
We can also now remove this whole block in lines 64-82:


```cpp
 #if defined __GNUC__
#    pragma GCC diagnostic push
#    pragma GCC diagnostic ignored "-Wformat-nonliteral"
#endif

    static size_t strftime(char *str, size_t count, const char *format, const std::tm *time)
    {
        return std::strftime(str, count, format, time);
    }

    static size_t strftime(wchar_t *str, size_t count, const wchar_t *format, const std::tm *time)
    {
        return std::wcsftime(str, count, format, time);
    }

#if defined(__GNUC__)
#    pragma GCC diagnostic pop
#endif

```

gabime:
Merged. Thanks @commel and @tt4g !

kloczek:
Is it possible to release new version because fmt 10 updates? 🤔

PR #2744, opened 2023-05-27
Fix build with latest fmt

eli-b:
Building with the latest external fmt requires this change.
I admit I don't understand the change that I made here, but it fixes the build for me.


gabime:
Thanks. This fixes issue reported in fmt repo: https://github.com/fmtlib/fmt/issues/3443.

PR #2753, opened 2023-06-05
qt_sink: add some warning on its usage

Ulmo-F:


gabime:
Thanks. I think a single warming in the top is sufficient

Ulmo-F:
Some IDE display the comment above the class/method/..., I thought it was what you wanted when you asked for comment in the header file.
examples:
image

image

If you favor a cleaner source file, I'll remove the duplicates.

gabime:
I’d rather have single warning or even better, *NOTE* at the top of the file instead.  I will merge anyway and update

PR #2778, opened 2023-06-26
add option for not printing backtrace start and end message

### LINKED ISSUE #2774: Possibility to delete Backtrace Start and Backtrace End
Hi,
is there a possibility to delete Backtrace Start and End message? 

If not: Is there another possibility to buffer/store some debug messages and only dump them, when a condition is true?

markusmu0:
fix #2774


gabime:
This breaks the ABI. And anyway this s a very specific need, so closing for now.

markusmu0:
PR was rejected by gabime. Thus, the problem still exists. Is there another solution for this?

Issue #2784, opened 2023-06-28
IPV6 support on tcp_client.h


Simon-Janos:
Hi,

I have noticed in connect() of tcp_client.h (and of tcp_client-windows.h as well) that it is tied to IPv4, could it be changed AF_UNSPEC for ipv6 support?
hints.ai_family = AF_INET;       // IPv4

https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/details/tcp_client.h#L61
https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/details/tcp_client-windows.h#L86

Sincerely,
Janis

tt4g:
That structure is a hint, and `AF_INET` specifies IPv4, so if you pass a hostname to `getaddrinfo()`, IPv4 will be returned.
However, it is only a hint. If you pass an IPv6 string to the `host` argument, it should connect to IPv6.

See: https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfo

Simon-Janos:
So on my ipv6 rocky8/redhat environments I get "::getaddrinfo failed: Name or service not known" error. Which clearly does exist and sure enough no problem on ipv4. If I try with the service ip address (IPv6) I got "address family not supported" whiles it does work with a proper ipv4 address.

tt4g:
Are you saying that passing an IPv6 string will not work? If so, it may have to be `AF_INET6` or `AF_UNSPEC` to work.

Simon-Janos:
So on some linux/c version - at least - they are incompatible. AF_UNSPEC is a better approach or at least we are better making it configurable.

gabime:
@Simon-Janos PR is welcome.

gabime:
Implemented in #2790

PR #2790, opened 2023-06-30
Removing IPv4 limitation from tcp_client

Simon-Janos:
For the issue: https://github.com/gabime/spdlog/issues/2784.


gabime:
Thanks @Simon-Janos

PR #2803, opened 2023-07-06
feat: add hierarchical logger support

Pro:
This adds support for hierarchical loggers (see also #1061)

By default, propagate is set to true, this means all log messages will be propagated to higher hierarchical levels.
Logger hierarchies are separated by using `.` in the logger name.

The given example code will print these lines:

```
lvl1 [propagate_lvl1.lvl2.lvl3.lvl4] [debug] I am a debug message at Level 4 but will be printed by Level 1 logger
lvl1 [propagate_lvl1.lvl2] [debug] I am a debug message at Level 2 but will be printed by Level 1 logger
lvl1 [multi_lvl1.lvl2.lvl3.lvl4] [debug] I am a debug message at Level 4 but will be printed by Level 1 logger
lvl2 [multi_lvl1.lvl2.lvl3.lvl4] [info] I am an info message at Level 4 but will be printed by Level 2, Level 1 and root logger
lvl1 [multi_lvl1.lvl2.lvl3.lvl4] [info] I am an info message at Level 4 but will be printed by Level 2, Level 1 and root logger
root [multi_lvl1.lvl2.lvl3.lvl4] [info] I am an info message at Level 4 but will be printed by Level 2, Level 1 and root logger
```

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -26,6 +26,7 @@ void udp_example();
 void custom_flags_example();
 void file_events_example();
 void replace_default_logger_example();
+void hierarchical_logger_example();
 
 #include "spdlog/spdlog.h"
 #include "spdlog/cfg/env.h"  // support for loading levels from the environment variable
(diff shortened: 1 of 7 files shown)


gabime:
Thanks @Pro . Currently I am not interested in adding this feature, since I don't want to add extra complexity to spdlog. 
Note that for controlling a set of loggers in a sub tree(e.g. set the level of all loggers  “mylogger.network.*”) a simpler solution is possible by just implementing registry::find_sub_loggers(logger), which would return vector of sub loggers solely based on the names.

PR #2814, opened 2023-07-13
Multi color range

drakbar:
### Why

So while _implementing a custom formatter_, I really wanted the ability to set multiple color ranges. Patterns clearly state only one set of delimiters can be used, and trying to set multiple color ranges using the data members in the `log_msg` struct would just reset the color range to the last one specified. After looking behind the curtain I discovered  why that was the case. I would like to see the ability to define multiple color ranges across both area and color dimensions.

Anyways I was just curious if this type of feature once stable would be welcomed to merge into the main branch. I am working on this in my free time, so I am just writing this idea out to disk.

### How

This set of commits represents the naive approach as I am not fully acquainted with the code base. After getting half a proof of concept completed, I realized the actual challenge to this problem. The special case that needs to be considered for a stable feature is nesting of color ranges. Color range really isn't the correct word, because you can do more than just adjust the font color, but also style such as bold. These commits solves the multiple, discrete and regions. It does not however handle nesting of styled regions. I think it should be possible to solve for that with a FIFO data structure like a queue. Storing the position and type (opening or closing) should allow for branching when consuming the tokens. There would probably need to be some bookkeeping to trigger when to set the next un-formatted region (i.e add for opening delimiters and subtract for closing, when at  zero the start of the next region was found). But all of that is a future set of commits.

### What

I have only tested the changes from `ansicolor_sink-inl.h` because I develop on linux, but the Windows stuff should work as it is very similar. I don't have a version of Qt5.15 so I left `qt_sink` unchanged. The unit tests reflect where the values are being stored, and handle the condition of a malformed pattern. `example.cpp` demonstrates multiple discrete color ranges.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,8 +115,10 @@ endif()
 
 if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
     option(SPDLOG_CLOCK_COARSE "Use CLOCK_REALTIME_COARSE instead of the regular clock," OFF)
+    option(SPDLOG_EXTENDED_STLYING "Use the extended styling system and syntax" OFF)
 else()
     set(SPDLOG_CLOCK_COARSE OFF CACHE BOOL "non supported option" FORCE)
+    set(SPDLOG_EXTENDED_STLYING "non supported option" FORCE)
 endif()
 
 option(SPDLOG_PREVENT_CHILD_FD "Prevent from child processes to inherit log file descriptors" OFF)
@@ -245,6 +247,7 @@ foreach(
     SPDLOG_WCHAR_FILENAMES
     SPDLOG_NO_EXCEPTIONS
     SPDLOG_CLOCK_COARSE
+    SPDLOG_EXTENDED_STLYING
     SPDLOG_PREVENT_CHILD_FD
     SPDLOG_NO_THREAD_ID
     SPDLOG_NO_TLS
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -52,7 +52,7 @@ int main(int, char *[])
     spdlog::debug("This message should be displayed..");
 
     // Customize msg format for all loggers
-    spdlog::set_pattern("[%H:%M:%S %z] [%^%L%$] [thread %t] %v");
+    spdlog::set_pattern("[%^%H:%M:%S %z%$] [%^%L%$] [thread %t] %^%v%$");
     spdlog::info("This an info message with custom format");
     spdlog::set_pattern("%+"); // back to default format
     spdlog::set_level(spdlog::level::info);
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -26,6 +26,7 @@ void udp_example();
 void custom_flags_example();
 void file_events_example();
 void replace_default_logger_example();
+void extended_stlying();
 
 #include "spdlog/spdlog.h"
 #include "spdlog/cfg/env.h"  // support for loading levels from the environment variable
@@ -87,6 +88,7 @@ int main(int, char *[])
         custom_flags_example();
         file_events_example();
         replace_default_logger_example();
+        extended_stlying();
 
         // Flush all *registered* loggers using a worker thread every 3 seconds.
         // note: registered loggers *must* be thread safe for this to work correctly!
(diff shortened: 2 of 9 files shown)


gabime:
Thanks @drakbar for this nice idea.
Pushing to a vector in such hot path is problematic due to allocations. Also, I'm unsure if there's enough  demand for such feature to justify the extra complexity.

drakbar:
Hi @gabime, just writing an update on the status of this particular feature.

I have rolled back the previous iteration. It was something I knew I would be doing. In its place, I have significantly extended the functionality. While I have not replaced the vector sitting in `log_msg`, I think I might have a good idea as how to work around the dynamic collection. More about that later.

### new features

So currently the `color_start_formatter` accepts a style spec similar to the concept of the alignment padding spec. The style spec syntax work as such: `%{style}^`. Where style is a keyword along the lines of `bold`, `fg_green`, `bg_black`, etc. The style spec can accept multiple keywords delimited by `;`, such as: `%{style;style}^` or a concretely `%{fg_green;bold}^`.  Multiple style ranges can be defined  in discrete ranges, nested ranges, or both. Style specs can be added to any formatter but are only applied to  `color_start_formatter`. The only hard rule about style specs are they must follow a alignment padding spec if defined.

```c++
%^%$                     // uses the default log level style
%{bold;fg_green}^%$      // style spec defining bold, green foreground
%{bold}^%{fg_green}^%$   // equivalent style as above
%24{bold;fg_red}^%$      // good alignment construction
%{bold;fg_red}24^%$      // bad alignment construction
``` 
So in addition to being able to define multiple ranges, you can also choose the style that is applied to the defined region.

### limited scope

This feature is only available on linux platforms. After reading some microsoft documentation, it became apparent the old windows console which the `wincolor_sink` targets uses a different mechanism to style the console. There is a new product, windows terminal, that theoretically would be compatible with the `ansi terminal` and thus this feature. However, this feature will not be enabled on the windows platform by default.  Furthermore, a compilation switch has been add to `CMakeList.txt` for the project to enable the code that controls this feature. And so it must be explicitly turned by the user at compile time to alter the vanilla behavior and enable this functionality.

### testing

The style spec while complex is surprisingly robust. It is pretty hard to bork it. The only really hard rule that I know is the interaction with the align/padding spec. Almost anything goes between the braces, but now that I am thinking about it I haven't tried to push something like an emoji :hankey:  into it. I have added tests cases to extend the testing and the vanilla version.

### feature roadmap

With regards to the nasty `std::vector`, could it be replaced with a fixed size `std::array`, that is sized for a reasonable amount of regions? Perhaps and alternative could be some type of static array that has a global shared pointer to it (similar to a singleton). Not sure how that would play with multi-threads. Normally only read access is required, but... a call to `set_pattern` would cause a recompile that would want to start writing. @gabime Do you have any thoughts about these two ideas?

Anyways check out `example.cpp` and see the extension in action. As to the demand for this feature, at the very least I want to have it, but if you build it they will come. Thanks for your feedback!

gabime:
fmt lib [already](https://fmt.dev/latest/api.html#color-api) has such feature so I don't see a reason to add it to spdlog.

Regarding the proposal to use std::array instead of vector, it seems like a good idea. However, as previously stated, I'm skeptical if the additional feature (multiple color ranges) justifies the increased complexity. This complexity extends to several areas, for instance, in async loggers, this array would also need to be passed to the logger queue, among other things. Up to this point, I haven't encountered any requests indicating that this feature is particularly needed.

drakbar:
Well, butter me up and call me a biscuit. 

I wasn't aware that `fmt` had this feature lurking away in that header.  And look at all those colors! With all that functionality, why doesn't `spdlog::ansicolor_sink` utilize it? I suppose that just creates more of a coupling. As far as your complexity concern goes, I assume that you mean that as a maintainer as opposed to a consumer. Here are three ways of expressing the same thing.
```c++
auto red    = fmt::emphasis::bold | fmt::fg(fmt::terminal_color::red);
auto green  = fmt::emphasis::bold | fmt::fg(fmt::terminal_color::green);
auto yellow = fmt::fg(fmt::terminal_color::yellow);

// using fmt with pattern
std::string w = fmt::format(red,"{}", "%H:%M:%S %z");
std::string x = fmt::format(yellow,"thread {}", "%t");
std::string y = fmt::format(green,"{}", "%L");
std::string z = fmt::format("[{}] [{}] [{}] %v", w, x, y);
spdlog::set_pattern(z);
spdlog::info("multiple color areas");

// using fmt with custom formatter 
void my_formatter::format(const details::log_msg &msg, memory_buf_t &dest)
{
    dest.push_back('[');
    std::string ts = fmt::format(red,"{}", get_time());
    auto timestamp = spdlog::level::to_string_view(ts);
    fmt_helper::append_string_view(timestamp, dest);
    dest.push_back(']');
    dest.push_back(' ');

    dest.push_back('[');
    std::string t = fmt::format(yellow,"thread {}", "%t");
    auto thread   = spdlog::level::to_string_view(t);
    fmt_helper::append_string_view(thread, dest);

    dest.push_back('[');
    std::string l   = fmt::format(green,"{}", msg.level);
    auto level_name = spdlog::level::to_string_view(l);
    fmt_helper::append_string_view(level_name, dest);
    dest.push_back(']');
    dest.push_back(' ');
    fmt_helper::append_string_view(msg.payload, dest);
    dest.push_back('\n');
}

// using styling extensions
spdlog::set_pattern("[%{fg_red;bold}^%H:%M:%S %z%$] [%{fg_yellow}^thread %t%$] [%{bold;fg_green}^%L%$] %v");
spdlog::info("multiple color areas");
```
The styling extensions definitely look more cryptic, and it certainly abstracts away exactly what is happening. Though `fmt` with the pattern is pretty succinct once you strip out all the namespacing. I unfortunately am relegated to the world of custom formatter as I need to insert time from an alternate source. As much as I would enjoy refining on this feature, I am skeptical that any amount of effort is enough to make convincing case to progress this PR.

gabime:
Closing for now. Thanks for the effort !

PR #2817, opened 2023-07-14
Qt Sink : Allow for darker colors (for light background).

krawq:
Default are too bright if background is light(white).
[Image to see result](https://imgur.com/a/cRf9f0m)


gabime:
Thanks @krawq

PR #2819, opened 2023-07-15
implemented async_logger::sync() to solve issue #1696

SaltedReed:
This pr is to solve issue  #1696 and `async_logger::sync()` allows users to wait for the completion of logging, so they can make sure everything they want to log is safe and can be passed for further processing.

You can use it like this:
```
    auto thrpool=std::make_shared<spdlog::details::thread_pool>(100, 2);
    auto aloger=std::make_shared<spdlog::async_logger>("aloger", std::make_shared<spdlog::sinks::stdout_color_sink_mt>(), thrpool);

    aloger->info(std::string(1000,'i')); // log something large

    bool fin=aloger->sync(100,1000); // call sync() to wait until logging is flushed

    std::cout << (fin ? "done\n" : "undone\n"); // this shows after the logging on console
```

```
    /**
     * @brief block until all messages in thread_pool_ are processed
     * @param intervalMs check whether should return or not on every 'intervalMs' milliseconds
     * @param timeoutMs timeout in milliseconds. -1 means to wait indefinitely
     * @return true: all messages in thread_pool_ are processed, false: otherwise
    */
    bool sync(int intervalMs=100, int timeoutMs=-1);
```

### DIFF
--- a/include/spdlog/async_logger-inl.h
+++ b/include/spdlog/async_logger-inl.h
@@ -13,6 +13,8 @@
 #include <memory>
 #include <string>
 
+#include <chrono>
+
 SPDLOG_INLINE spdlog::async_logger::async_logger(
     std::string logger_name, sinks_init_list sinks_list, std::weak_ptr<details::thread_pool> tp, async_overflow_policy overflow_policy)
     : async_logger(std::move(logger_name), sinks_list.begin(), sinks_list.end(), std::move(tp), overflow_policy)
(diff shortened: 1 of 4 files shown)


tt4g:
IMO: Why is this feature implemented in async logger and not in thread pool?
When I saw `spdlog::async_logger::sync()`, I thought that one async logger would wait until the log messages it sent to the thread pool were processed.
However, since the thread pool is shared by multiple async loggers, it actually waits until the thread pool has completed all processing.

SaltedReed:
> IMO: Why is this feature implemented in async logger and not in thread pool? When I saw `spdlog::async_logger::sync()`, I thought that one async logger would wait until the log messages it sent to the thread pool were processed. However, since the thread pool is shared by multiple async loggers, it actually waits until the thread pool has completed all processing.

You are right. I didn't take into consideration the sharing of thread_pool by multiple async_loggers. I'll fix that.

walkerlala:
I think we should reimplement the async_logger so that it has its own message queue that could be pulled continously by worker threads. Having the queue inside a thread pool make the whole thing more complicated and unnecessary.

By the way, I think the "async_logger::flush()" interface is broken. A "flush" operation should have the semantic of flushing the in-memory data into file (into the OS file-cache at least). A "sync" operation should persist the in-memory cache data into disk. These are long established conventions in the C / C++ / POSIX world and the database world: https://man7.org/linux/man-pages/man3/fflush.3.html  , https://man7.org/linux/man-pages/man2/fsync.2.html , and spdlog's implementation is counter-intuitive.  This is why there are so many issues and discussions all the way up.  https://github.com/gabime/spdlog/discussions/2982  https://github.com/gabime/spdlog/issues/1696  https://github.com/gabime/spdlog/pull/2819

PR #2836, opened 2023-07-29
fix:Modify the bug of qtextedit;

mahuifa:
1.Modifying qtextedit cannot display Chinese problems;
    2.When modifying the qtextedit display, it will not automatically scroll to the bottom problem.

### DIFF
--- a/include/spdlog/sinks/qt_sinks.h
+++ b/include/spdlog/sinks/qt_sinks.h
@@ -162,7 +162,7 @@ protected:
 
         const string_view_t str = string_view_t(formatted.data(), formatted.size());
         // apply the color to the color range in the formatted message.
-        auto payload = QString::fromLatin1(str.data(), static_cast<int>(str.size()));
+        auto payload = QString::fromUtf8(str.data(), static_cast<int>(str.size()));
 
         invoke_params params{max_lines_,             // max lines
             qt_text_edit_,                           // text edit to append to
@@ -215,6 +215,7 @@ protected:
         // insert the text after the color range with default format
         cursor.setCharFormat(params.default_color);
         cursor.insertText(params.payload.mid(params.color_range_end));
+        params.q_text_edit->setTextCursor(cursor);
     }
 
     QTextEdit *qt_text_edit_;


include/spdlog/sinks/qt_sinks.h
```diff
@@ -162,7 +162,7 @@ class qt_color_sink : public base_sink<Mutex>
 
         const string_view_t str = string_view_t(formatted.data(), formatted.size());
         // apply the color to the color range in the formatted message.
-        auto payload = QString::fromLatin1(str.data(), static_cast<int>(str.size()));
+        auto payload = QString::fromUtf8(str.data(), static_cast<int>(str.size()));
```
gabime, line 165: The problem with utf8 is that the color position might be be wrong and needs to get calculated precisely somehow.

mahuifa:
For Chinese users, displaying Chinese is more important than displaying colors; Or can it be set as optional.
&nbsp;


白露
***@***.***



&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Gabi ***@***.***&gt;; 
发送时间: 2023年7月29日(星期六) 下午5:33
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [gabime/spdlog] fix:Modify the bug of qtextedit; (PR #2836)





 
@gabime commented on this pull request.
 
 
In include/spdlog/sinks/qt_sinks.h:
 &gt; @@ -162,7 +162,7 @@ class qt_color_sink : public base_sink<Mutex&gt;            const string_view_t str = string_view_t(formatted.data(), formatted.size());          // apply the color to the color range in the formatted message. -        auto payload = QString::fromLatin1(str.data(), static_cast<int&gt;(str.size())); +        auto payload = QString::fromUtf8(str.data(), static_cast<int&gt;(str.size()));  
The problem with utf8 is that the color position might be be wrong and needs to get calculated precisely somehow.

gabime:
If you dont need colors, you can use the `qt_sink_mt` sink. It [uses](https://github.com/gabime/spdlog/blob/2ee8bac78e6525a8ad9a9196e65d502ce390d83a/include/spdlog/sinks/qt_sinks.h#L55) utf8

PR #2839, opened 2023-07-31
Set CMAKE_BUILD_TYPE only for single-config generators

moritz-h:
spdlog currently sets CMAKE_BUILD_TYPE also when a CMake multi-config generator is used (e.g. Visual Studio). This adds an unneccessary CMake cache entry in that case.


gabime:
Thanks @moritz-h

PR #2846, opened 2023-08-07
Provide spdlog_header_only in build directory export

### LINKED ISSUE #2678: CMake: Build directory `spdlogConfigTargets.cmake` doesn't provide a `spdlog::spdlog_header_only` target
The installed version of `spdlogConfigTargets.cmake` includes both the `spdlog::spdlog` and `spdlog::spdlog_header_only` targets. The build directory version only provides the compiled library target, which causes issues for dowstream consumers of the header only version.

This is caused by a mismatch between what targets are exported via the `install` and `export` commands.
Updating line https://github.com/gabime/spdlog/blob/v1.x/CMakeLists.txt#L339 to also include spdlog_header_only would fix this issue.

robertmaynard:
Fixes #2678

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -344,7 +344,7 @@ if(SPDLOG_INSTALL)
     # ---------------------------------------------------------------------------------------
     # Install CMake config files
     # ---------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @robertmaynard

PR #2853, opened 2023-08-10
Add SPDLOG_VERSION check support

kegechen:
You can use SPDLOG_VERSION to select the latest spdlog features
where available while falling back to older implementations otherwise.
Using SPDLOG_TO_VERSION() for the value to compare with is recommended.
for Example:
```c++
 void sink_it_(const details::log_msg &msg) override
 {
 #if SPDLOG_VERSION < SPDLOG_TO_VERSION(1,4,0)
     fmt::memory_buffer formatted;
 #else
     memory_buf_t formatted;
 #endif
     sinks::base_sink<Mutex>::formatter_->format(msg, formatted);
     // bala bala...
 }


gabime:
Thanks. Since the macro by itself doesn't actually check anything, I think it should be named differently, perhaps `SPDLOG_GET_VERSION(..)` or similar.

kegechen:
> Thanks. Since the macro by itself doesn't actually check anything, I think it should be named differently, perhaps `SPDLOG_GET_VERSION(..)` or similar.

https://doc.qt.io/qt-6/qtversionchecks-proxy.html#QT_VERSION_CHECK
that's how Qt is used, but what you say makes sense，I prefer SPDLOG_MAKE_VERSION

gabime:
But now it might confuse users with the make tool :)
How about `SPDLOG_TO_VERSION()` ?

kegechen:
> But now it might confuse users with the make tool :) How about `SPDLOG_TO_VERSION()` ?

done

gabime:
Thanks @kegechen

PR #2858, opened 2023-08-18
Modify the condition of SPDLOG_CONSTEXPR_FUNC to match that of fmt

### LINKED ISSUE #2856: Different use_constexpr definitions in spdlog and fmt causes compile time error in NVCC (and possibly other compilers)
# Description
~~~
~~~

# Cause

# Fix

ShujianQian:
* fix the issue where constexpr function in spdlog may call non-constexpr function in the bundled fmt because FMT_USE_CONSTEXPR is not defined

fix #2856

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -16,6 +16,44 @@
 #include <functional>
 #include <cstdio>
 
+#ifdef __has_feature
+#    define SPDLOG_HAS_FEATURE(x) __has_feature(x)
+#else
+#    define SPDLOG_HAS_FEATURE(x) 0
+#endif
+
+#ifdef __ICL
+#    define SPDLOG_ICC_VERSION __ICL
+#elif defined(__INTEL_COMPILER)
+#    define SPDLOG_ICC_VERSION __INTEL_COMPILER
+#else
+#    define SPDLOG_ICC_VERSION 0
+#endif
+
+#ifdef _MSC_VER
+#    define SPDLOG_MSC_VERSION _MSC_VER
+#else
+#    define SPDLOG_MSC_VERSION 0
+#endif
+
+#if defined(__clang__) && !defined(__ibmxl__)
+#    define SPDLOG_CLANG_VERSION (__clang_major__ * 100 + __clang_minor__)
+#else
+#    define SPDLOG_CLANG_VERSION 0
+#endif
+
+#ifdef _MSVC_LANG
+#    define SPDLOG_CPLUSPLUS _MSVC_LANG
+#else
+#    define SPDLOG_CPLUSPLUS __cplusplus
+#endif
+
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER) && !defined(__NVCOMPILER)
+#    define SPDLOG_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
+#else
+#    define SPDLOG_GCC_VERSION 0
+#endif
+
 #ifdef SPDLOG_USE_STD_FORMAT
 #    include <version>
 #    if __cpp_lib_format >= 202207L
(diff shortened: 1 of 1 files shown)


gabime:
Thanks, however thats way too much additions and new macros for this simple fix.

ShujianQian:
I was hoping to match the condition in fmt for ease of maintenance in the future. Would you prefer a simpler condition for the SPDLOG_CONSTEXPR_FUNC macro or marking the affected function as inline instead?

gabime:
Since this is nvcc specific, simple ifdef nvcc is sufficient.

ShujianQian:
I believe this will affect other compilers too. The compile time error will happen whenever SPDLOG_CONSTEXPR_FUNC is defined but FMT_USE_CONSTEXPR is not. 

The FMT_USE_CONSTEXPR condition in the bundled fmt is the following:
~~~cpp
#  if (FMT_HAS_FEATURE(cxx_relaxed_constexpr) || FMT_MSC_VERSION >= 1912 || \
       (FMT_GCC_VERSION >= 600 && FMT_CPLUSPLUS >= 201402L)) &&             \
      !FMT_ICC_VERSION && !defined(__NVCC__)
~~~

The issue is also very visible because it happens on any kind of logging as simple as `spdlog::info("text");`.

If you believe `ifdef nvcc` is sufficient, I can compose another PR.

gabime:
> @ShujianQian I believe this will affect other compilers too

So how come it is not a common problem and wasn't reported many times so far?

ShujianQian:
I guess the condition is rather hard to meet and some compilers do not raise an error in some cases. However, I'm able to verify that the same issue happens on gcc-5 with the following source file:

~~~cpp
#include "spdlog/spdlog.h"

int main(int argc, char **argv)
{
        spdlog::info("info");
        return 0;
}
~~~

~~~
> gcc-5 --version
gcc-5 (Homebrew GCC 5.5.0_8) 5.5.0
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

> gcc-5 ./test_spdlog.cpp -I ./spdlog/include -o test_spdlog -std=c++14
In file included from ./spdlog/include/spdlog/spdlog.h:12:0,
                 from ./test_spdlog.cpp:1:
./spdlog/include/spdlog/common.h: In function 'constexpr spdlog::string_view_t spdlog::details::to_string_view(const memory_buf_t&)':
./spdlog/include/spdlog/common.h:351:42: error: call to non-constexpr function 'const T* fmt::v9::detail::buffer<T>::data() const [with T = char]'
     return spdlog::string_view_t{buf.data(), buf.size()};
                                          ^
~~~

Some versions of ICC (which is another corner case in the FMT_CONSTEXPR condition) allows constexpr functions to call non-constexpr functions (as shown in [this example](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIMxqkrgAyeAyYAHI%2BAEaYxCAAbKQADqgKhE4MHt6%2B/oGp6Y4CoeFRLLHxSXaYDplCBEzEBNk%2BfgG2mPZFDPWNBCWRMXGJtg1NLbntCmP9YYPlwwkAlLaoXsTI7BxhBADUBJjTXBBLuyYA7FYaAIK7u8SYBOsMuxomAMxX1xcAIiY3/2uaAY00wqmSxF2O32hwIJzOl0BdweT2ILzuMKOJw%2BX1%2BHBWtE4AFZeH4OFpSKhOG5rNZdgo1htMGczO8eKQCJp8SsANZ5fScSSkrmUzi8BQgQKc8n40hwWBINAsZJ0OLkShKlX0eJ4ZDILhJLAAN11mAAanhMAB3ADyyUYnHZNFoB2IEog0RF0TCjQAno7eN7mMRfTbotoatL2Uq2IIbQxaP6ZaQsNEvMA3GJaBLuLwsCxDMBxMn8A9akbDiKwTUvAcA%2BRBJ0RbQ8NFiH6PFh6wRiHgWAGVlQDMAFBbrXaHbmZIIRGJ2FJp/IlGoRbouPpCyhaZZ9K2JZAVqhkt0cwBaaboD4/UyWaxSXioCvEXtYfcnDpdTIuBjuTytPQhHMZQVHoBQZAIEx%2BOuYHdAMwHDOu1S1AIvTjH%2BuSIZ0kZ1DMcFDPEiEzJBejTH0eELARKwMusmwSASxLCsmVIcLsqgABwJKeCSSFCeq7AaAB0GgCe8uwQLghAkCybJLLw0paEsvL8oSHBCqQZIUsx4qShyXIrPKiAgGsBDJLW6oQJqqrEBErBbOxnHcbswB8QAnAJZi8Jg%2BBEC%2Bej8DOojiAu/lLio6jJmupBWu2yQDgKHAkupIrMTatamXsqBUKxHFcTxurIPxCRCSJYkeMqVnSVwsm6TKimkHy7QqWpGkPmKtg6fJsoGRAiqoOV2rmZZ2ogPlARmFwLmBMappjra9pkk6dCuu6nrJkGfr1utIZhhGDj1jGjAEPGiYiqm6aZrQ2b1vmhbFhSpbYRWOYUtWyC1ls7I7E2yYtm2HYYFsFI9n2cVDkwI6zROC28CFs5BdIIWKGFq7%2BBuRhbreO6/W%2Bh7HpkOYPk%2BL6VgeH7Yc4ECuMRRLSIBpT4SA0gwZk1NM2k4EMORIHSEh3Soc06F%2BDTZPIT0uFAQzPNEYLIDC6RTRc8MkhUYytFVfFiUtaKLH2blvEFWNXACW5GhiRJPmVdVnV1Q1gRNYxmltRKUp6XKMCGcZ6WDX1WpxDZbCcLrjn5bshvG0JnneSQeDoH5shw/OCOyEjK4Raj0VMLFub0QlDutRwqUmbWuyZdlDl5XxYcm6VPsVeY7xmFbelKY1gp59r2ku7VOceUlTFtdbKxPukziSEAA)). I have verified that spdlog compiles with the latest ICC.

I'm not able to test on MSVC.

gabime:
I don't mimd a minimal fix, with minimal or no new macros. sodlog defines too many macros already imo.

PR #2859, opened 2023-08-18
Modify the condition of SPDLOG_CONSTEXPR_FUNC to match that of fmt

### LINKED ISSUE #2856: Different use_constexpr definitions in spdlog and fmt causes compile time error in NVCC (and possibly other compilers)

ShujianQian:
fix the issue where constexpr function in spdlog may call non-constexpr function in the bundled fmt because FMT_USE_CONSTEXPR is not defined.

fix #2856

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -64,15 +64,21 @@
 #if defined(_MSC_VER) && (_MSC_VER < 1900)
 #    define SPDLOG_NOEXCEPT _NOEXCEPT
 #    define SPDLOG_CONSTEXPR
-#    define SPDLOG_CONSTEXPR_FUNC inline
 #else
 #    define SPDLOG_NOEXCEPT noexcept
 #    define SPDLOG_CONSTEXPR constexpr
-#    if __cplusplus >= 201402L
-#        define SPDLOG_CONSTEXPR_FUNC constexpr
-#    else
-#        define SPDLOG_CONSTEXPR_FUNC inline
-#    endif
+#endif
+
+#ifndef __has_feature
+#    define __has_feature(x) 0
+#endif
+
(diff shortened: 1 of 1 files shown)


include/spdlog/common.h
```diff
@@ -64,15 +64,21 @@
...
 #    define SPDLOG_CONSTEXPR constexpr
-#    if __cplusplus >= 201402L
-#        define SPDLOG_CONSTEXPR_FUNC constexpr
-#    else
-#        define SPDLOG_CONSTEXPR_FUNC inline
-#    endif
+#endif
+
+#ifndef __has_feature
+#    define __has_feature(x) 0
```
gabime, line 73: Please don't define standard macros, even if to 0. It will surely break users code.

gabime (changes requested):

include/spdlog/common.h
```diff
@@ -64,15 +64,21 @@
...
+#endif
+
+#if (__has_feature(cxx_relaxed_constexpr) || (defined(_MSC_VER) && (_MSC_VER >= 1912)) ||                                                  \
+     (defined(__GNUC__) && __GNUC__ >= 6 && defined(__cplusplus) && __cplusplus >= 201402L)) &&                                            \
+    !defined(__ICL) && !defined(__INTEL_COMPILER) && !defined(__NVCC__)
```
gabime, line 78: Can be simplified greatly. No need to check the defined(..), only the comparisons since undefined macro values equal to 0. See https://stackoverflow.com/questions/5085392/what-is-the-value-of-an-undefined-constant-used-in-if

Also please add comment describing the purpose of those checks.It will be hard to understand why without description.

kkraus14:
This seemed stale so I have created a new PR based on this PR to try to help it over the finish line: https://github.com/gabime/spdlog/pull/2901

PR #2862, opened 2023-08-22
Fix encoding issue in qt_sinks

neothenil:
`QString::fromLatin1` doesn't handle non-latin characters properly. When applied to Chinese characters, the message appeared in QTextEdit looks like corrupted.
image

This PR fixes this issue.
image

### DIFF
--- a/include/spdlog/sinks/qt_sinks.h
+++ b/include/spdlog/sinks/qt_sinks.h
@@ -164,10 +164,13 @@ protected:
         // apply the color to the color range in the formatted message.
         auto payload = QString::fromUtf8(str.data(), static_cast<int>(str.size()));
         // convert color ranges from byte index to character index.
+        int color_range_start, color_range_end;
+        color_range_start = static_cast<int>(msg.color_range_start);
+        color_range_end = static_cast<int>(msg.color_range_end);
         if (msg.color_range_start < msg.color_range_end) {
             QByteArray bytes(str.data(), str.size());
-            msg.color_range_start = QString::fromUtf8(bytes.left(msg.color_range_start)).size();
-            msg.color_range_end = QString::fromUtf8(bytes.left(msg.color_range_end)).size();
+            color_range_start = QString::fromUtf8(bytes.left(msg.color_range_start)).size();
+            color_range_end = QString::fromUtf8(bytes.left(msg.color_range_end)).size();
         }
 
         invoke_params params{max_lines_,             // max lines
@@ -175,8 +178,8 @@ protected:
             std::move(payload),                      // text to append
             default_color_,                          // default color
             colors_.at(msg.level),                   // color to apply
-            static_cast<int>(msg.color_range_start), // color range start
-            static_cast<int>(msg.color_range_end)};  // color range end
+            color_range_start,                       // color range start
+            color_range_end};                        // color range end
 
         QMetaObject::invokeMethod(
             qt_text_edit_, [params]() { invoke_method_(params); }, Qt::AutoConnection);
--- a/include/spdlog/sinks/qt_sinks.h
+++ b/include/spdlog/sinks/qt_sinks.h
@@ -66,14 +66,15 @@ private:
 // Color location is determined by the sink log pattern like in the rest of spdlog sinks.
 // Colors can be modified if needed using sink->set_color(level, qtTextCharFormat).
 // max_lines is the maximum number of lines that the sink will hold before removing the oldest lines.
-// Note: Only ascii (latin1) is supported by this sink.
+// By default, only ascii (latin1) is supported by this sink. Set is_utf8 to true if utf8 support is needed.
 template<typename Mutex>
 class qt_color_sink : public base_sink<Mutex>
 {
 public:
-    qt_color_sink(QTextEdit *qt_text_edit, int max_lines, bool dark_colors=false)
+    qt_color_sink(QTextEdit *qt_text_edit, int max_lines, bool dark_colors=false, bool is_utf8=false)
         : qt_text_edit_(qt_text_edit)
         , max_lines_(max_lines)
+        , is_utf8_(is_utf8)
     {
         if (!qt_text_edit_)
         {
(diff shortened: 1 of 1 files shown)


gabime:
Duplicate of the rejected #2836. Colors ranges must an addressed somehow

neothenil:
Oh, I see. The problem is color ranges are byte indices, but they are passed to `QString::left` and `QString::mid` where the expected parameter is character index. How about converting color ranges from byte index to character index before passing to `invoke_method_` like [this commit](https://github.com/gabime/spdlog/pull/2862/commits/4a50f00350cd6d880ae6c547a297ab1b7c950745) does?

gabime:
Not sure. This should be debugged. And anyway msg contents should not be changed directly since they may be used by other sinks

neothenil:
For my use case, the modification works fine. The characters in QTextEdit are normal. The color ranges which were incorrect before the last commit also appears correct. And as long as the underlying string is UTF8-encoded, this modification should work well. In case the changed msg contents affect other sinks, I can make the converted color ranges local variables without changing msg directly.

include/spdlog/sinks/qt_sinks.h
```diff
@@ -162,15 +162,24 @@ class qt_color_sink : public base_sink<Mutex>
...
         const string_view_t str = string_view_t(formatted.data(), formatted.size());
         // apply the color to the color range in the formatted message.
-        auto payload = QString::fromLatin1(str.data(), static_cast<int>(str.size()));
+        auto payload = QString::fromUtf8(str.data(), static_cast<int>(str.size()));
+        // convert color ranges from byte index to character index.
+        int color_range_start, color_range_end;
+        color_range_start = static_cast<int>(msg.color_range_start);
+        color_range_end = static_cast<int>(msg.color_range_end);
+        if (msg.color_range_start < msg.color_range_end) {
+            QByteArray bytes(str.data(), str.size());
```
gabime: No need to create a deep copy of QByteArray. You can use directly
```c++
color_range_start = QString::fromUtf8(str.data(), msg.color_range_start).size();
color_range_end = QString::fromUtf8(str.data(), msg.color_range_end).size();

```

gabime (changes requested):

include/spdlog/sinks/qt_sinks.h
neothenil: Updated.

tt4g (commented):
Sorry, a link to the source code was missing. See [next comment](https://github.com/gabime/spdlog/pull/2862#discussion_r1311057663).

include/spdlog/sinks/qt_sinks.h
```diff
@@ -162,15 +162,23 @@ class qt_color_sink : public base_sink<Mutex>
...
         // apply the color to the color range in the formatted message.
-        auto payload = QString::fromLatin1(str.data(), static_cast<int>(str.size()));
+        auto payload = QString::fromUtf8(str.data(), static_cast<int>(str.size()));
+        // convert color ranges from byte index to character index.
+        int color_range_start, color_range_end;
+        color_range_start = static_cast<int>(msg.color_range_start);
+        color_range_end = static_cast<int>(msg.color_range_end);
+        if (msg.color_range_start < msg.color_range_end) {
+            color_range_start = QString::fromUtf8(str.data(), msg.color_range_start).size();
+            color_range_end = QString::fromUtf8(str.data(), msg.color_range_end).size();
```
tt4g: Why use `QString`?
I think you can use `QStringRef` here.

See: https://stackoverflow.com/questions/62910966/how-to-use-qstringview-or-qstringref-in-qt-widgets

neothenil: To convert byte index/number to character index/number, there must be an API that takes byte index/number as input and returns character index/number. This is what `QString::fromUtf8(const char *str, qsizetype size)` and `QString::size()` can accomplish. But I don't see any `QStringRef` API taking byte index/number as input. Or I missed something.

As for `QStringView`, it assumes the string is UTF16-encoded. So not suitable here.

I also looked for `QUtf8StringView`. But can't find any API returning character index/number. (I thought it should return character index/number, but the following test code showed me wrong.)

```C++
    QString test = "测试字符串";
    qDebug() << test.size();    // return 5
    QByteArray bytes = test.toUtf8();
    QUtf8StringView testView(bytes.data(), bytes.size());
    qDebug() << testView.size();    // return 15
```

gabime: I think we need to add  to the ctor a bool flag to indicate if the log is utf8 or not(default false). This way all those expensive conversions can be avoided if the utf8 option is false.

tt4g: I don't know what version of Qt we should support, can't we use the `QByteArrayView` that was added in Q

neothenil: I have no performance optimization experience. Does this operation really that expensive?

tt4g: Memory consumption is proportional to the length of log messages.
We want to avoid memory copying just to find the index of a character as much as possible.

neothenil: I find an efficient way for counting character number in UTF8 encoded string. (See: https://stackoverflow.com/questions/4063146/getting-the-actual-length-of-a-utf-8-encoded-stdstring)

Do you think we can replace `QString::fromUtf8` and `QString::size` with this code?

tt4g: If you know the character to look for, you can use `QString::indexOf()`.

https://doc.qt.io/qt-6/qstring.html#indexOf-1

neothenil: But how can I know which one is what I want if there are multiple occurrences of this character?

tt4g: If a user added the same characters as the color code to a message, it would not be distinguishable from the color code generated by the spdlog format.

BTW, I don`t understand why you don`t use [`QStringRef::QStringRef(const [QString](https://doc.qt.io/qt-6/qstring.html) *string, int position, int length)`](https://doc.qt.io/qt-5/qstringref.html#QStringRef-1) (I suggested it once).
Are you saying that the following code does not work?

```cpp
color_range_start = QStringRef(&str msg.color_range_start).size();
color_range_end = QStringRef(&str, msg.color_range_end).size();
```

neothenil: No, `QStringRef` doesn't work like you expected. After changing to use `QStringRef`, the color ranges are incorrect.
```cpp
color_range_start = QStringRef(&payload, 0, msg.color_range_start).size();
color_range_end = QStringRef(&payload, 0, msg.color_range_end).size();
```

Test code:
```cpp
auto logger = spdlog::qt_color_logger_mt("qt_logger", ui->textEdit, 500);
logger->set_pattern("%Y-%m-%d %H:%M:%S [测试]: %^%v%$[结束]");
logger->info("测试字符串");
```

Result:
image

Expected:
image

tt4g: It seems that the API supported by Qt is incompatible with UTF-8.
It seems that `QString::fromUtf8()` has to be used.
Thank you for your time.

tt4g (approved):

gabime:
This results in 3 `QString::fromUtf8()`  per each log call. It is expensive and should be avoided if only ascii/latin1 is used.

Please add a bool flag `is_utf8` to the constructor and perform  the `fromUtf8()` only if set to `true`.

neothenil:
The bool flag `is_utf8` is added on your demand.

gabime:
Thanks @neothenil and @tt4g . Merged.

PR #2870, opened 2023-08-29
Fixing deprecation warning Wdeprecated-literal-operator on LLVM 16.0.6

burcelee:
See https://reviews.llvm.org/rG5ce5e983f82c802e44faa8ed42d605d70c045ba9

Since LLVM 16.0.6 (which I am using via Emscripten), the following code can generate a warning due to use the deprecated syntax of the literal operator.

This fixed the issue in my project. Please let me know if my PR is improperly formatted or there's anything else I need to do.

### DIFF
--- a/include/spdlog/fmt/bundled/format.h
+++ b/include/spdlog/fmt/bundled/format.h
@@ -4165,7 +4165,7 @@ template <detail_exported::fixed_string Str> constexpr auto operator""_a() {
   return detail::udl_arg<char_t, sizeof(Str.data) / sizeof(char_t), Str>();
 }
 #  else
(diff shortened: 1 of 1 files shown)


tt4g:
All source code under the directory `include/spdlog/fmt/bundled/` is copied from the fmt library.
Please reopen it in [fmt repository](https://github.com/fmtlib/fmt).

tt4g:
Rejected this change in fmt: fmtlib/fmt#3608

Current PR: fmtlib/fmt#3610

burcelee:
Thanks for the heads up. Sounds like this change is redundant then, so we can probably close?

tt4g:
Yes. close it.

PR #2871, opened 2023-08-30
Remove obsolete part from cmake configuration files

albert-github:
Spdlog requires CMake version 3.10 or higher (by means of `cmake_minimum_required(VERSION 3.10...3.21)`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,9 +129,7 @@ option(
 option(SPDLOG_DISABLE_DEFAULT_LOGGER "Disable default logger creation" OFF)
 
 # clang-tidy
-if(${CMAKE_VERSION} VERSION_GREATER "3.5")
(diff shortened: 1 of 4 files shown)


albert-github:
In the `bench/CMakeLists.txt` there is a test against `CMAKE_VERSION` `3.11` maybe increase the minimum CMake version so also this test can be removed.

gabime:
> In the bench/CMakeLists.txt there is a test against CMAKE_VERSION 3.11 maybe increase the minimum CMake version so also this test can be removed.

Ok. please update the pr

albert-github:
@gabime Do you mean to increase the minimum version, but to which version, 3.11 of 3.12 ...?

gabime:
3.11

gabime:
Thanks @albert-github

PR #2883, opened 2023-09-25
Added a function to add callbacks that are called when a logger is registered

oddko:
Use case:

We have a collection of libraries that use spdlog in a somewhat "bare" fashion (setting up a colored multithreaded stdout log for the most part).

We wish to be able to redirect those logs to a file or a custom sink to view within a UI and whatnot. While we can easily do that for logs that have already been setup through the registry and automatic registration (pretty much all libraries directly linked to, that initialize it in a static variable), using the already existing `apply_all` function, this cannot work for libraries that may be loaded dynamically later as plugins (e.g., in our case, mostly QML plugins in the Qt framework) since they'll be loaded after our "main" routine. Calling `apply_all` late (thus potentially missing out on early log) or doing some hack with a timer that repeatedly calls `apply_all` don't seem like good solutions.

Thus we wish to be able to register functions that are called when a logger is registered, mostly to be able to tweak its sinks vector and redirect the log.

Thank you in advance for your comments in hopes this feature suits you. Have a nice day !

### DIFF
--- a/include/spdlog/details/registry.h
+++ b/include/spdlog/details/registry.h
@@ -92,7 +92,7 @@ public:
 
     void apply_logger_env_levels(std::shared_ptr<logger> new_logger);
 
-    void add_on_registration_callback(const std::function<void(const std::shared_ptr<logger>&)>& callback);
+    void add_on_registration_callback(const std::function<void(std::shared_ptr<logger>)>& callback);
 
     void drop_all_on_registration_callbacks();
 
@@ -116,7 +116,7 @@ private:
     std::shared_ptr<logger> default_logger_;
     bool automatic_registration_ = true;
     size_t backtrace_n_messages_ = 0;
-    std::vector<std::function<void(const std::shared_ptr<logger>&)>> on_registration_callbacks_;
+    std::vector<std::function<void(std::shared_ptr<logger>)>> on_registration_callbacks_;
 };
 
 }  // namespace details
--- a/include/spdlog/details/registry.h
+++ b/include/spdlog/details/registry.h
@@ -92,6 +92,10 @@ public:
 
     void apply_logger_env_levels(std::shared_ptr<logger> new_logger);
 
+    void add_on_registration_callback(const std::function<void(const std::shared_ptr<logger>&)>& callback);
+
+    void drop_all_on_registration_callbacks();
+
 private:
     registry();
     ~registry();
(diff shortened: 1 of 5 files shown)


gabime:
Thanks. Problem is the registry is likely to be removed in spdlog v2.x. I don't mind adding this to v1.x though. 
If you still want it in v1.x, please use `std::vector` instead of `std::list`, as this is what used across the code base already.

oddko:
Hi, thanks for your comment !

No prob for the code change, will do, and yes I'm interested in this feature.

Any insight into what said design change for v2.x would be ? Do you intend to replace it by something else or do you want to leave it to people to track their loggers with their own DYI solution ?

gabime:
I think to let people to track their loggers with their own say. There are just too many ways  to do this, and it feels  out of spdlog’s scope (not to mention it promotes a global state).

include/spdlog/details/registry.h
```diff
@@ -112,6 +116,7 @@ class SPDLOG_API registry {
     std::shared_ptr<logger> default_logger_;
     bool automatic_registration_ = true;
     size_t backtrace_n_messages_ = 0;
+    std::vector<std::function<void(const std::shared_ptr<logger>&)>> on_registration_callbacks_;
```
gabime: Better not to pass ref to shared_ptr here.

oddko: Fixed

oddko:
Managing a collection of loggers is probably better done by the user of the library, however a somewhat lightweight tracker of all loggers that were created inside of the library eases greatly the integration of existing loggers in "low-level" libraries when they're integrated into a bigger application (and that could'nt/should'nt have known how it'd be integrated inside said top-level application).

Said application may want to uniformize logging formats, print everything to a file, provide a live view of all logs that can be filtered by something better than regex and whatnot. This forces the use of a global state anyway, and can be easily done in this library since loggers and their sinks are somewhat decoupled (modifying the lists of sinks has its caveats obviously, but unless the library aggressively spins other threads, which is kind of dangerous on its own, should be fine).

While it wouldn't take us a century to cultivate our own vector of loggers somewhere, I think it'd be a missed opportunity to prevent integration/redirection of a logger from a library outside our own internal ecosystem of library (if it were to happen).

gabime:
Sorry @oddko, I had to revert this .
While playing with this feature I got myself to a deadlock way too easily since the callbacks are protected under the `logger_map_mutex_`, which is used throughout the registry in many places.
All it takes is for the callback to invoke any spdlog:: function that is protected by the mutex, and a deadlock happens.

To solve this would mean to add a new recursive mutex to the registry class, and add even more complexity to the code which I am not interested to indroduce.

oddko:
Ah indeed if you call functions of the registry itself, bummer, and I guess making the logger map mutex recursive would be a step too far, no big deal if you intend to remove the registry alltogether anyway.

EDIT: doesn't the current `apply_all` have the same issue anyway ?

PR #2895, opened 2023-10-09
Hide non-API functions

urisimchoni:
Usage of -fvisibility=hidden is the recommended default for DSO's, and is harmless in statically-linked objects.

If the spdlog build is for a shared library, then API symbols have SPDLOG_API marking on them, and cause the symbol to be exported, with other symbols now hidden by default.

If the build is for a static library with position-independent code, then spdlog is to be embedded as implementation detail in another shared library or executable. In that case we don't want to export any symbol. Conveniently, SPDLPG_API in that case resolves to nothing, so with the default of hidden all symbols are made hidden.


gabime:
Thanks. Why only in PIC ? I think it can be used without pic as well.

urisimchoni:
> Thanks. Why only in PIC ? I think it can be used without pic as well.

I suppose it's not really tied to PIC, although the context in which it matters is dynamically-linked executables, which require PIC. Without the condition, the -fvisibility=hidden flag is given to static builds as well, where it should have no effect, so I suppose I can remove the condition.

gabime:
We need to be careful since the bundled fmt use [default](https://github.com/fmtlib/fmt/blob/130cf54cbc77abf0c8200a1206638cb70c9e26f1/include/fmt/core.h#L201) visibilty. Not sure what would happen. i.e. If users of spdlog will be able to use the  bundled fmt lib. Could you check?

urisimchoni:
If I understand correctly, the concern is that an aplication should be able to use fmt::XXX embedded in the shared libspdlog.so.
I ran the following quick check:
* Add the following lines to example.cpp:
``` 
    auto s = fmt::format("i={}",1);
    printf("%s\n", s.c_str());
```
* Build spdlog with `cmake -DSPDLOG_BUILD_SHARED=ON -DSPDLOG_BUILD_PIC=ON ../spdlog && make -j`
* Run the example

I can also see that libspdlog.so built that way exports some fmt symbols.

PR #2715, opened 2023-04-22
os-inl.h: fix for missing pthread_threadid_np

barracuda156:
`pthread_threadid_np` is missing on macOS < 10.6, as well as on any PowerPC (including 10.6 PPC and 10.6.8 Rosetta, therefore additional condition).

PR #2897, opened 2023-10-11
Fix OS availability check of pthread_threadid_np for iOS

lowdesertpunk:
The OS availability check of `pthread_threadid_np` (introduced in https://github.com/gabime/spdlog/pull/2715) is not considering other Apple OSes, where the macOS specific macros are not defined as such these C preprocessor undefined values evaluate to zero causing the #if expression to evaluate in an unexpected way (e.g. unnecessary weak linking check on iOS).


gabime:
Thanks @lowdesertpunk

PR #2899, opened 2023-10-11
Add .git-blame-ignore-revs to ignore clang-format related commits

lowdesertpunk:
Recent clang-format related commits made have overtaken the blame view making history harder to navigate. GitHub supports the [`ignore-revs-file`](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt) via the [`.git-blame-ignore-revs` file](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).


gabime:
Thanks @lowdesertpunk

PR #2901, opened 2023-10-11
Match SPDLOG_CONSTEXPR_FUNC to FMT_CONSTEXPR

### LINKED ISSUE #2856: Different use_constexpr definitions in spdlog and fmt causes compile time error in NVCC (and possibly other compilers)

kkraus14:
Fixes #2856.

PR based on #2858 and #2859.

Fixes the issue where a constexpr function in spdlog may call a non-constexpr function in fmt because `FMT_CONSTEXPR` is not defined or is defined differently for a given compiler.

Ideally, a test could be added of building a `.cu` file using nvcc and add that to CI to prevent nvcc breakages moving forward.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -70,13 +70,19 @@
     #define SPDLOG_CONSTEXPR constexpr
 #endif
 
-// If building with fmt SPDLOG_CONSTEXPR_FUNC needs to be set the same as FMT_CONSTEXPR
-// to avoid situations where a constexpr function in spdlog could end up calling
-// a non-constexpr function in fmt depending on the compiler
-#ifdef FMT_CONSTEXPR
-    #define SPDLOG_CONSTEXPR_FUNC FMT_CONSTEXPR
-#else
+// If building with std::format, can just use constexpr, otherwise if building with fmt
+// SPDLOG_CONSTEXPR_FUNC needs to be set the same as FMT_CONSTEXPR to avoid situations where
+// a constexpr function in spdlog could end up calling a non-constexpr function in fmt
+// depending on the compiler
+// If fmt determines it can't use constexpr, we should inline the function instead
+#ifdef SPDLOG_USE_STD_FORMAT
     #define SPDLOG_CONSTEXPR_FUNC constexpr
+#else  // Being built with fmt
+    #if FMT_USE_CONSTEXPR
+        #define SPDLOG_CONSTEXPR_FUNC FMT_CONSTEXPR
+    #else
+	#define SPDLOG_CONSTEXPR_FUNC inline
+    #endif
 #endif
 
 #if defined(__GNUC__) || defined(__clang__)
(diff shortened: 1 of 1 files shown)


kkraus14:
An alternative approach that might work here is to reuse `FMT_CONSTEXPR` with something like:
```cpp
#ifdef FMT_CONSTEXPR
    #define SPDLOG_CONSTEXPR_FUNC FMT_CONSTEXPR
#else
    #if __cplusplus >= 201402L
        #define SPDLOG_CONSTEXPR_FUNC constexpr
    #else
        #define SPDLOG_CONSTEXPR_FUNC inline
    #endif
#endif
```

This may be friendlier to when we're using `std::format` as opposed to `fmt` as well.

gabime:
> An alternative approach that might work here is to reuse FMT_CONSTEXPR with something like

Agreed. Seems cleaner and to the point (use same constexpr as fmt).

include/spdlog/common.h
```diff
@@ -65,14 +65,23 @@
...
+// depending on the compiler
+// If fmt determines it can't use constexpr, we should inline the function instead
+#ifdef SPDLOG_USE_STD_FORMAT
+    #define SPDLOG_CONSTEXPR_FUNC constexpr
+#else  // Being built with fmt
+    #if FMT_USE_CONSTEXPR
+        #define SPDLOG_CONSTEXPR_FUNC FMT_CONSTEXPR
     #else
-        #define SPDLOG_CONSTEXPR_FUNC inline
+	#define SPDLOG_CONSTEXPR_FUNC inline
```
kkraus14, line 84: I needed to change this slightly to maintain the current behavior. If fmt determines it can't use constexpr then it defines `FMT_CONSTEXPR` as empty https://github.com/gabime/spdlog/blob/cedfeeb95f3af11df7d3b1e7e0d3b86b334dc23b/include/spdlog/fmt/bundled/core.h#L92-L107 while spdlog's current behavior is to define it as `inline`

kkraus14: We could probably use some preprocessor magic to detect when `FMT_CONSTEXPR` is empty and use `inline` instead in those situations, but that adds some non-trivial complexity.

gabime: Why would it be empty?

kkraus14: It gets defined as empty for some reason: https://github.com/gabime/spdlog/blob/cedfeeb95f3af11df7d3b1e7e0d3b86b334dc23b/include/spdlog/fmt/bundled/core.h#L106

gabime: I think it won't matter for spdlog. It can be empty for spdlog as well.

kkraus14: Updated based on that

kkraus14:
Looks like without the inline in C++11 (in C++14 and newer it would be constexpr) we hit some ODR violations from the header being included in multiple translation units that then get linked into a single artifact.

gabime:
So  I guess we need to use your first suggestion after all.

kkraus14:
Reverted the change where I think this should be good to go. Thanks for your review and feedback @gabime!

gabime:
Thanks @kkraus14

PR #2918, opened 2023-10-23
sinks: Make syslog_sink.h's syslog_prio_from_level protected

shannonbooth:
To allow for using this function from a derived sink.


gabime:
Thanks @shannonbooth

PR #2919, opened 2023-10-23
Fix wrong thread_id (TID) in systemd_sink.h

M010:
details::os::thread_id() is not always thread_id of the _msg_. We should use _msg_ field instead


gabime:
Thanks @M010

PR #2923, opened 2023-10-24
fmt/*.h: include tweakme.h to set SPDLOG_FMT_EXTERNAL according to system

### LINKED ISSUE #2922: fmt/*.h needs to #include <spdlog/tweakme.h> to set `SPDLOG_FMT_EXTERNAL` correctly
Experiencing build [failures on](https://github.com/gnuradio/gnuradio/issues/6927) machines where tweakme.h set `SPDLOG_FMT_EXTERNAL` when we 

```C++
#include <spdlog/fmt/fmt.h>
```

in header files which do not otherwise include spdlog before.

The reason is simple: 

spdlog/fmt/fmt.h checks for `SPDLOG_FMT_EXTERNAL` to select whether to include the bundled or the system FMT header.

Without tweakme.h being included there, this goes wrong, like in https://github.com/gnuradio/gnuradio/issues/6927, where we [simply include `spdlog/fmt/fmt.h`](https://github.com/gnuradio/gnuradio/blob/5eee613e70ff12a02c9d8d1302f8ff29bcd95df3/gnuradio-runtime/include/gnuradio/io_signature.h#L16) in order to define a formatter, without including the whole of spdlog in a public header (which would potentially have wonderful API effects)

marcusmueller:
Fixes #2922

Since people need to include the version of FMT that spdlog uses when they do things like defining custom formatters in headers, it should be possible to `#include <spdlog/fmt/fmt.h>` without previously including other spdlog headers.

Currently that's not possible (#2922), and simply including the tweakme.h solves that.

### DIFF
--- a/include/spdlog/fmt/chrono.h
+++ b/include/spdlog/fmt/chrono.h
@@ -7,6 +7,7 @@
 //
 // include bundled or external copy of fmtlib's chrono support
 //
+#include <spdlog/tweakme.h>
 
 #if !defined(SPDLOG_USE_STD_FORMAT)
     #if !defined(SPDLOG_FMT_EXTERNAL)
--- a/include/spdlog/fmt/compile.h
+++ b/include/spdlog/fmt/compile.h
@@ -7,6 +7,7 @@
 //
 // include bundled or external copy of fmtlib's compile-time support
 //
+#include <spdlog/tweakme.h>
 
 #if !defined(SPDLOG_USE_STD_FORMAT)
     #if !defined(SPDLOG_FMT_EXTERNAL)
--- a/include/spdlog/fmt/fmt.h
+++ b/include/spdlog/fmt/fmt.h
@@ -9,6 +9,7 @@
 // Include a bundled header-only copy of fmtlib or an external one.
 // By default spdlog include its own copy.
 //
+#include <spdlog/tweakme.h>
 
 #if defined(SPDLOG_USE_STD_FORMAT)  // SPDLOG_USE_STD_FORMAT is defined - use std::format
     #include <format>
--- a/include/spdlog/fmt/ostr.h
+++ b/include/spdlog/fmt/ostr.h
@@ -7,6 +7,7 @@
 //
 // include bundled or external copy of fmtlib's ostream support
 //
+#include <spdlog/tweakme.h>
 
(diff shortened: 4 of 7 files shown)


gabime:
Thanks @marcusmueller

marcusmueller:
@gabime no problem! Thanks for maintaining the library!

Question on working around this for now in our project, arising from [my workaround](https://github.com/gnuradio/gnuradio/pull/6928)

Would you recommend I include 
1.  `tweakme.h`
2. `common.h`
3. do something completely different

on systems with current (i.e., pre-this-patch) spdlog, where I have public headers that need to include `spdlog/fmt/fmt.h`?

gabime:
Probably including tweakme.h. That being said, not sure why the gnuradio code includes spdlog/fmt/fmt.h everywhere instead of just including directly the external fmt.h.

Also, defining SPDLOG_FMT_EXTERNAL in the build might be more robust than messing with tweakme.h

marcusmueller:
> not sure why the gnuradio code includes spdlog/fmt/fmt.h everywhere instead of just including directly the external fmt.h.

Because GNU Radio (nor any other software that has a public interface through which people can access spdlog or fmt interfaces) cannot know whether on the user's target platform, the system-provided spdlog uses external or internal fmt. By the way, because fmt and spdlog aren't compatible in many versions  that actual distros actually ship, Fedora, 
uses internal fmt, ubuntu 22.04 external, Ubuntu 20.04 internal…)

So, every library that depends on spdlog  and wants to use fmt needs to make sure that it's using the same fmt as spdlog does.

And bundling spdlog isn't an option either, because you want to enable the consumers of your library to use the same spdlog as they would – and that's the system-supplied one. (Aside from the fact that bundling such a library would go against distro guidelines – thankfully.)

>Also, defining SPDLOG_FMT_EXTERNAL in the build might be more robust than messing with tweakme.h

Sadly, nope. You must not do that on a system where the system's tweakme does not set SPDLOG_FMT_EXTERNAL (trust me, we tried).

(This situation would be marginally better if the CMake config file for spdlog would export that define; alas, it doesn't, so you need to cannot know, without compiling a program with the same settings as you'd use for your actual binaries, whether you'll have an external- or internal-fmt spdlog :( )

Issue #2113, opened 2021-09-22
logger object is not ensured to outlive all static local variables that it uses


wangjunZeroZeroSeven:
In this library, it is preferred that local static variables defines in a function scope, in this way this static variable will construct the first time we call this function .

Since static variables will destruct in the inverted order of construction, if the function is called after program starts, the static varaible inside the function will construct when program is running and it will destruct earlier than any normal destructors which execute at the end of program. 

As a result, we can not safely use logger object in normal destructors since it may call some static variables that already destruct !  I met this program in my project , could you `force` all static local variables outlive `logger` object ?

tt4g:
The lifetime of the spdlog logger is managed by `std::shared_ptr`.
If you want to access it safely in a destructor, just copy `std::shared_ptr<spdlog::logger>` into a member variable.

wangjunZeroZeroSeven:
`std::shared_ptr` only controll the lifetime of spdlog logger, but it can't controll these basic things:

the `s_mutex` in :
```
struct console_mutex
{
    using mutex_t = std::mutex;
    static mutex_t &mutex()
    {
        static mutex_t s_mutex;
        return s_mutex;
    }
};
```

the `s_mutex` in :
```
struct console_nullmutex
{
    using mutex_t = null_mutex;
    static mutex_t &mutex()
    {
        static mutex_t s_mutex;
        return s_mutex;
    }
};
```

the `mutex` in :
```
SPDLOG_INLINE void logger::err_handler_(const std::string &msg)
{
    if (custom_err_handler_)
    {
        custom_err_handler_(msg);
    }
    else
    {
        using std::chrono::system_clock;
        static std::mutex mutex;
        static std::chrono::system_clock::time_point last_report_time;
        static size_t err_counter = 0;
        std::lock_guard<std::mutex> lk{mutex};
        auto now = system_clock::now();
        err_counter++;
        if (now - last_report_time < std::chrono::seconds(1))
        {
            return;
        }
        last_report_time = now;
        auto tm_time = details::os::localtime(system_clock::to_time_t(now));
        char date_buf[64];
        std::strftime(date_buf, sizeof(date_buf), "%Y-%m-%d %H:%M:%S", &tm_time);
#if defined(USING_R) && defined(R_R_H) // if in R environment
        REprintf("[*** LOG ERROR #%04zu ***] [%s] [%s] {%s}\n", err_counter, date_buf, name().c_str(), msg.c_str());
#else
[...]

PR #2929, opened 2023-11-03
Memory leak fix on shutdown

plekakis:
Fix for false-positive memory leak when using leak detection methods (such as CRT dump memory leaks).

Since we cannot rely on static de-initialisation for freeing the registry::s_instance, this is now lazily allocated instead and freed during spdlog::shutdown() function. In my test case, saves ~15 reported memory leaks.

### DIFF
--- a/include/spdlog/details/registry-inl.h
+++ b/include/spdlog/details/registry-inl.h
@@ -30,7 +30,7 @@
 namespace spdlog {
 namespace details {
 
-registry* registry::s_instance = nullptr;
+std::atomic<registry *> registry::s_instance{nullptr};
 std::mutex registry::s_instanceMutex;
 
 SPDLOG_INLINE registry::registry()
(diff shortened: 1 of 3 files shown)


include/spdlog/details/registry-inl.h
```diff
@@ -237,9 +240,19 @@ SPDLOG_INLINE void registry::set_levels(log_levels levels, level::level_enum *gl
...
-    static registry s_instance;
-    return s_instance;
+SPDLOG_INLINE void registry::free() {
+    std::lock_guard<std::mutex> lock(s_instanceMutex);
+
+    delete s_instance;
+    s_instance = nullptr;
+}
+
+SPDLOG_INLINE registry &registry::instance() {
```
gabime, line 250: Should be double checked locking. This is expensive

gabime:
I don’t know. Seems expensive to lock the singleton on each access. Double check locking with barriers could be used, but I never understood it completely. see https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/

plekakis:
I have updated the function to do double locking. In practice it's simple, you rely on atomics to effectively early out the expensive mutex lock path, if the instance has already been allocated.
The expensive lock path will now only occur once in the lifetime of the logger and not always.

gabime:
Shouldn’t there be `std::atomic_thread_fence(std::memory_order_acquire);` and `std::atomic_thread_fence(std::memory_order_release);` somewhere?

gabime:
I think this would break the current behavior. Currently, calling shutdown is not mandatory since the registry object would be automatically destroyed. With this pr, the registry will be never destroyed if not calling shutdown.
Please post to the v2.x branch.

plekakis:
Ah okay, I thought that shutdown was supposed to be called.
Let's close this PR, I will have a look at the v2 branch for now.

gabime:
@plekakis  See #2113 for a discussion and a fix [proposal](https://github.com/gabime/spdlog/issues/2113#issuecomment-1743261882)

PR #2935, opened 2023-11-09
`async_msg`: take ownership of filename/funcname

### LINKED ISSUE #2867: [source_loc] support for owning strings
`source_loc` does not own the strings for file name and function name. this is a perfectly reasonable default in most use cases where this information is filled by the preprocessor.

however i'm running into a funny situation when trying to re-direct qt's logging to spdlog and using multithreaded logging:
* the qml engine uses heap-allocated strings for file/function name and fill `QMessageLogContext` with raw `const char*` pointers (compare https://github.com/qt/qtdeclarative/blob/abb9cde13d6d0f503a8a6aad12f4361cd3848c79/src/qml/qml/qqmlengine.cpp#L1580-L1605)
* i directly map `QMessageLogContext` onto `source_loc`
* in multi-thread spdlog we run into use-after-free issues (since the worker thread accesses the strings that are freed by the other thread)

---

i've been meditating a bit about this. it's a bit of an edge case, but it would be great if spdlog could provide a second api similar to `source_loc`, that uses owning `std::string` to allow applications to fill a `source_loc` from source locations other than c/c++

timblechmann:
the async logger cannot ensure that filename/funcname pointers are still valid at the time when the async worker is executed:
* filename/funcname may be provided by a VM
* filename/funcname may point to a readonly text field, in a shared object that had been `dlclose`d

we add an api that can be enabled via `SPDLOG_ASYNC_OWNING_SOURCELOC_STRINGS` to take ownership of the strings

fixes #2867

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,10 @@ option(SPDLOG_SANITIZE_ADDRESS "Enable address sanitizer in tests" OFF)
 # warning options
 option(SPDLOG_BUILD_WARNINGS "Enable compiler warnings" OFF)
 
+# owning sourceloc strings options
+option(SPDLOG_ASYNC_OWNING_SOURCELOC_STRINGS "Async loggers take owership of source_loc strings" OFF)
+
+
 # install options
 option(SPDLOG_SYSTEM_INCLUDES "Include as system headers (skip for clang-tidy)." OFF)
 option(SPDLOG_INSTALL "Generate the install target" ${SPDLOG_MASTER_PROJECT})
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -113,6 +113,8 @@ else()
     set(SPDLOG_WCHAR_FILENAMES OFF CACHE BOOL "non supported option" FORCE)
 endif()
 
+set(SPDLOG_ASYNC_OWNING_SOURCELOC_STRINGS BOOL "Async loggers take owership of source_loc strings" OFF)
```
tt4g: You should use `option()` instead of `set()`.

Also, it seems halfway to define a new option in this position.
Why not define it where other options are defined together (e.g., after `option(SPDLOG_DISABLE_DEFAULT_LOGGER ...)`)?

PR #2963, opened 2023-12-22
Update example.cpp to fix the vector issue in bin_example

zhuzhzh:
with std::vector<char> buf(80),  80 elements are put in the vector, which is not the expected behavior.


gabime:
Thanks @zhuzhzh

PR #2973, opened 2024-01-08
Better integrate with std::source_location

### LINKED ISSUE #1959: Example code how to use your library with the c++20 source_location
I created an example code on how to use your library with the c++20 `source_location` instead of your macros:

```cpp
#include <spdlog/spdlog.h>

#include <experimental/source_location>
#include <iostream>
#include <string_view>

namespace logging {
using source_location = std::experimental::source_location;
[[nodiscard]] constexpr auto get_log_source_location(
    const source_location &location) {
  return spdlog::source_loc{location.file_name(),
                            static_cast<std::int32_t>(location.line()),
                            location.function_name()};
}

struct format_with_location {
  std::string_view value;
  spdlog::source_loc loc;

  template <typename String>
  format_with_location(
      const String &s,
      const source_location &location = source_location::current())
      : value{s}, loc{get_log_source_location(location)} {}
};

template <typename... Args>
void warn(format_with_location fmt, Args &&...args) {
  spdlog::default_logger_raw()->log(fmt.loc, spdlog::level::warn, fmt.value,
                                    std::forward<Args>(args)...);
}
}  // namespace logging

int main() {
  spdlog::set_level(spdlog::level::trace);
  std::string message = "hello {}\n";
  auto value = 42;
  logging::warn(message, value);
}
```

https://godbolt.org/z/aqdYjo6nE

Because it took some time to fiddle with the type deduction I wanted to share this. Maybe you could add it to the documentation?

Thanks for your great library!

stephanlachnit:
This PR aims to better integrate spdlog v1.x with [`std::source_location`](https://en.cppreference.com/w/cpp/utility/source_location). Effectively closes https://github.com/gabime/spdlog/issues/1959.

The main idea is to allow constructing `spdlog::source_loc` from `std::source_location`. This approach is significantly simpler compared to https://github.com/gabime/spdlog/pull/2667 and still works as one would expect:
```cpp
spdlog::log(std::source_location::current(), spdlog::level::info, "test");
```

The neat part is that thanks to `std::source_location` being `constexpr`, it has the same runtime performance as the macro solution. The only caveat is that `std::source_location` uses `std::uint_least32_t` instead of `int` for the line number - I adapted `spdlog::source_loc` to use this as well, which might result in some compiler/linter warnings when upgrading (though they are trivial to fix by changing the cast type).

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -9,6 +9,7 @@
 #include <atomic>
 #include <chrono>
 #include <cstdio>
+#include <cstdint>
 #include <exception>
 #include <functional>
 #include <initializer_list>
@@ -319,14 +320,14 @@ private:
 
 struct source_loc {
     SPDLOG_CONSTEXPR source_loc() = default;
-    SPDLOG_CONSTEXPR source_loc(const char *filename_in, int line_in, const char *funcname_in)
+    SPDLOG_CONSTEXPR source_loc(const char *filename_in, std::uint_least32_t line_in, const char *funcname_in)
         : filename{filename_in},
           line{line_in},
           funcname{funcname_in} {}
 
     SPDLOG_CONSTEXPR bool empty() const SPDLOG_NOEXCEPT { return line == 0; }
     const char *filename{nullptr};
-    int line{0};
+    std::uint_least32_t line{0};
     const char *funcname{nullptr};
 };
 
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -16,8 +16,13 @@
 #include <string>
 #include <type_traits>
 
+#ifdef __has_include
+    #if __has_include(<version>)
+        #include <version>
+    #endif
+#endif
+
 #ifdef SPDLOG_USE_STD_FORMAT
-    #include <version>
     #if __cpp_lib_format >= 202207L
         #include <format>
     #else
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -107,7 +107,13 @@
 #endif
 
 #ifndef SPDLOG_FUNCTION
-    #define SPDLOG_FUNCTION static_cast<const char *>(__FUNCTION__)
+    // use std::source_location instead of macros if available
+    #ifdef __cpp_lib_source_location
+        #define SPDLOG_STD_SOURCE_LOCATION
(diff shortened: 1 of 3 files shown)


gabime:
Thanks @stephanlachnit . All new features  go to the v2.x branch, and since it already contains source_location integration I will have to pass.

stephanlachnit:
> Thanks @stephanlachnit . All new features go to the v2.x branch, and since it already contains source_location integration I will have to pass.

Can you maybe reconsider this? This is not really a new "feature", just a single new constructor for convenience for the exact same feature that already exists. If the `int` -> `std::uint_least32_t` change bothers you, that can be replaced with a static cast. I would love to use the 2.x branch, but without a release that is unfortunately not possible.

stephanlachnit:
@gabime I now removed the `int` -> `std::uint_least32_t` change (I think you need to reopen the PR to see the change or check [here](https://github.com/gabime/spdlog/compare/v1.x...stephanlachnit:spdlog:p-std-source-location)).

The PR now only adds a single symbol if `std::source_location` is available and is otherwise fully transparent. The change is only 34 lines of changed code, if you count the new code it is only 22 pretty trivial lines, which I think is rather few to close an issue ticket.

gabime:
Not sure I understand why a new implementation of the SPDLOG_LOGGER_CALL macro benefits  library users.  
The main point of using source_location would be to log source locations without using the macros.

stephanlachnit:
> Not sure I understand why a new implementation of the SPDLOG_LOGGER_CALL macro benefits library users. The main point of using source_location would be to log source locations without using the macros.

The main point of having `std::source_location` would be indeed logging without macros. For example, we defined custom level names and thus have a custom enum that we cast to `spdlog::level::level_enum`, basically:


```cpp
void custom_log(Level level, std::string_view message, const std::source_lcation& loc = std::source_location::current()) {
    spdlog_logger_->log(loc, static_cast<spdlog::level::level_enum>(level), message);
}
```
We can of course also create `std::source_location` there (which we currently do), but I think since this makes more sense being integrated in the library, especially since this has been requested before in https://github.com/gabime/spdlog/issues/1959.

If you want I can drop the use of `std::source_location` in `SPDLOG_LOGGER_CALL`, it's just something you get for free.

PR #2974, opened 2024-01-09
Fix MSVC compile flag for no exceptions

Programicus:
Looks like there was a small oversite when the disable exceptions flag was implemented for msvc. As per [msdn](https://learn.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-170), the correct flag for disabling exceptions has the hyphens.

This issue also exists on the 2.x branch, however I do not know enough about github if its possible to make this pull request apply to both branches


gabime:
Thanks @Programicus

PR #2978, opened 2024-01-13
Also use _stat() on Windows to be more UTF8 friendly

### LINKED ISSUE #2977: details::os::path_exists(const std::string&) doesn't handle UTF8 filename correctly under Windows
My application must support both Windows and Linux, so every path and filename inside the app is encoded UTF8 with std::string, and WCHAR is not used (SPDLOG_WCHAR_FILENAMES is not enabled).

 But under Windows (code taken from the latest spdlog 1.13.0):
```
SPDLOG_INLINE bool path_exists(const filename_t &filename) SPDLOG_NOEXCEPT {
#ifdef _WIN32
    #ifdef SPDLOG_WCHAR_FILENAMES
    auto attribs = ::GetFileAttributesW(filename.c_str());
    #else
    auto attribs = ::GetFileAttributesA(filename.c_str());
    #endif
    return attribs != INVALID_FILE_ATTRIBUTES;
#else  // common linux/unix all have the stat system call
```
```GetFileAttributesA()``` will use CP_ACP (the default system code page) not CP_UTF8, which will fail if the path contains UTF8-encoded non-ascii characters.
My current work around for this issue is:
```
SPDLOG_INLINE bool path_exists(const filename_t &filename) SPDLOG_NOEXCEPT {
    std::error_code ec;
    return std::filesystem::exists(filename, ec);
}
```
But I know std::filesystem requires C++17, so it doesn't count as a general solution for spdlog.

Note for the above code to work under Windows, we need to call
```setlocale(LC_ALL, ".UTF8")```
at the beginning of the app (https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170#utf-8-support), so that ```std::filesystem::path``` will treat the pathname as UTF8 encoded string.

I just open this issue for discussion, still looking for a more general solution. Thanks.

liubing:
This PR tries to fix #2977.

The advantage of using ```_stat()``` function over ```GetFileAttributesA()``` API is that:

- When ```SPDLOG_WCHAR_FILENAMES``` is not defined:
-- If ```setlocale(LC_ALL, ".UTF8")``` is called before, then filename will be treated as UTF8 string, and this is the main use case for supporting UTF8 filenames with std::string under Windows.
-- If ```setlocale(LC_ALL, ".UTF8")``` is not called before, then filename will be treated with the current code page (CP_ACP), just like the old behavior.
- When ```SPDLOG_WCHAR_FILENAMES``` is defined, the behavior is the same as if ```GetFileAttributesW()``` was used.

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -177,11 +177,11 @@ SPDLOG_INLINE int rename(const filename_t &filename1, const filename_t &filename
 // Return true if path exists (file or directory)
 SPDLOG_INLINE bool path_exists(const filename_t &filename) SPDLOG_NOEXCEPT {
 #ifdef _WIN32
-    struct _stat64i32 buffer;
+    struct _stat buffer;
     #ifdef SPDLOG_WCHAR_FILENAMES
-    return (::_wstat64i32(filename.c_str(), &buffer) == 0);
+    return (::_wstat(filename.c_str(), &buffer) == 0);
     #else
-    return (::_stat64i32(filename.c_str(), &buffer) == 0);
+    return (::_stat(filename.c_str(), &buffer) == 0);
     #endif
 #else  // common linux/unix all have the stat system call
     struct stat buffer;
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @liubing

PR #2979, opened 2024-01-13
Add constructor to `spdlog::source_loc` using `std::source_location`

### LINKED ISSUE #1959: Example code how to use your library with the c++20 source_location

stephanlachnit:
Alternative to #2973, closes #1959.

Compared to #1959, this now only adds a constructor to `spdlog::source_loc` using `std::source_location` (if available), but does not change the type of `line` in the struct or use `std::source_location::current()` in the `SPDLOG_LOGGER_CALL` macro.

Motivation for this change is given in https://github.com/gabime/spdlog/pull/2973#issuecomment-1887390376.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -105,6 +105,12 @@
     #endif
 #endif
 
+// Add constructor to spdlog::source_loc using std::source_location
+#if __cpp_lib_source_location >= 201907L
+    #define SPDLOG_STD_SOURCE_LOCATION
+    #include <source_location>
+#endif
+
 #ifndef SPDLOG_FUNCTION
     #define SPDLOG_FUNCTION static_cast<const char *>(__FUNCTION__)
 #endif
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. However, Like  I already mentioned, I prefer not to deal with std::source_location in branch 1.x. 
As a side note, I think passing const ref might be less performant than passing by value as this is small constpexr struct.

Issue #2975, opened 2024-01-10
The macro SPDLOG_ACTIVE_LEVEL is defined as SPDLOG_LEVEL_TRACE, but SPDLOG_TRACE is not effective.


Ccccx159:
## Summery

When I use the code in the Basic Usage section, macro `SPDLOG_TRACE` is not effective. Even if I define `SPDLOG_ACTIVE_LEVEL` as `SPDLOG_LEVEL_TRACE`.

## Reproduce

source code:

```cpp
#include "spdlog/spdlog.h"

int main() 
{
    spdlog::info("Welcome to spdlog!");
    spdlog::error("Some error message with arg: {}", 1);
    
    spdlog::warn("Easy padding in numbers like {:08d}", 12);
    spdlog::critical("Support for int: {0:d};  hex: {0:x};  oct: {0:o}; bin: {0:b}", 42);
    spdlog::info("Support for floats {:03.2f}", 1.23456);
    spdlog::info("Positional args are {1} {0}..", "too", "supported");
    spdlog::info("{:<30}", "left aligned");
    
    spdlog::set_level(spdlog::level::debug); // Set global log level to debug
    spdlog::debug("This message should be displayed..");    
    
    // change log pattern
    spdlog::set_pattern("[%H:%M:%S %z] [%n] [%^---%L---%$] [thread %t] %v");
    
    // Compile time log levels
    // define SPDLOG_ACTIVE_LEVEL to desired level
    SPDLOG_TRACE("Some trace message with param {}", 42);
    SPDLOG_DEBUG("Some debug message");
}
```

compiled command:

 `g++ -I include -std=c++11 -O0 -g -DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_TRACE -o spdlog-main ./spdlog-main.cc`

Here is the result of compiling and running:

```bash
code-server [13:57:01] log-example  ➜ g++ -I include -std=c++11 -O0 -g -DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_TRACE -o spdlog-main ./spdlog-main.cc 
code-server [13:57:10] log-example  ➜ ./spdlog-main 
[2024-01-10 13:57:14.143] [info] Welcome to spdlog!
[2024-01-10 13:57:14.143] [error] Some error message with arg: 1
[2024-01-10 13:57:14.143] [warning] Easy padding in numbers like 00000012
[...]

PR #2981, opened 2024-01-15
Adds details about how compile time macros work

egaban:
Makes it more clear about how compile time macros work. Currently, it may induce the user to think that the macros will override the current log level, and they won't. This is related to issue #2975.


gabime:
Thanks @egaban

Issue #2985, opened 2024-01-22
Mark logging functions as noexcept


mtnpke:
Hi!

I'm having a small issue with spdlog in that the logging functions are not marked as `noexcept` even though they should not throw - according to the documentation [here](https://github.com/gabime/spdlog/wiki/Error-handling):

> Spdlog will not throw exceptions while logging

This leads to the static analysis tool SonarQube reporting potential issues when logging functions are called in destructors:

![image](https://github.com/gabime/spdlog/assets/72438900/fb076b1e-c24c-4b66-b98b-507424209863)

Since the rule of not throwing exceptions in destructors is generally valid and valuable, I'd like to keep it enabled. It does, however, detect correctly that the logging functions are

* not marked `noexcept` and
* can potentially throw (see https://github.com/gabime/spdlog/blob/696db97f672e9082e50e50af315d0f4234c82397/include/spdlog/logger.h#L42 which is done due to #533 )

Ideally, I think the functions could just be marked `noexcept`, but I'm not sure how that would interact with `pthread_cancel`. A new define might be needed to be able to disable the code path that is specific to `pthread_cancel` (it is not the most common of use cases I guess?) and/or to introduce the `noexcept`. I'm willing to do the code changes and PR if we can reach an agreement on how to proceed :)

tt4g:
In my opinion, that is extremely difficult.

First, there is no `noexcept` specified for user-defined error handlers.
It is possible for an exception to be `throw` in the error handler after spdlog notifies the error handler.

And since almost all functions, constructors (copy-constructor, move-constructor), and destructors in STL have no `noexcept`, they can be `throw` by any function.
In particular, it is almost impossible to guarantee that the argument copy-constructor and move-constructor will not throw an exception.

tt4g:
It is well known that `throw` in the destructor can cause crashes: [c\+\+ \- If you shouldn't throw exceptions in a destructor, how do you handle errors in it? \- Stack Overflow](https://stackoverflow.com/questions/130117/if-you-shouldnt-throw-exceptions-in-a-destructor-how-do-you-handle-errors-in-i)

However, it can be prevented by simply using `try`-`catch` in the destructor to prevent the exception from propagating outside the destructor.

mtnpke:
> First, there is no noexcept specified for user-defined error handlers.
> It is possible for an exception to be throw in the error handler after spdlog notifies the error handler.

[...]

PR #2986, opened 2024-01-22
Add SPDLOG_LOGGER_NOEXCEPT option to declare logging noexcept

mtnpke:
Use to declare all output functionality in spdlog namespace and on spdlog::logger as
noexcept. This may allow the compiler to perform further optimizations when logging is performed.
Additionally, static analysis can assume that logging never throws, which may be useful
when logging is used in inherently exception-unsafe places such as destructors.

Note that this will result in program termination if the registered error handler
throws. It might also interact badly with mechanisms that inject exceptions into running
code (such as the pthread_cancel implementation of NPTL).

Implements #2985.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -81,7 +81,7 @@
     #if FMT_USE_CONSTEXPR
         #define SPDLOG_CONSTEXPR_FUNC FMT_CONSTEXPR
     #else
-	#define SPDLOG_CONSTEXPR_FUNC inline
+        #define SPDLOG_CONSTEXPR_FUNC inline
     #endif
 #endif
 
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -70,6 +70,12 @@
     #define SPDLOG_CONSTEXPR constexpr
 #endif
 
+#ifdef SPDLOG_LOGGER_NOEXCEPT
+    #define SPDLOG_COND_NOEXCEPT SPDLOG_NOEXCEPT
+#else
+    #define SPDLOG_COND_NOEXCEPT
+#endif
+
 // If building with std::format, can just use constexpr, otherwise if building with fmt
 // SPDLOG_CONSTEXPR_FUNC needs to be set the same as FMT_CONSTEXPR to avoid situations where
 // a constexpr function in spdlog could end up calling a non-constexpr function in fmt
--- a/include/spdlog/logger.h
+++ b/include/spdlog/logger.h
@@ -76,17 +76,20 @@ public:
     void swap(spdlog::logger &other) SPDLOG_NOEXCEPT;
 
     template <typename... Args>
-    void log(source_loc loc, level::level_enum lvl, format_string_t<Args...> fmt, Args &&...args) {
+    void log(source_loc loc, level::level_enum lvl, format_string_t<Args...> fmt, Args &&...args)
+        SPDLOG_COND_NOEXCEPT {
         log_(loc, lvl, details::to_string_view(fmt), std::forward<Args>(args)...);
     }
 
     template <typename... Args>
-    void log(level::level_enum lvl, format_string_t<Args...> fmt, Args &&...args) {
+    void log(level::level_enum lvl,
+             format_string_t<Args...> fmt,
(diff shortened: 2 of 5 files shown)


gabime:
Thanks. I am reluctant to add so many macros to the  API. What  is the performance gain you see with this?

mtnpke:
To be upfront: Performance was not a motivation at all for me; my motivation is to aid static analysis as described in #2985. That said, I did not put a lot of time into benchmarking. I only ran the spdlog benchmark executable once in a release build with/without the macro to see that the performance is not getting worse. I did not observe any statistically significant difference. The optimization opportunities for the compiler, as far as I can tell, would materialize not in spdlog, but in conjunction with certain situations in the *calling code* (due to added reordering opportunities and so on), so it cannot be measured with a simple benchmark and is inherently difficult to quantize. I can remove the paragraph about the performance if it helps make the motivation clearer.

What problem do you perceive with the macros?

mtnpke:
On another note, there is zero change to the library compared to the current code when the define is not set, which is the default :)

gabime:
Closed. This branch should change the api as minimal as possible. Adding macros (even if disabled by default) change the api

mtnpke:
* How do macros that are disabled by default change the API?
* Adding `noexcept` *is* about the API. The only way to add it without changing the API is to add new methods or classes. Would that be the way forward?

gabime:
For starters, It adds to the cognitive load of the API users. Now users would have to take into account that the value of the macro might be different then the default. Even reading the API would be harder.
Also if this macro used, it  would break existing code with custom throwing error handlers (and possibly the abi I think).

And all of this for no real value, except for making some linter happy when examining destructors. There is even no performance gain as you mentioned. 

We can consider adding a noexcept to the v2.x api though (not a macro).  This would of course also break existing throwing  error handlers, but since it is new major version, I think it is more acceptable.

mtnpke:
> For starters, It adds to the cognitive load of the API users. Now users would have to take into account that the value of the macro might be different then the default. Even reading the API would be harder.

That's a valid point, thanks for explaining. For me, this level of macros on signatures is not unusual to have for a major C++ library, but I can see how there is a tradeoff to be made here.

> Also if this macro used, it would break existing code with custom throwing error handlers (and possibly the abi I think).

Indeed, but that's why it is a macro, it's off by default, and this fact is specifically mentioned in the macro's documentation. ABI is not defined by the standard so YMMV, but the [internet says at least gcc and clang don't care](https://stackoverflow.com/questions/51429454/does-adding-noexcept-break-binary-compatibility).

> And all of this for no real value, except for making some linter happy when examining destructors. 

I do consider this project being easier to use in conjunction with linters/static analyzers (which are really helpful in general!) real value. Else I wouldn't put time into this discussion and PR :)

> There is even no performance gain as you mentioned.

Yep I don't consider performance a major motivation here, but to be clear we can't exactly declare that there will be no performance gain at all. It mostly depends on the call site. Impossible to say without benchmarking a real application that does lots of logging.

> We can consider adding a noexcept to the v2.x api though (not a macro). This would of course also break existing throwing error handlers, but since it is new major version, I think it is more acceptable.

That would be nice! Two thoughts on this:

* Is throwing exceptions in error handlers a relevant use case? Because that will be completely impossible without a macro switch. I personally think always having the `noexcept` is fine, but just something to think about.
* How does `noexcept` interact with NPTL `pthread_cancel`? I searched around for a bit, but was unable to find a definite answer. If it does not interact well, always having the `noexcept` might preclude using spdlog in conjunction with `pthread_cancel`.

gabime:
> Is throwing exceptions in error handlers a relevant use case? Because that will be completely impossible without a macro switch. I personally think always having the noexcept is fine, but just something to think about.

The original design was to support throwing error handlers. For example for use cases where failing to log is critical error that need to handled immediately by the calling code. On the other hand most users don’t need this, and I never saw this in actual use, so I am not sure :(

> How does noexcept interact with NPTL pthread_cancel? I searched around for a bit, but was unable to find a definite answer. If it does not interact well, always having the noexcept might preclude using spdlog in conjunction with pthread_cancel.

Please provide more details. Why would  NPTL pthread_cancel be problematic?

PR #2993, opened 2024-01-29
Update daily_file_sink.h

AaronKonishi:
solve compiler problem: not supported std::put_time() under version 5


AaronKonishi:
when I use 'arm-cortexa9-linux-gnueabihf-g++',  I get an error: ‘put_time’ is not a member of ‘std’

gabime:
CI test fail. And anyway spdlog doesnt support such old compiler

PR #3000, opened 2024-02-05
Added support of Swift Package Manager

IhorShevchuk:
This allows to add spdlog as dependency to other Swift Packages or Xcode projects


gabime:
Thanks. spdlog only supports cmake build. It cant support the many other build systems out there, so i need to pass on this one.

PR #3001, opened 2024-02-05
Redundant message when using multi-config generators

WarEagle451:
Multi-config generators such as "Ninja Multi-Config" will not define CMAKE_BUILD_TYPE, therefore the "Build type" message should not be called.


gabime:
Thanks. However I kinda of like it that it prints even an empty string to make it clear

PR #3009, opened 2024-02-12
Expose the flusher thread object to user in order to allow setting of thread name and thread affinity when needed

cohdan:
In our app there is a need to set the thread name and thread affinity for all spawned thread. In order to do so with the spdlog flusher thread, I had to expose the thread from the internal data structure  (periodic_worker) via the registery. I thought this might be worthy of a merge back into the main stream.

### DIFF
--- a/include/spdlog/details/periodic_worker.h
+++ b/include/spdlog/details/periodic_worker.h
@@ -38,6 +38,7 @@ public:
             }
         });
     }
+    std::thread *get_thread() { return &worker_thread_; }
     periodic_worker(const periodic_worker &) = delete;
     periodic_worker &operator=(const periodic_worker &) = delete;
     // stop the worker thread and join it
--- a/include/spdlog/details/periodic_worker.h
+++ b/include/spdlog/details/periodic_worker.h
@@ -38,7 +38,7 @@ public:
             }
         });
     }
-    std::thread *get_thread() { return &worker_thread_; }
+    std::thread &get_thread() { return worker_thread_; }
     periodic_worker(const periodic_worker &) = delete;
     periodic_worker &operator=(const periodic_worker &) = delete;
     // stop the worker thread and join it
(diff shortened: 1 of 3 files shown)


include/spdlog/details/periodic_worker.h
```diff
@@ -38,6 +38,7 @@ class SPDLOG_API periodic_worker {
             }
         });
     }
+    std::thread *get_thread() { return &worker_thread_; }
```
gabime: why not return reference like in the rest of this pr?

cohdan: Fair question :) 
To be honest I was just lazy to change some legacy APIs in my code. But that's not a good enough reason and indeed it makes no sense. I'll update the PR.

include/spdlog/spdlog-inl.h
```diff
@@ -89,4 +89,9 @@ SPDLOG_INLINE void apply_logger_env_levels(std::shared_ptr<logger> logger) {
     details::registry::instance().apply_logger_env_levels(std::move(logger));
 }
 
+SPDLOG_INLINE std::unique_ptr<spdlog::details::periodic_worker> &get_flusher()
+{
+    return details::registry::instance().get_flusher();
```
gabime: This change is not needed, as the registry can be accessed

cohdan: It is not a must, I agree, but it allows to access directly from spdlog namespace (similar to spdlog::register_logger for example) which makes it neater IMHO.

gabime: Better not add new api if not really necessary.

cohdan: ok
uploaded a new PR without this.

gabime: There is a missing mutex lock in the `get_flusher()`

Add 
```c++
std::lock_guard<std::mutex> lock(flusher_mutex_);
```

cohdan: Got it. Fixed.

PR #3010, opened 2024-02-12
Add wide character formatting and output support to wincolor_sink.

DominikGrabiec:
Fixes printing of unicode characters to the windows console such as microsecond suffix for std::chrono types.

I replicated the code to convert from utf8 to wicde characters found in the `msvc_sink` and added it to the `wincolor_sink`. This means it's controlled by the `SPDLOG_WCHAR_TO_UTF8_SUPPORT` define, which at this stage seems to be the general "convert from utf8 to wide characters before printing" flag.

I tested this by adding this change as a patch to vcpkg locally on my system.


DominikGrabiec:
Apologies but I might have jumped the gun a little bit, I need to test the `WriteFile` path a bit more. If you want you can close this or put it on hold, I don't know much about pull requests.

gabime:
No problem. Will close for now.

PR #3023, opened 2024-02-24
Feature | Adds `std::string_view` overloads for logger accessor

BenLeadbetter:
# Adds `std::string_view` overloads for logger accessor

## Motivation 

For clients who store logger ids as string literals, or other non-`std::string` data types, it would be very useful to add a `std::string_view` overload for the `spdlog::get` function. This would save such clients needing to make a temporary, dynamically allocated copy of their logger name data in order to access a specific logger.

### Only enabled in C++17

`std::string_view` is a C++17 feature, so the additional overloads are only included if the project is compiled against C++17 or greater.

### Additional `const char*` overload

With the addition of the `spdlog::get(std::string_view)` overload, calls to this function with a `const char*` become ambiguous.

```cpp
const char *const name = "my_logger";
spdlog::get(name); // error: call to spdlog::get is ambiguous.
```

To disambiguate we also add a `spdlog::get(const char*)` overload which forwards calls on to the `spdlog::get(std::string_view)` overload.

### DIFF
--- a/include/spdlog/details/registry-inl.h
+++ b/include/spdlog/details/registry-inl.h
@@ -84,6 +84,22 @@ SPDLOG_INLINE std::shared_ptr<logger> registry::get(const std::string &logger_na
     return found == loggers_.end() ? nullptr : found->second;
 }
 
+#if __cplusplus >= 201703L  // C++17
+SPDLOG_INLINE std::shared_ptr<logger> registry::get(std::string_view logger_name) {
+    std::lock_guard<std::mutex> lock(logger_map_mutex_);
+    for (const auto &[key, val] : loggers_) {
+        if (key == logger_name) {
+            return val;
+        }
+    }
+    return nullptr;
+}
+
+SPDLOG_INLINE std::shared_ptr<logger> registry::get(const char *logger_name) {
+    return get(std::string_view(logger_name));
+}
+#endif
+
 SPDLOG_INLINE std::shared_ptr<logger> registry::default_logger() {
     std::lock_guard<std::mutex> lock(logger_map_mutex_);
     return default_logger_;
--- a/include/spdlog/details/registry.h
+++ b/include/spdlog/details/registry.h
@@ -18,6 +18,10 @@
 #include <string>
 #include <unordered_map>
 
+#if __cplusplus >= 201703L  // C++17
+    #include <string_view>
+#endif
+
 namespace spdlog {
 class logger;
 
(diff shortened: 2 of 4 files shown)


include/spdlog/details/registry-inl.h
```diff
@@ -84,6 +85,22 @@ SPDLOG_INLINE std::shared_ptr<logger> registry::get(const std::string &logger_na
...
+SPDLOG_INLINE std::shared_ptr<logger> registry::get(std::string_view logger_name) {
+    std::lock_guard<std::mutex> lock(logger_map_mutex_);
+    using key_value_pair = std::pair<std::string, std::shared_ptr<spdlog::logger>>;
+    const auto match = [logger_name](const key_value_pair &item) {
+        return item.first == logger_name;
+    };
+    const auto found = std::find_if(loggers_.cbegin(), loggers_.cend(), match);
+    return found == loggers_.end() ? nullptr : found->second;
+}
+
```
gabime, line 97: since this is cpp17, a much shorter version would be

```

std::lock_guard<std::mutex> lock(logger_map_mutex_);
for (auto const& [key, val] : loggers_) {
  if(logger_name == key) return val;
}
return nullptr;
```

And then also no need to `#include <algorithm>`

gabime (changes requested):

include/spdlog/details/registry-inl.h
```diff
@@ -21,6 +21,7 @@
     #endif
 #endif  // SPDLOG_DISABLE_DEFAULT_LOGGER
 
+#include <algorithm>
```
gabime: Please remove this include

gabime (changes requested):

BenLeadbetter:
@gabime Any idea what's happening with this `gcc 7` check? Looks like a `git` error?

gabime:
Merged. Thanks. The gcc 7 is a git workflow issue indeed

gabime:
@BenLeadbetter I had to revert this since it caused backward compatabilty problems for c++17 that links to spdlog that compiled with c++11. See #3081

BenLeadbetter:
Ah, that's a shame.

Fortunately the project for which the feature was intended went in a different direction in terms of it's spdlog integration.

In case you're interested, we had issues with the mutex locking in the registry on calls to `splog::get`, so we ended up writing a custom non-blocking registry for our loggers.

PR #3026, opened 2024-02-29
Add missing include in circular_q.h

magnus-nomono:
`circular_q` uses the SPDLOG_NOEXCEPT macro, which is defined in `spdlog/common.h`. Due to some recent reorganization of includes in `ringbuffer_sink.h`, `common.h` was no longer included before `circular_q.h`, causing my build to break when updating to v1.13.0


gabime:
Thanks @magnus-nomono

PR #3036, opened 2024-03-11
Add test support in VxWorks 7.

wenshan1:
The API system() isn't supported by VxWorks7.
Use posix APIs implements a new function
vxRemoveDir() to remove a whole directory.


gabime:
No need to recursively delete files and folders for tests. It is enough to delete all files in the direcroy since tests don't create sub folders

PR #3044, opened 2024-03-19
Fix #3038

### LINKED ISSUE #3038: wstr_to_utf8buf throw exception
In "wstr_to_utf8buf" function, memory_buf_t 's size is a const value of 250, (wstr_size+1)*2 is less than 250, so there is no resizing of the result_size. In next WideCharToMultiByte call, this step went wrong.The last error is ERROR_INSUFFICIENT_BUFFER(122)(0x7A).
My test string is"OnPlayerStateChanged:state not changed 〈1〉叶心音的父亲遭人陷害入狱，男友不救反而出国，反而被他弟弟陆景霄救下。可她不知道就是男友陷害的父亲，而自己却已经爱上男友的弟弟！《总裁的金丝雀》,2,1", and can reproduce of this issue. 

wstr_to_utf8buf

Tomas-Zhu:
WideCharToMultiByte is a convert from unicode to utf8, utf8 can use one to four one-byte (8-bit) code units.So (wstr_size+1)*2 is wrong in sometime.

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -439,7 +439,7 @@ SPDLOG_INLINE bool in_terminal(FILE *file) SPDLOG_NOEXCEPT {
 
 #if (defined(SPDLOG_WCHAR_TO_UTF8_SUPPORT) || defined(SPDLOG_WCHAR_FILENAMES)) && defined(_WIN32)
 SPDLOG_INLINE void wstr_to_utf8buf(wstring_view_t wstr, memory_buf_t &target) {
-    if (wstr.size() > static_cast<size_t>((std::numeric_limits<int>::max)()) / 2 - 1) {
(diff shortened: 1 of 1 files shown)


include/spdlog/details/os-inl.h
```diff
@@ -450,7 +450,7 @@ SPDLOG_INLINE void wstr_to_utf8buf(wstring_view_t wstr, memory_buf_t &target) {
     }
 
     int result_size = static_cast<int>(target.capacity());
-    if ((wstr_size + 1) * 2 > result_size) {
+    if ((wstr_size + 1) * 4 > result_size) {
```
gabime, line 453: Line 442 should also be fixes then.

Tomas-Zhu: Get it,I push again,please review.

gabime:
Thanks @Tomas-Zhu

PR #3049, opened 2024-03-22
Make async_logger::flush() synchronous and wait for the flush to comp…

walkerlala:
This patch make the async_logger::flush() interface a synchronous operation, similar to that of normal logger::flush(). The flush operation return only after it succeeds or fail. If failed, the error handler would be invoked with a exception message.

This patch also modify the minimum c++ stdandard to be c++14, to be able to use functions like std::make_unique().

### DIFF
--- a/include/spdlog/details/thread_pool-inl.h
+++ b/include/spdlog/details/thread_pool-inl.h
@@ -62,9 +62,13 @@ void SPDLOG_INLINE thread_pool::post_log(async_logger_ptr &&worker_ptr,
     post_async_msg_(std::move(async_m), overflow_policy);
 }
 
-void SPDLOG_INLINE thread_pool::post_flush(async_logger_ptr &&worker_ptr,
-                                           async_overflow_policy overflow_policy) {
-    post_async_msg_(async_msg(std::move(worker_ptr), async_msg_type::flush), overflow_policy);
+std::future<void> SPDLOG_INLINE thread_pool::post_flush(async_logger_ptr &&worker_ptr,
+                                                        async_overflow_policy overflow_policy) {
+    std::promise<void> promise;
+    std::future<void> future = promise.get_future();
+    post_async_msg_(async_msg(std::move(worker_ptr), async_msg_type::flush, std::move(promise)),
+                    overflow_policy);
+    return future;
 }
 
 size_t SPDLOG_INLINE thread_pool::overrun_counter() { return q_.overrun_counter(); }
@@ -108,6 +112,7 @@ bool SPDLOG_INLINE thread_pool::process_next_msg_() {
         }
         case async_msg_type::flush: {
             incoming_async_msg.worker_ptr->backend_flush_();
+            incoming_async_msg.flush_promise.set_value();
             return true;
         }
 
--- a/tests/test_async.cpp
+++ b/tests/test_async.cpp
@@ -93,6 +93,49 @@ TEST_CASE("flush", "[async]") {
     REQUIRE(test_sink->flush_counter() == 1);
 }
 
+TEST_CASE("multithread flush", "[async]") {
+    auto test_sink = std::make_shared<spdlog::sinks::test_sink_mt>();
+    size_t queue_size = 2;
+    size_t messages = 10;
+    size_t n_threads = 10;
+    size_t flush_count = 1024;
+    std::mutex mtx;
+    std::vector<std::string> errmsgs;
+    {
+        auto tp = std::make_shared<spdlog::details::thread_pool>(queue_size, 1);
+        auto logger = std::make_shared<spdlog::async_logger>(
+            "as", test_sink, tp, spdlog::async_overflow_policy::discard_new);
+
+        logger->set_error_handler([&](const std::string &) {
+            std::unique_lock<std::mutex> lock(mtx);
+            errmsgs.push_back("Broken promise");
+        });
+
+        for (size_t i = 0; i < messages; i++) {
+            logger->info("Hello message #{}", i);
+        }
+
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if(SPDLOG_USE_STD_FORMAT)
     set(CMAKE_CXX_STANDARD 20)
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
 elseif(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD 14)
```
gabime: This breaks backward compatibility. In any case i dont think unique ptr should be involved. see the usage example in https://en.cppreference.com/w/cpp/thread/promise

include/spdlog/details/thread_pool-inl.h
```diff
@@ -101,26 +106,34 @@ bool SPDLOG_INLINE thread_pool::process_next_msg_() {
...
+            break;
         }
 
         default: {
             assert(false);
         }
     }
 
-    return true;
+    if (incoming_async_msg.promise) {
```
gabime: No need for all those changes. since promise only used in flush, it can be dealt directly in the switch case and not involve the rest of the code. Also for clarity I would call the struct member it async_msg.flush _promise since this is its prupose.

gabime (changes requested):

include/spdlog/details/thread_pool-inl.h
walkerlala: Done.

CMakeLists.txt
walkerlala: Sure. I was thinking that using a ptr would make it more flexible so that other kind of log message (e.g., terminate) could add a future dynamically. But let's see whether we would need it in the future.

tests/test_async.cpp
```diff
@@ -93,6 +93,49 @@ TEST_CASE("flush", "[async]") {
...
+        }
+
+        for (auto &t : threads) {
+            t.join();
+        }
+    }
+    REQUIRE(test_sink->flush_counter() >= 1);
+    REQUIRE(test_sink->flush_counter() + errmsgs.size() == n_threads * flush_count);
+    REQUIRE(errmsgs.size() >= 1);
+    REQUIRE(errmsgs[0] == "std::future_error: Broken promise");
```
gabime: It shouldn't rely on non standard error string.

gabime (changes requested):

gabime:
It won’t help if the queue was created with discard flag and the flush message gets discarded..

tests/test_async.cpp
walkerlala: Done

walkerlala:
> It won’t help if the queue was created with discard flag and the flush message gets discarded..

I don't understand the problem. Could you clarify?

I you mean that the flush signal gets discarded: if the async_msg containing a flush promise gets discard, the promise would be "broken" and then the future.get() call would get and broken promise exception, which would be passed into the error handler for the logger.

gabime:
Yes, and this means that again flush() won’t work as expected even after this fix. If queue is full, flush() fails without any indication(returns void).

walkerlala:
> Yes, and this means that again flush() won’t work as expected even after this fix. If queue is full, flush() fails without any indication(returns void).

If queue is full, flush() will fail, but, there is indication: a std::promise destructed without calling set_value() will cause exception to be thrown. Here is an example:

```
walkerlala@linuxbox:/tmp
> cat promise.cc
#include <future>
#include <iostream>
#include <stdexcept>

std::future<int> promise_no_set_value() {
    std::promise<int> p;
    std::future<int> f = p.get_future();
    return f;
}

int main() {
    auto f = promise_no_set_value();
    try {
        int res = f.get();
        std::cout << "Got future result: " << res << std::endl;
    } catch (const std::future_error& e) {
        std::cerr << "Caught a future_error: " << e.what() << std::endl;
    } catch (const std::exception& e) {
        std::cout << "exception: " << e.what() << std::endl;
    } catch (...) {
        std::cout << "unknown exception" << std::endl;
    }
}
walkerlala@linuxbox:/tmp
> g++ -Wall -std=c++11 promise.cc -o promise && ./promise
Caught a future_error: std::future_error: Broken promise
```

So there would be indicatin through the global error handler.

I do noticed the issue that flush() does not return anything. But changing this would change the interface and would break API compatibility, so it might need to go to v2.x. And since spdlog use exception and logger error handler for error propogation, I continue to use this pattern.

As for queue-full-and-drop-message design, I think this is by design and not needed to change. And if we do need to change that, it would not be a easy task because the current design share thread pool and queue between multiple logger.

walkerlala:
By the way, I am seeing "Some checks were not successful" and "1 review requesting changes" but I don't understand what is requested to be changed.

gabime (approved):

gabime:
> So there would be indicatin through the global error handler.

Ok, seems this is the best we can do without breaking the API. 
In any case the situation improves after this fix, so I will merge.

gabime:
Thanks @walkerlala

gabime:
@walkerlala Can you check why msvc test fails [here](https://ci.appveyor.com/project/gabime/spdlog/builds/49467523/job/4s5pq32kh9ru3568)?

walkerlala:
> @walkerlala Can you check why msvc test fails [here](https://ci.appveyor.com/project/gabime/spdlog/builds/49467523/job/4s5pq32kh9ru3568)?

Sure. I would check it as soon as possible.

Issue #3050, opened 2024-03-22
Lambda Expression Does Not Compile in GCC 4.8.1 Version


KaganCanSit:
### General Info

* Hi, I recently wanted to update the spdn library I was using in my project. The version I was using was version 1.9.2. When I update the library contents and header files and start compiling, I get an error while compiling the expression containing **Lambda Expression**.

### Error Case
> In file included from ../../../common/spdlog/fmt/bundled/core.h:3321:0,
                 from ../../../common/spdlog/fmt/fmt.h:28,
                 from ../../../common/spdlog/common.h:50,
                 from ../../../common/spdlog/spdlog.h:12,
                 from ../../../common/log/Log.cpp:22:
../../../common/spdlog/fmt/bundled/format.h: In static member function 'static int fmt::v9::detail::count_digits(UInt) [with int BITS = 1; UInt = __int128 unsigned]::__lambda2::_FUN(__int128 unsigned)':
../../../common/spdlog/fmt/bundled/format.h:1110:11: internal compiler error: in expand_expr_addr_expr_1, at expr.c:7628
   return [](UInt m) {
           ^
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://sourceforge.net/projects/mingw-w64> for instructions.

### Workaround (I'm not sure, I didn't have enough time.)
* The part determined as the error point is located in the "format.h" file. [This](https://github.com/gabime/spdlog/blob/73e2e02b42b2d52fc07b88ceb7f91cf95c5f564f/include/spdlog/fmt/bundled/format.h#L1110) When I write the part here as a general function without a lambda expression, I can compile and use it without getting this error.

`int num_digits = 0;`
`do {`
`++num_digits;`
`} while ((n >>= BITS) != 0);`
`return num_digits;`
 
### Information is provided as follows in the Install information file.
> Tested on:
> gcc 4.8.1 and above
> clang 3.5
> Visual Studio 2013

### Information about my development environment;
OS: Windows 10
IDE: Netbeans 8.1
GCC and G++ Version: 4.8.1
![spdnLogBug](https://github.com/gabime/spdlog/assets/59123356/f5c7d283-7c34-40d0-bf77-df7acbf86021)

**Am I missing something? Maybe revision differences may cause this. Or can we find a better solution?**

tt4g:
Duplicate #2333, #2834, #2835

gcc 4.8 is not full support C++11.

KaganCanSit:
Thanks your answer,
[...]

PR #3052, opened 2024-03-22
 INSTALL.md has been updated to provide current status information. 

KaganCanSit:
The purpose of this regulation is to better inform the user about many current issues.

Spdlog uses fmt repository directly. Compilation errors occurring in file contents obtained from this repository are not handled by the spd community. (Consider the cost and complexity of retroactive support.)

You can read general information about this issue in the issue content (#3050)

**! Note:** 
- Minimum gcc version information and requirements can be discovered later.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -1,24 +1,27 @@
-Header only version:
+Header Only Version
 ==================================================================
-Just copy the files to your build tree and use a C++11 compiler.
+Just copy the files to your build tree and use a C++11 compiler.  
 Or use CMake:
+``` 
   add_executable(example_header_only example.cpp)
   target_link_libraries(example_header_only spdlog::spdlog_header_only)
+```
 
-
-Compiled library version:
+Compiled Library Version
 ==================================================================
 CMake:
+```  
   add_executable(example example.cpp)
   target_link_libraries(example spdlog::spdlog)
+```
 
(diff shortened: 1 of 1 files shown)


gabime:
Instead of removing this info, lwts find which min version do work. Also the fmt warning belongs to the wiki FAQ instead of main readme.

KaganCanSit:
@gabime Thanks for the guidance. I can remove the fmt warning from install.md. Or it could be in two places. I don't know. It's your decision.

Unfortunately, I do not have an environment where I can try GCC versions. If there is a tool for this, can you share it? Or if you have the information, let's update the test section.

If you direct me, I can take care of it in my free time.

gabime:
Yes, please remove the fmt warning from install.md. It belongs to the wiki.
Regarding trying gcc versions, the easiest is use gcc docker images (e.g https://hub.docker.com/_/gcc/tags?page=1&name=4.8 or https://hub.docker.com/_/gcc/tags?page=1&name=4.9).

KaganCanSit:
@gabime I made the necessary changes to INSTALL.md. Now I will write the gcc versions I tried in the issue. Let other people find it when they search for it. I may have made a mistake while testing. I haven't used Docker much!

gabime:
Thanks @KaganCanSit

PR #3092, opened 2024-05-21
Add wide character formatting and output support to console in wincolor_sink.

DominikGrabiec:
Fixes printing of unicode characters to the windows console such as microsecond suffix for `std::chrono` types.

This is a proper implementation of https://github.com/gabime/spdlog/pull/3013 as discussed in that issue. I had to open a new pull request because of some automated stuff Github did.

Please let me know if there's anything else I need to do.

For the v2 branch it might be useful to have a `wchar_t` output option for all Windows loggers as sometimes things like Visual Studio assume a certain encoding even in files.


gabime:
Thanks @DominikGrabiec

Issue #1838, opened 2021-02-21
Add color support on file sinks


patrick-rotsaert:
I'd like to have a rotating file sink with ansi coloring. Colored log files are a lot easier to read (e.g. with less -R) than plain text.
Colored logging to a file is possible using an `ansicolor_sink` with the `FILE *` c'tor, but then there is no rotation.
Any chance this may be added?

alex-176:
@gabime this issue is closed as completed. but there is no color_rotating_sink in the codebase (or i missed it?). i can add a PR

gabime:
It was closed because this can be achieved using custom formatter (or perhaps adding this functionality to the current pattern_formatter)

Also, there are many tools that display file contents with color.

alex-176:
thanks

patrick-rotsaert:
Please see this [PR](https://github.com/gabime/spdlog/pull/3112).

PR #3112, opened 2024-06-16
Ansi colored file sink

patrick-rotsaert:
This PR addresses #1838.

I could not make it work with the [suggested](https://github.com/gabime/spdlog/issues/1838#issuecomment-1177245221) approach. The `details::log_msg` members `color_range_start` and `color_range_end` are ignored in the methods `basic_file_sink::sink_it_` and `rotating_file_sink::sink_it_`.
Only the class `ansicolor_sink` uses the color range.

My first idea was to subclass `ansicolor_sink` and pass a stream to the opened file as the `FILE *target_file` argument to the base class constructor. However this had two problems:
1. The `ansicolor_sink::mutex_` member is a reference to `ConsoleMutex::mutex()` which is a singleton. This is not suitable for file sinks.
2. File rotation is not easily added.

To address this, I factored out the ANSI coloring and file rotation into separate classes, to be reused in the new sink classes `ansicolor_file_sink` and `ansicolor_rotating_file_sink`.

My apologies for the whitespace differences, my code editor is configured to remove trailing spaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ see example [CMakeLists.txt](https://github.com/gabime/spdlog/blob/v1.x/example/
 * [Custom](https://github.com/gabime/spdlog/wiki/3.-Custom-formatting) formatting.
 * Multi/Single threaded loggers.
 * Various log targets:
-  * Rotating log files.
+  * Rotating log files (ANSI colors supported).
   * Daily log files.
   * Console logging (colors supported).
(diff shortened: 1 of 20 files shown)


gabime:
Thanks. However, like I said before, this is a very specific need that can be solved using custom formatter, other tools, or  a custom sink. So I don't think it should merged.

PR #3124, opened 2024-07-03
Allow customization of syslog_sink

Silex:
I also took the liberty of allowing access to `syslog_levels_`, this would simplify my implementation. Feel free to disagree.


Silex:
Thanks! 👍

Issue #3083, opened 2024-05-02
Async logging using MDC


felixjulianheitmann:
With https://github.com/gabime/spdlog/pull/2907 it became possible to log additional key-value formatted attributes with a message.

We like the idea could use something like this for a first step in the direction of structured logging.

The approach is a little limited, though. In our app, we use multiple different async loggers. The PR #2907 uses thread local storage to store the key-value pairs.

This means, multi-threaded systems get different messages depending on which thread logs the actual message. Async logging is not possible this way as noted by the original author.

Additionally, even in a single-threaded application multiple loggers would all be logging the same attributes as they all query the same global context variable. This is not ideal - at least not for our use-case.

To allow async logging, we would need to make the `context` variable not a `thread_local` static variable but an actual global static variable:
https://github.com/gabime/spdlog/blob/27cb4c76708608465c413f6d0e6b8d99a4d84302/include/spdlog/mdc.h#L41-L42

Additionally, of course all access would need to be secured with a mutex. With a static class member however, we would need a source file to define it in (at least in C++11).

That only solves the async logging. Not the global state across all loggers.

To solve the second issue, we would need to make the `mdc` class/context a member of the `mdc_formatter`.
https://github.com/gabime/spdlog/blob/27cb4c76708608465c413f6d0e6b8d99a4d84302/include/spdlog/pattern_formatter-inl.h#L790

At that point we are already instantiating the `mdc` class and don't need to resort to `static` and `thread_local` variables. We could just make the context a regular member that gets initialized on creation.

This is a breaking change to the previous behavior, however.

What I am getting at is the following:

- do you think it's work investigating this further?
- how would we deal with backwards compatibility?
  - runtime options: incur a runtime overhead
  - compile time options: incur complexity during testing / conditional builds

I am pretty confident, I can implement this and make a PR but I would rather get your blessing about how you want me to approach this, beforehand.

felixjulianheitmann:
An MWE which unfortunately, doesn't compile on godbolt is this:
```c++
#include <spdlog/spdlog.h>
#include <spdlog/sinks/stdout_sinks.h>
#include <spdlog/mdc.h>

int main()
{
    auto create_logger = [](auto name) {
        auto logger = std::make_shared<spdlog::logger>(
            name,
            std::make_shared<spdlog::sinks::stdout_sink_st>()
        );
        logger->set_pattern("[%H:%M:%S %z] [%^%L%$] [%&] %v");
        logger->info("Logger created");
        return logger;
[...]

PR #2907, opened 2023-10-13
Added Mapped Diagnostic Context (MDC) support

massimiliano96:
### Description
This pull request introduces support for Mapped Diagnostic Context (MDC). MDC allows users to associate contextual information with log events, providing valuable insights into specific execution flows.

### Changes:
Added mdc_formatter class to handle MDC formatting.
Introduced MDC class to manage contextual information.
Provided test cases to validate MDC functionality.

### Usage:
```
// Setting MDC values
spdlog::mdc::put("mdc_key_1", "mdc_value_1");
spdlog::mdc::put("mdc_key_2", "mdc_value_2");

// Creating a logger with MDC support
auto formatter = std::make_shared<spdlog::pattern_formatter>("\n");
formatter->set_pattern("[%n] [%l] [%&] %v");

// Logging a message
memory_buf_t formatted;
spdlog::details::log_msg msg(spdlog::source_loc{}, "logger-name", spdlog::level::info, "some message");
formatter->format(msg, formatted);

// Expected output: [logger-name] [info] [mdc_key_1:mdc_value_1 mdc_key_2:mdc_value_2] some message

// Removal of an MDC value
spdlog::mdc::remove("mdc_key_2");

```

### Test Cases:

- Validated MDC formatting with multiple values.
- Checked MDC value update.
- Tested MDC behavior across different threads.
- Verified MDC key removal.
- Empty MDC

PR #3128, opened 2024-07-05
V2.x - Attribute support

felixjulianheitmann:
This is quite similar to the MDC support by #2907 except for one major difference: MDC is stored in global thread local storage vs. attributes are stored per logger.

Each logger has an `attributes_` map which is a `std::string/std::string` map. During logging, these attributes are passed with the log message to the formatter. An `attributes_formatter` (`%*`) will format these the same way that the `mdc_formatter` does.

I could use inheritance or a common function to reduce the code duplication.

## What's the use case?

We are using spdlog on a webserver and have multiple threads handling requests. They share a common logger instance. Since there are multiple threads writing to the same logger, we cannot use MDC to associate additional context to the log messages. The thread writing to the sink will always be a different one than the one executing the web-handler.

This approach will provide the ability to have context associated with a logger instance rather than with a thread. I can have multiple threads writing to the same logger and providing context.

### 1. It's thread safe but is it useful?

After the implementation, I noticed a flaw in my train of thoughts. If I have multiple threads that can safely put context on a logger and write them together with the log messages, I still get a form of race condition.

Logger `A` wants to do some file operation and calls `logger.attrs().put("file-path", "my/path.txt")` to have the file path be logged with any subsequent logs.

Logger `B` may fail parsing an incoming JSON and calls `logger.attrs().put("error", exception.what())`.

If these two then log "at the same time", the logging message will contain both context attributes.

Maybe I shouldn't reuse the same logger for entirely different operations. But even if they are similar, they are not the same and logs from one task don't matter for the other. I am now quite confident, this is what @tt4g tried to explain to me in https://github.com/gabime/spdlog/issues/3083#issuecomment-2092550414 that I am just grasping now.

### 2. using a different map
Currently every attribute has to be formatted as a string before inserting it into the API. Maybe this can be worked around by providing a `formattable` type or similar in the attributes API rather than a string. I haven't dug into fmtlib/std::format enough to see whether that's possible somehow without going down the template rabbit hole.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,7 +146,8 @@ set(SPDLOG_HEADERS
         "include/spdlog/sinks/systemd_sink.h"
         "include/spdlog/sinks/tcp_sink.h"
         "include/spdlog/sinks/udp_sink.h"
-        "include/spdlog/sinks/async_sink.h")
+        "include/spdlog/sinks/async_sink.h"
+        "include/spdlog/attributes.h")
 set(SPDLOG_SRCS
         "src/common.cpp"
         "src/logger.cpp"
--- a/include/spdlog/attributes.h
+++ b/include/spdlog/attributes.h
@@ -2,14 +2,15 @@
 
 #include <spdlog/common.h>
 
+#include <map>
 #include <mutex>
-#include <unordered_map>
+
 
 namespace spdlog {
 
 class SPDLOG_API log_attributes {
 public:
-    using attr_map_t = std::unordered_map<std::string, std::string>;
+    using attr_map_t = std::map<std::string, std::string>;
     using key_t = attr_map_t::key_type;
     using value_t = attr_map_t::mapped_type;
     using const_iter = attr_map_t::const_iterator;
--- a/include/spdlog/attributes.h
+++ b/include/spdlog/attributes.h
@@ -1,0 +1,88 @@
+#pragma once
+
+#include <spdlog/common.h>
+
+#include <mutex>
+#include <unordered_map>
+
+namespace spdlog {
+
+class SPDLOG_API log_attributes {
+public:
+    using attr_map_t = std::unordered_map<std::string, std::string>;
+    using key_t = attr_map_t::key_type;
+    using value_t = attr_map_t::mapped_type;
+    using const_iter = attr_map_t::const_iterator;
+
+    log_attributes() = default;
+    log_attributes(const log_attributes& other) {
+        const auto& o_map = other.get_map();
+        attr_ctx(o_map.begin(), o_map.end());
+    }
+    log_attributes& operator=(const log_attributes& other) {
+        if (this != &other) {
+            clear();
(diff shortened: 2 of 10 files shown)


felixjulianheitmann:
> 1. It's thread safe but is it useful?

Maybe the goal is slightly different. Maybe, I do want thread local storage but on a per-logger basis and I would want my contextual info to be forwarded to the thread pool actually writing to the sinks.

I cannot use single-threaded loggers but I also don't really see the benefit of having multiple threads mix their current individual context

felixjulianheitmann:
Okay, some builds seem to fail. I will have a look at that next week.

felixjulianheitmann:
Regardless, of whether the additional builds pass, this PR is in a philosophical dilemma.
As far as I can see, there is two ways of implementing contextual attributes for log messages:

1. an attribute map valid across all logger instances on a per thread basis
   -> current MDC implementation does that
2. an attribute map valid across all threads on a per logger instane basis
   -> this PR implements that.

Shortcomings of 1.:
- no async loggers possible
- unrelated loggers logging on the same thread share the context state

Shortcomings of 2.:
- contextual info is shared across possibly unrelated tasks. Our use-case of a thread pool for web-handlers sharing a single logger instance would mix the contextual state of possibly different tasks.

A final less convenient solution: per-message context info
We could overload the `info`/`warn`/`error`/... calls to allow the inclusion of per message contextual info. That kind of defeats the purpose though. If I want to include additional info in _this specific message_ ... just write it into the message.

My bottom line is: this PR makes sense to me, personally. For the project I want to use it for, our structure needs to be updated. If we used a separate logger instance per web-handler thread pool, this PR would perfectly solve our issues.

I will go ahead and see, if I can fix the build issues and discuss internally, how to proceed.

tt4g:
As far as I know, logging libraries are torn between implementing either of these two.
Personally, I prefer this PR implementation.

felixjulianheitmann:
Future improvements could include templated versions that accept any form of attribute type that's formattable by the fmt library. This implementation is kept simple for a first implementations sake.

Any future work on this should be doable without breaking the interface.

bobhansen:
In https://github.com/bobhansen/structured_spdlog, it was important that the data context be thread local (so inner library functions could include context they were oblivious to), but also be able to pass it between threads for context-sharing (e.g. a webserver passing the session ID to worker threads).

We included an explicit method to snapshot and assume contexts (which we implemented as linked lists of `std::shared_ptr`s).

felixjulianheitmann:
Explicit methods for passing the thread memory boundary or for putting something in thread local memory seems reasonable.

This has been open for quite some time and we've been thinking about it. Usually, the context you're working on stays on the same thread. You push context, perform the work and kill the context.

If you have specific worker threads - a web server or some offloaded calculations -, these threads are started explicitly and are loaded with work deliberately. That's where it seems reasonable to also push the context onto the thread.

It doesn't fit our current code architecture but seems to make the most sense. I like your approach

PR #3132, opened 2024-07-09
Clear previous background color for win32 console before setting new one

asinpro:


gabime:
Could you please provide more details why is it needed ? so far it worked fine

asinpro:
As stated in the c++ comment above, we "change only the foreground bits(low 4 bits)", but these are only low two bits. And we don't take to background color which can be changed globally in console.

PR #3139, opened 2024-07-21
fix: set `/Zc:__cplusplus` and `/MP` to MSVC only

zjyhjqs:
1. macro `__cplusplus` is enabled by clang-cl
2. `/MP` is not supported by clang-cl (warning `-Wunused-command-line-argument` will be generated)

see also: https://clang.llvm.org/docs/UsersManual.html#id11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ elseif(NOT CMAKE_CXX_STANDARD)
 endif()
 
 # make sure __cplusplus is defined when using msvc and enable parallel build
-if(MSVC)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     string(APPEND CMAKE_CXX_FLAGS " /Zc:__cplusplus /MP")
 endif()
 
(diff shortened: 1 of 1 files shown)


gabime:
I dont get it. MSVC is defined by cmake when using msvc compiler (https://cmake.org/cmake/help/latest/variable/MSVC.html)

zjyhjqs:
> Set to true when the compiler is some version of Microsoft Visual C++ **_or another compiler simulating the Visual C++ cl command-line syntax_**.

MSVC is also defined when clang-cl is used.

CMakeLists.txt
```diff
@@ -161,7 +161,7 @@ if(SPDLOG_BUILD_SHARED OR BUILD_SHARED_LIBS)
     endif()
     add_library(spdlog SHARED ${SPDLOG_SRCS} ${SPDLOG_ALL_HEADERS})
     target_compile_definitions(spdlog PUBLIC SPDLOG_SHARED_LIB)
-    if(MSVC)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
```
tt4g, line 164: This `if` statement is not necessary because `$<CXX_COMPILER_ID:MSVC>` appears in the following cmake-generator-expressions.

```cmake
        target_compile_options(spdlog PUBLIC $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<NOT:$<COMPILE_LANGUAGE:CUDA>>>:/wd4251
                                             /wd4275>)
```

gabime: > appears in the following cmake-generator-expressions.

but it is reached only if build shared.

tt4g: > but it is reached only if build shared.

I overlooked it. In that sense, we need to set the same option for `spdlog::spdlog_header_only`.

gabime: perhaps, but it is not directly related to this pr, so i will approve for now

gabime:
Merged. Thanks @zjyhjqs

PR #3154, opened 2024-08-11
Add syslog_formatter

sivachandran:
The formatter enables remote(udp/tcp) syslog sink

Example:
```c++
    // run "nc -klu 514" to receive the syslog message
    spdlog::sinks::udp_sink_config udp_sink_config("127.0.0.1", 514);
    auto udp_logger = spdlog::udp_logger_mt("udp_logger", udp_sink_config);
    spdlog::syslog_formatter syslog_formatter(spdlog::facility::user, "localhost", "example");
    udp_logger->set_formatter(syslog_formatter.clone());
    udp_logger->info("this message will be logged in local syslog receiver");
```

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -27,6 +27,7 @@ void custom_flags_example();
 void file_events_example();
 void replace_default_logger_example();
 void mdc_example();
+void syslog_formatter_example();
 
 #include "spdlog/spdlog.h"
 #include "spdlog/cfg/env.h"   // support for loading levels from the environment variable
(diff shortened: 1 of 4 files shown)


gabime:
Thanks. Since this formatter’s sole purpose  is for remote syslog, i suggest having a “remote_syslog” sink instead which would use udp/tcp sink under the hood with the right formatting. This also would make usage much easier without having to use a custom formatter.
Also note that some of the ci tests fail.
I will close this for now (PR with a new sink is welcome).

PR #3155, opened 2024-08-13
android: Support __android_log_write_log_message based logging

jplcz:
This function has been introduced in API Level 30. This commit enables support for dynamic use of the new API (in case the application would run on older OS versions).


gabime:
Thanks. However all those ifdefs makes this sink code bit messy. what this new api gives us anyway? Also the android doc state that __android_log_is_loggable should be used before using.

PR #3157, opened 2024-08-17
Initialize function-local static variables using "T& t = *new T"

lixingcong:
It would be safer if init the static object with reference according to [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables).

I compile the static lib, link to my program, then get segmentation fault on calling  `mdc::get_context()`. After applying these patches the bug was gone.  :100:

### DIFF
--- a/include/spdlog/details/console_globals.h
+++ b/include/spdlog/details/console_globals.h
@@ -12,7 +12,7 @@ namespace details {
 struct console_mutex {
     using mutex_t = std::mutex;
     static mutex_t &mutex() {
-        static mutex_t s_mutex;
+        static mutex_t& s_mutex = *new mutex_t;
         return s_mutex;
     }
 };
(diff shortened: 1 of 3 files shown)


gabime:
Thanks but now tests fails with “LeakSanitizer: detected memory leaks”. A possible solution would be to delete those objects in spdlog::shutdown() and call it at the end of tests.

lixingcong:
So hard to free memory of `thread_local` object created by `T& t=*new T`

```cpp
#include <iostream>
#include <thread>

struct A {
	A(int i)
	    : a(i)
	{
		std::cout << "ctor(), i=" << a << std::endl;
	}

	~A()
	{
		std::cout << "dtor(), i=" << a << std::endl;
	}

	const int a;
};

static A& get_a()
{
	static thread_local A& a = *new A(8);
	static thread_local A  b(100);
	return a;
}

int main()
{
	std::thread t([]() { get_a(); });
	t.join();
	return 0;
}
```

Compile with gcc 11 it got weird print:

```
ctor(), i=8
ctor(), i=100
dtor(), i=100
```

That is why LeakSanitizer tell us memory leak happened.

gabime:
This has to solved somehow

moi15moi:
I tested this PR.
The first commit does fix the segmentation fault, but the second doesn't.

PR #3158, opened 2024-08-26
V1.x : fixed "suggest override" errors for gcc 8.5 and gcc 9.1 

esmirno:
fixed errors for gcc 8.5 and gcc 9.1 compilers, which seems reporting an error if no override presented with final. 
I saw FMT_OVERRIDE option used to be present in the repo but then gone, I hope due to it is not an error on newer compiler to have both specifiers.

checked here:
https://godbolt.org/z/33csE9W1P


gabime:
Thanks @esmirno

PR #3170, opened 2024-09-09
Add info about max_files in the docstrings of hourly/daily file sinks

themightyoarfish:
What branch should I target here?


gabime:
Thanks @themightyoarfish

PR #3172, opened 2024-09-10
[docs] Update conan install command in README

uilianries:
Hello!

I observed that the Conan "instruction" in the readme file is only pointing the spdlog version to be installed, but not the command itself. I updated it with the entire command.

The latest spdlog available in Conan Center can be found on the web page: https://conan.io/center/recipes/spdlog?version=1.14.1

About Conan 2.x install command: https://docs.conan.io/2/reference/commands/install.html

The `[*]` indicates to install the latest version available.


gabime:
Thanks @uilianries

PR #3179, opened 2024-09-11
use std::lock_guard instead of std::unique_lock

wsehjk:
I think `overrun_counter`, `size`, and `reset_overrun_counter` func don't need the flexibility of `std::unique_lock`, it's better to use `std::lock_guard` with minimal overhead;


gabime:
Merged. Thanks @wsehjk

PR #3225, opened 2024-10-20
fixing vs analyser's "here could be constexpr" warning

wtom76:
Please consider replacing const with constexpr which an analyser marked with
"warning C26498: The function '...' is constexpr, mark variable '...' constexpr if compile-time evaluation is desired (con.5)."


tt4g:
The source code in the `include/spdlog/fmt/bundled/` directory is source code copied from the [fmt repository](https://github.com/fmtlib/fmt).
Please send this PR to the [fmt repository](https://github.com/fmtlib/fmt).

PR #3228, opened 2024-10-25
Exchange promise for condition_variable when flushing (fixes #3221)

### LINKED ISSUE #3221: 1.14.1 unresolvable R_X86_64_TPOFF32 relocation
Updating spdlog to 1.14.1, I get the following linker error:
```
/usr/bin/c++ -Wno-maybe-uninitialized  -ggdb -O3 -DNDEBUG -pie -Wl,-z,nodlopen -Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now -Wl,--as-needed -Wl,--no-copy-dt-needed-entries test/CMakeFiles/EtcdTests.dir/EtcdTests.cpp.o -o /opt/ichor/src/bin/EtcdTests  /opt/ichor/src/bin/libichor.a  /opt/ichor/src/bin/libCatch2Main.a  -lsystemd  /usr/lib/libboost_coroutine.a  /usr/lib/libboost_fiber.a  /usr/lib/libboost_context.a  /usr/lib/libboost_filesystem.a  /usr/lib/libssl.a  /usr/lib/libcrypto.a  -ldl  -lhiredis  -ldl  -lrt  /opt/ichor/src/bin/libCatch2.a && :
/usr/bin/ld: /opt/ichor/src/bin/libichor.a(async.cpp.o)(.text+0x266e): unresolvable R_X86_64_TPOFF32 relocation against symbol `_ZSt15__once_callable@@GLIBCXX_3.4.11'
```

This doesn't happen with 1.13.0, or when `ftls-model=local-exec` is removed from the compiler flags.

Compiler: gcc 12.3.0
Linux: Ubuntu Jammy
cxxflags: `-O2 -std=c++20 -fpie -fstack-protector-strong -fcf-protection -fstack-clash-protection -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -D_GLIBCXX_ASSERTIONS -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_FAST -ftls-model=local-exec`
linkflags: `-ggdb -O3 -DNDEBUG -pie -Wl,-z,nodlopen -Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now -Wl,--as-needed -Wl,--no-copy-dt-needed-entries`

Oipo:
### Description

std::promise and std::future use std::call_once under the hood, which requires the tls-model to be at least initial_exec, excluding local_exec.

Furthermore, gcc has a bug regarding exceptions in std::call_once that is best avoided. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66146 for more info.

Coincidentally, this change also fixes the flush functionality when compiling with vs2013, as the move constructor/operator for vs2013 didn't move the `flush_promise` member. I also initially missed this, as I don't test on windows.

### Considerations

* Each flush request now uses a `std::condition_variable` that checks every 100 ms for a flag change or until a `cv.notify_all()` happens. The latter might race for some reason, so there may be a delay of up to 100 ms until the change is detected in rare cases.
* The `std::future` solution threw exceptions if the `async_msg` got dropped. This behaviour is still there, though it is technically possible to return a boolean now and call the `err_handler_` without an exception, saving the cost of throwing one at the expense of an if statement on the non-error case. I don't have any benchmarks on if that would be a worthy tradeoff.

### Tested on

Tested on linux x64, with and without sanitizers, with `-ftls-model=local-exec` and `-ftls-model=initial-exec`.

Please let me know if you want to see any changes.

### DIFF
--- a/include/spdlog/async_logger-inl.h
+++ b/include/spdlog/async_logger-inl.h
@@ -32,28 +32,30 @@ SPDLOG_INLINE spdlog::async_logger::async_logger(std::string logger_name,
 
 // send the log message to the thread pool
 SPDLOG_INLINE void spdlog::async_logger::sink_it_(const details::log_msg &msg){
-    SPDLOG_TRY{if (auto pool_ptr = thread_pool_.lock()){
-        pool_ptr->post_log(shared_from_this(), msg, overflow_policy_);
-}
-else {
-    throw_spdlog_ex("async log: thread pool doesn't exist anymore");
-}
-}
-SPDLOG_LOGGER_CATCH(msg.source)
+    SPDLOG_TRY {
+        if (auto pool_ptr = thread_pool_.lock()){
+            pool_ptr->post_log(shared_from_this(), msg, overflow_policy_);
+        }
+        else {
+            throw_spdlog_ex("async log: thread pool doesn't exist anymore");
+        }
+    }
+    SPDLOG_LOGGER_CATCH(msg.source)
 }
 
 // send flush request to the thread pool
-SPDLOG_INLINE void spdlog::async_logger::flush_(){SPDLOG_TRY{auto pool_ptr = thread_pool_.lock();
-if (!pool_ptr) {
-    throw_spdlog_ex("async flush: thread pool doesn't exist anymore");
-}
+SPDLOG_INLINE void spdlog::async_logger::flush_() {
+    SPDLOG_TRY {
+        auto pool_ptr = thread_pool_.lock();
+        if (!pool_ptr) {
+            throw_spdlog_ex("async flush: thread pool doesn't exist anymore");
+        }
 
-std::future<void> future = pool_ptr->post_flush(shared_from_this(), overflow_policy_);
-// Wait for the flush operation to complete.
-// This might throw exception if the flush message get dropped because of overflow.
-future.get();
-}
-SPDLOG_LOGGER_CATCH(source_loc())
+        // Wait for the flush operation to complete.
+        // This might throw exception if the flush message get dropped because of overflow.
+        pool_ptr->post_and_wait_for_flush(shared_from_this(), overflow_policy_);
+    }
+    SPDLOG_LOGGER_CATCH(source_loc())
 }
 
 //
--- a/include/spdlog/details/thread_pool-inl.h
+++ b/include/spdlog/details/thread_pool-inl.h
@@ -62,13 +62,25 @@ void SPDLOG_INLINE thread_pool::post_log(async_logger_ptr &&worker_ptr,
     post_async_msg_(std::move(async_m), overflow_policy);
 }
 
-std::future<void> SPDLOG_INLINE thread_pool::post_flush(async_logger_ptr &&worker_ptr,
+void SPDLOG_INLINE thread_pool::post_and_wait_for_flush(async_logger_ptr &&worker_ptr,
                                                         async_overflow_policy overflow_policy) {
-    std::promise<void> promise;
-    std::future<void> future = promise.get_future();
-    post_async_msg_(async_msg(std::move(worker_ptr), async_msg_type::flush, std::move(promise)),
-                    overflow_policy);
-    return future;
+    std::mutex m;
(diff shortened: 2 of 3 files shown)


gabime:
Thanks. Why the 100 ms loop?

Oipo:
> Thanks. Why the 100 ms loop?

Unfortunately, changing the `wait_for()` into a `wait()` without the loop results in the `condition_variable` sometimes getting stuck. My guess is that it is possible for the thread that is responsible for flushing to call `notify_all()` before the `condition_variable` is inside the `wait()`. That's also the main reason (other than spurious wakeups) that the `cv_flag` variable is necessary. To circumvent this racing behaviour.

gabime:
I think locking the mutex inside the callback (instead of atomic mechanism) would solve this. Also, the mutex and the cv should be members, otherwise it doesn't make sense to me.

Oipo:
> I think locking the mutex inside the callback (instead of atomic mechanism) would solve this. 

I tried it, this also results in hangs. I do not know why and I do not have more time to figure that out.

> Also, the mutex and the cv should be members, otherwise it doesn't make sense to me.

That would drastically change the semantics of the `flush()` call. Before the change to making the `flush()` call synchronous, every call would result in a flush message getting posted to the queue without waiting. Then, using promises, every call to `flush()` would result in a `mutex` and `condition_variable` being stored inside the promise for every `async_msg`, but there are no dependencies between each flush call. What this PR proposes is the same behaviour: every call to `flush()` only waits on a message posted in that call.

Making the `mutex` and `condition_variable` member variables would result in that every flush now has to wait for any previous flush to finish before being able to post a new one. Moving the posting of a message outside of the lock would result in data races because there is no guarantee that the `notify_one()` call wakes up the first thread. Which thread is woken up is explicitly unspecified. There's also a whole new can of worms that I consider outside the scope of this PR (which is just to fix a couple of bugs): what to do with multiple, relatively closely emitted, requests to flush? Coalesce them into a single wait? 

If member variables are still desired, then I would strongly advise to put them in `async_msg`, which would make use cases that do not call `flush()` pay for that in terms of memory and some extra constructors being called for each message.

gabime:
What about the 
```c++
if(cv_flag.load(std::memory_order_relaxed) == async_msg_flush::synced_not_flushed) {
        throw spdlog_ex("Request for flushing got dropped.");
    }
```
? When can this happen? (Wouldn’t the while loop would loop forever if the message got discarded?)

Oipo:
> ? When can this happen? (Wouldn’t the while loop would loop forever if the message got discarded?)

The [destructor of async_msg](https://github.com/gabime/spdlog/pull/3228/files#diff-0536d0a240341c31d644c353cece6fc3f4b2ef354a1e7c6227acb96c78653b09R37) calls the callback function if it is set. So no, it does not loop forever. This case is specifically tested in the [multhreaded flush test](https://github.com/gabime/spdlog/blob/22b0f4fc06980d82fb06d82a249215b98e3142fe/tests/test_async.cpp#L96).

gabime:
Merged. Thanks @Oipo

gabime:
> The [destructor of async_msg](https://github.com/gabime/spdlog/pull/3228/files#diff-0536d0a240341c31d644c353cece6fc3f4b2ef354a1e7c6227acb96c78653b09R37) calls the callback function if it is set. So no, it does not loop forever. This case is specifically tested in the [multhreaded flush test](https://github.com/gabime/spdlog/blob/22b0f4fc06980d82fb06d82a249215b98e3142fe/tests/test_async.cpp#L96).

@Oipo If i recall correctly it is not enough since the items in the underlying circular queue get recycled  rather than destructed using move assignments.

Oipo:
@gabime as far as I can see in [circular_q](https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/details/circular_q.h#L48), moving into an existing element of vector calls the move-assign operator.

Ah. I now see why the `wait` didn't work: the move-assign operator should swap the callbacks instead of setting one of them to `nullptr`. Let me open a new PR.

PR #3232, opened 2024-10-28
Ensure flush callback gets called in move-assign operator

Oipo:
While this change makes the condition_variable more reliable, without the 100 ms loop results in spdlog-utests getting stuck sometimes. With the `wait_for` loop, the [multithreaded flush test](https://github.com/gabime/spdlog/blob/22b0f4fc06980d82fb06d82a249215b98e3142fe/tests/test_async.cpp#L96), which inserts 10240 messages of which most get overridden, continues to work.

PR #3235, opened 2024-11-01
Gabime/async flush

gabime:
Revert PR #3049 to make ```async_logger->flush()``` fully async since it might get stuck under multi-thread load and since std::future have gcc problems (see #3221, #3228, #3232)


rtzoeller:
@gabime is there a way to force a synchronous flush after this change? I have some unit tests which validated that messages were being logged as expected, and they relied on calling `logger->flush();` to ensure stdout was updated.

gabime:
The closest I can think of is  wait until the queue is empty, send a flush message and then wait again for the queue to empty.

PR #3244, opened 2024-11-05
utf8_to_wstrbuf now handles invalid utf8 sequences

### LINKED ISSUE #3241: msvc_sink throws on invalid utf-8
This is what `msvc_sink` is doing:
```cpp
    #if defined(SPDLOG_WCHAR_TO_UTF8_SUPPORT)
        wmemory_buf_t wformatted;
        details::os::utf8_to_wstrbuf(string_view_t(formatted.data(), formatted.size()), wformatted);
        OutputDebugStringW(wformatted.data());
    #else
        OutputDebugStringA(formatted.data());
    #endif
```

Then in `utf8_to_wstrbuf` we have:
```cpp
    int result_size =
        ::MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, str.data(), str_size, NULL, 0);
```

`MB_ERR_INVALID_CHARS` causes the function to fail on invalid characters, and then `utf8_to_wstrbuf` throws.

We've run into this when trying to output game data that wasn't utf-8 encoded, had to roll out our own implementation of `msvc_sink` (https://github.com/OpenEnroth/OpenEnroth/pull/1825).

My feel here is that logging should work at best effort basis and should not throw unless absolutely necessary (as exception can lead to `std::terminate`). Log calls can be buried somewhere deep in error handling code, and the usual expectation there is that logging calls don't throw.

In this particular case writing out text that's a bit garbled is perfectly OK.

captainurist:
Fixes #3241

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -483,12 +483,12 @@ SPDLOG_INLINE void utf8_to_wstrbuf(string_view_t str, wmemory_buf_t &target) {
 
     // find the size to allocate for the result buffer
     int result_size =
-        ::MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, str.data(), str_size, NULL, 0);
+        ::MultiByteToWideChar(CP_UTF8, 0, str.data(), str_size, NULL, 0);
 
     if (result_size > 0) {
         target.resize(result_size);
-        result_size = ::MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, str.data(), str_size,
-                                            target.data(), result_size);
+        result_size = ::MultiByteToWideChar(CP_UTF8, 0, str.data(), str_size, target.data(),
+                                            result_size);
         if (result_size > 0) {
             assert(result_size == target.size());
             return;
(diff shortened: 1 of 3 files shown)


gabime:
Thanks @captainurist

gabime:
@captainurist Seems that some of the new tests [fail](https://ci.appveyor.com/project/gabime/spdlog/builds/50930381). Can you take a look?

captainurist:
Yes, looking at it now.

captainurist:
@gabime OK test failures are easy to fix, preparing a PR.

For the compilation errors, these make no sense to me:
```
C:\projects\spdlog\tests\test_misc.cpp(176): error C2678: binary '==': no operator found which takes a left-hand operand of type 'Catch::ExprLhs<spdlog::wmemory_buf_t &>' (or there is no acceptable conversion) [C:\projects\spdlog\build\tests\spdlog-utests.vcxproj]

...

C:\projects\spdlog\tests\test_misc.cpp(176): note: while trying to match the argument list '(Catch::ExprLhs<spdlog::wmemory_buf_t &>, const wchar_t [1])'
```

This comparison is not in the code. Let's rerun tests with the new PR and see if it helps resolve this too.

PR #3249, opened 2024-11-09
[docs] Removed duplicate line in daily_file_sink document

F1F88:
I removed a duplicate line in the daily_file_sink comment documentation.

Because I noticed that there were two identical comments in the daily_file_sink documentation in [v1.5.0](https://github.com/gabime/spdlog/blob/8e5613379f5140fefb0b60412fbf1f5406e7c7f8/include/spdlog/sinks/daily_file_sink.h#L64). 

There were none in [v1.4.0](https://github.com/gabime/spdlog/blob/0a53eafe18d983c7c8ba4cadd02d0cc7f7308f28/include/spdlog/sinks/daily_file_sink.h#L64).


F1F88:
I think this may have been an unintentional oversight.

Issue #3253, opened 2024-11-12
After using the fvisibility=hidden parameter, a direct access function warning appears during compilation


zhaowq32:
# warnging info
```
ld: warning: direct access in function 'std::__1::__shared_ptr_emplace<spdlog::sinks::rotating_file_sink<std::__1::mutex>, std::__1::allocator<spdlog::sinks::rotating_file_sink<std::__1::mutex> > >::__on_zero_shared()' from file 'CMakeFiles/spdlog_test.dir/main.cpp.o' to global weak symbol 'vtable for spdlog::sinks::base_sink<std::__1::mutex>' from file 'macos/spdlog/x86_64/Release/lib/libspdlog.a(spdlog.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
ld: warning: direct access in function 'std::__1::__shared_ptr_emplace<spdlog::sinks::rotating_file_sink<std::__1::mutex>, std::__1::allocator<spdlog::sinks::rotating_file_sink<std::__1::mutex> > >::__on_zero_shared()' from file 'CMakeFiles/spdlog_test.dir/main.cpp.o' to global weak symbol 'vtable for spdlog::sinks::rotating_file_sink<std::__1::mutex>' from file 'macos/spdlog/x86_64/Release/lib/libspdlog.a(file_sinks.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
ld: warning: direct access in function 'spdlog::sinks::rotating_file_sink<std::__1::mutex>::~rotating_file_sink()' from file 'CMakeFiles/spdlog_test.dir/main.cpp.o' to global weak symbol 'vtable for spdlog::sinks::base_sink<std::__1::mutex>' from file 'macos/spdlog/x86_64/Release/lib/libspdlog.a(spdlog.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
[...]

PR #3254, opened 2024-11-12
add hidden symbols options

zhaowq32:
#3253


gabime:
No need to duplicate standard cmake flag. Use it directly

PR #3259, opened 2024-11-18
Add explicit `fmt::` and `std::` to avoid ambiguous call

LiAuTraver:
Use three major compilers (clang++ 19, msvc 2022, g++ 14) when `std::format_to` and `fmt::format_to` are both present, the call in
```c++
       return format_to(ctx.out(), "[my_type i={}]", my.i);
```
would be ambiguous. The reason is unknown(even if it's in `fmt::formatter` or `std::formatter`), but it's really easy to fix this. specify `std::` or `fmt::` before `format_to` will solve the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ struct fmt::formatter<my_type> : fmt::formatter<std::string>
 {
     auto format(my_type my, format_context &ctx) const -> decltype(ctx.out())
     {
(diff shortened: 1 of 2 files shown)


LiAuTraver:
Oops, it seems the `f` in `fmt` was eaten by commit message. never mind.😂

gabime:
Thanks @LiAuTraver

PR #3260, opened 2024-11-18
pass /utf-8 only when compiler is MSVC

miyanyan:
this pr is same as https://github.com/fmtlib/fmt/pull/4159

when using clang++ on windows, "/utf-8" is not work.


gabime:
@miyanyan Why the ```if(MSVC)``` on top isn't enough here? Also why the (redundant ?) ```COMPILE_LANGUAGE:CXX```check ?

miyanyan:
> @miyanyan Why the `if(MSVC)` on top isn't enough here? Also why the (redundant ?) `COMPILE_LANGUAGE:CXX`check ?

seems clang will simulate msvc, like define`_MSC_VER`
https://cmake.org/cmake/help/latest/variable/MSVC.html
https://clang.llvm.org/docs/MSVCCompatibility.html
> Set to true when the compiler is some version of Microsoft Visual C++ or another compiler simulating the Visual C++ cl command-line syntax.

gabime:
Megred. Thanks @miyanyan

PR #3269, opened 2024-11-27
Allow manual rotation of rotating_file_sink

hjs-ast:
This PR adds the ability to force a `rotating_file_sink` to rotate its log files on-demand, in addition to the normal size-based rotation.

This is a feature that I implemented as a custom sink for a project where we needed to be able to manually trigger rotation of logs.  That implementation effectively required that the entire `rotating_file_sink` be copy/pasted (as `rotating_file_sink` is marked as `final`, and `rotate_()` is marked as `private`).

This feels like a generally-useful extension to the standard `rotating_file_sink`, that I thought it worth offering upstream.

### DIFF
--- a/include/spdlog/sinks/rotating_file_sink-inl.h
+++ b/include/spdlog/sinks/rotating_file_sink-inl.h
@@ -71,8 +71,7 @@ SPDLOG_INLINE filename_t rotating_file_sink<Mutex>::filename() {
 
 template <typename Mutex>
 SPDLOG_INLINE void rotating_file_sink<Mutex>::rotate_now() {
-    SPDLOG_TRY { rotate_(); }
-    SPDLOG_CATCH_STD
+    rotate_();
 }
 
 template <typename Mutex>
--- a/include/spdlog/sinks/rotating_file_sink-inl.h
+++ b/include/spdlog/sinks/rotating_file_sink-inl.h
@@ -70,7 +70,7 @@ SPDLOG_INLINE filename_t rotating_file_sink<Mutex>::filename() {
 }
 
 template <typename Mutex>
-SPDLOG_INLINE void rotating_file_sink<Mutex>::force_rotation() {
+SPDLOG_INLINE void rotating_file_sink<Mutex>::rotate_now() {
     SPDLOG_TRY { rotate_(); }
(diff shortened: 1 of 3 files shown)


gabime:
Thanks. I think calling it rotate_files () or rotate() or rotate_now() is clearer though.

hjs-ast:
I like `rotate_now()` - I've pushed a change to use that name.  It looks like I also broke the pipeline tests on Windows.  I think I've also fixed that (but haven't been able to test that locally).

include/spdlog/sinks/rotating_file_sink-inl.h
```diff
@@ -69,6 +69,12 @@ SPDLOG_INLINE filename_t rotating_file_sink<Mutex>::filename() {
     return file_helper_.filename();
 }
 
+template <typename Mutex>
+SPDLOG_INLINE void rotating_file_sink<Mutex>::rotate_now() {
+    SPDLOG_TRY { rotate_(); }
+    SPDLOG_CATCH_STD
```
gabime: As a general rule, sinks do not swallow exception. The calling logger catches and handles them, so please remove the try/catch here (where is SPDLOG_CATCH_STD defined?)

gabime (changes requested):

include/spdlog/sinks/rotating_file_sink-inl.h
hjs-ast: Removed, as requested.

`SPDLOG_CATCH_STD` is defined in [common.h:118](https://github.com/gabime/spdlog/blob/15f539685bbb21b3bb01475ac89c6d8d53b5b60f/include/spdlog/common.h#L118).

gabime: Thanks. Then don”t forget to catch it in your code then.

matteodelseppia:
@gabime just a curiosity. Why rotate_now() here does not require the lock? Is there an implicit lock I'm missing somewhere?

gabime:
Good catch. A lock is required indeed

hjs-ast:
I went to fix this, but I see that it's already been done under: https://github.com/gabime/spdlog/pull/3281

Thanks!

PR #3276, opened 2024-11-30
Gabime/fwrite unlocked

gabime:
Use `fwrite_unlocked()`  instead of fwrite() if available (_fwrite_nolock under windows). Improves latency about 5-10ns per call on my linux  machine


jiudianren:
I have a question about this modification.
if complie spdlog to dynamic link  on one machine which support fwrite_unlocked in stdio.h, but  when use this dynamic link on a machine which not support fwrite_unlocked in stdio.h, is there any problem ?

PR #3277, opened 2024-11-30
Add missing xchar.h header for fmt > 11.0.2 (v2)

### LINKED ISSUE #3202: In fmt master, basic_format_string has moved from base.h to xchar.h
Took [place in this commit](https://github.com/fmtlib/fmt/commit/e2b723870787da77c040f5e705882121d618aaef)

This change causes compile time errors due to:-

1. Using SPDLOG_FMT_EXTERNAL and using fmt master
2. spdlog/common needing basic_format_string without SPDLOG_USE_STD_FORMAT https://github.com/gabime/spdlog/blob/e593f6695c6065e6b345fe2862f04a519ed484e0/include/spdlog/common.h#L367-L370 
3. xchar.h not being included due to absence of the specified definitions
https://github.com/gabime/spdlog/blob/e593f6695c6065e6b345fe2862f04a519ed484e0/include/spdlog/common.h#L56-L58

ahoarau:
Better version than https://github.com/gabime/spdlog/pull/3274
Fixes https://github.com/gabime/spdlog/issues/3202

NOTE: I cannot conditionally add the header based on the FMT version, as it's stuck to `#define FMT_VERSION 110002`, even though we're passed that version.

Tested fine on `Ubuntu` + `clang` & `gcc`, on both release and debug.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -53,9 +53,7 @@
     FMT_VERSION >= 80000  // backward compatibility with fmt versions older than 8
     #define SPDLOG_FMT_RUNTIME(format_string) fmt::runtime(format_string)
     #define SPDLOG_FMT_STRING(format_string) FMT_STRING(format_string)
-    #if defined(SPDLOG_WCHAR_FILENAMES) || defined(SPDLOG_WCHAR_TO_UTF8_SUPPORT)
-        #include <spdlog/fmt/xchar.h>
-    #endif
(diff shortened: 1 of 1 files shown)


gabime:
fmt/xchar.h is expensive to include if wchar is not required. Why can't you include it in your code?

tt4g:
#3301 fixes the problem without including `xchar.h`.

PR #3280, opened 2024-12-02
Adding on demand truncation for basic file sinks

matteodelseppia:
Sometimes log rotation can be overkill. There are use cases where it is sufficient to simply truncate without the need for managing rotated logs or archiving. To address this, I have introduced a feature that allows a basic file sink to truncate its log file on demand.

### DIFF
--- a/include/spdlog/sinks/basic_file_sink-inl.h
+++ b/include/spdlog/sinks/basic_file_sink-inl.h
@@ -28,6 +28,7 @@ SPDLOG_INLINE const filename_t &basic_file_sink<Mutex>::filename() const {
 
 template <typename Mutex>
 SPDLOG_INLINE void basic_file_sink<Mutex>::truncate() {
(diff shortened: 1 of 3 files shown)


include/spdlog/sinks/basic_file_sink-inl.h
```diff
@@ -26,6 +26,11 @@ SPDLOG_INLINE const filename_t &basic_file_sink<Mutex>::filename() const {
     return file_helper_.filename();
 }
 
+template <typename Mutex>
+SPDLOG_INLINE void basic_file_sink<Mutex>::truncate() {
+    file_helper_.reopen(true);
```
gabime, line 32: This should be locked using the mutex in the base class. See other sinks for example code.

matteodelseppia: Gosh, sorry for that. Updated.

gabime:
Thanks @matteodelseppia

PR #3281, opened 2024-12-03
Adding lock to rotate_now()

matteodelseppia:
This change fixes a missing lock in the recently introduced rotate_now() of rotating file sinks.


gabime:
Thanks @matteodelseppia

Issue #3295, opened 2024-12-17
[feature request] Ask for customized log filename format


dentiny:
Context, [ray](https://github.com/ray-project/ray/tree/00163c69bb448b5cfd56685b082b26d7cddb609d) has its own log filename logic here to handle concurrently running processes and create unique logging filename: https://github.com/ray-project/ray/blob/00163c69bb448b5cfd56685b082b26d7cddb609d/python/ray/_private/node.py#L828-L830, which unfortunately it perfectly conflicts with spdlog's rotation filename formatting.
So if we enable spglog log rotation and unique filename features at the same time, we cannot tell whether the new file created is from which process.

For the current implementation, spdlog doesn't allow users to customize `calc_filename` function.
https://github.com/gabime/spdlog/blob/24dde318fe034f0c7e809daa5a6e81dc85e1155a/include/spdlog/sinks/rotating_file_sink-inl.h#L52-L64

My feature request and proposal is to allow applications to customize the logging filename format function via a function pointer, or any callable. The callable could set via constructor, or via a function similar to `RotationHandle::SetRotationFnameFormat`.

Example code change:
```
rotating_file_sink(
    filename_t base_filename,
    std::size_t max_size,
    std::size_t max_files,
    bool rotate_on_open,
    const file_event_handlers &event_handlers,
    std::function<std::string(basename, index, extension)>);
```
or 
```
void rotating_file_sink::OverrideRotationLogFilename(std::function<std::string(basename, index, extension)>);
```

Alternatives considered:
- I understand users could always write their customized sink to achieve whatever they want, but here the only thing I ask for is the filename decision function, having a whole class to maintain seems an overkill.

It's similar to previous feature request:
- https://github.com/gabime/spdlog/issues/2612
- https://github.com/gabime/spdlog/issues/2491

If you think the proposal makes sense to you, I'm more than happy to help with the implementation.

tt4g:
I think there is a restriction that the filenames must be generated from the index in order to rotate the log files, which prevents customization of `calc_filename`.
If you have an idea to work around that problem, please send a PR.

dentiny:
@tt4g I made a PR https://github.com/gabime/spdlog/pull/3296, let me know if you have any comments / feedbacks :)

Issue #2612, opened 2023-01-18
how to modify log filename on rotating_file_sink 


diandianzz:
current the rotating_file_sink  general log file with name example.log, example.1.log,example.2.log.......
but I need the file name example_2023-01-18.log,  example_2023-01-18.1.log,  example_2023-01-18.2.log.....
Does this library support this function? What shall I do?

tt4g:
File rotation sinks do not have such a feature.

You can implement a custom sink to add the features you want.
Wiki: https://github.com/gabime/spdlog/wiki/4.-Sinks#implementing-your-own-sink

Issue #2491, opened 2022-09-20
How can I Set/Change the log rename in log rotation?


Candy-Cao:

Candy-Cao:
rotating log rename with index,but i want to replace it with datatime;

tt4g:
> i want to replace it with datatime;

datetime? if so, you can use `spdlog::sinks::daily_file_sink`.
Wiki: https://github.com/gabime/spdlog/wiki/4.-Sinks

jgreber:
But what if you don't want a new file every day? Rotating file by size and then add the current timestamp to the freshly renamed file. That would be a great feature!

tt4g:
@jgreber Timestamp and index rotation are not supported.

See `spdlog::sinks::rotating_file_sink` rotation method.
Rotation is done by calculating the index to be given at the end of the file from `max_files_`.

https://github.com/gabime/spdlog/blob/81de01c02c2b49aa723806b0c1d557dd05721635/include/spdlog/sinks/rotating_file_sink-inl.h#L115-L137

If the timestamp is added to the filename, rotation will not be possible.
This is because the timestamp set in the file will not be known when calculating the index.

PR #3296, opened 2024-12-17
Implement customized rotation filename

dentiny:
Address issues:
- https://github.com/gabime/spdlog/issues/3295
- https://github.com/gabime/spdlog/issues/2612
- https://github.com/gabime/spdlog/issues/2491

This PR allows application to customize their own rotation log filename via setting a customized function.

~~Implementation alternative considered:
Pass the functor as an extra argument for constructor
Since in almost all cases, default rotation naming function is used, I think it's better to have an optional setter function.~~

Update: In the PR, I pass a formatter function from application to decide the output rotation filename.

### DIFF
--- a/include/spdlog/sinks/rotating_file_sink-inl.h
+++ b/include/spdlog/sinks/rotating_file_sink-inl.h
@@ -13,6 +13,7 @@
 #include <spdlog/details/null_mutex.h>
 #include <spdlog/fmt/fmt.h>
 
+#include <cassert>
 #include <cerrno>
 #include <chrono>
 #include <ctime>
@@ -41,12 +42,26 @@ SPDLOG_INLINE rotating_file_sink<Mutex>::rotating_file_sink(
     if (max_files > 200000) {
         throw_spdlog_ex("rotating sink constructor: max_files arg cannot exceed 200000");
     }
-    file_helper_.open(calc_filename(base_filename_, 0));
+    file_helper_.open(get_filename_for_rotation_(base_filename_, 0));
     current_size_ = file_helper_.size();  // expensive. called only once
     if (rotate_on_open && current_size_ > 0) {
         rotate_();
         current_size_ = 0;
     }
+}
+
+template <typename Mutex>
+SPDLOG_INLINE void rotating_file_sink<Mutex>::set_rotate_filename_format(std::function<filename_t(const filename_t &filename, std::size_t index)> rotation_file_format) {
+    assert(!rotation_file_format_);
+    rotation_file_format_ = std::move(rotation_file_format);
+}
+
+template <typename Mutex>
+SPDLOG_INLINE filename_t rotating_file_sink<Mutex>::get_filename_for_rotation_(const filename_t &filename, std::size_t index) {
+    if (rotation_file_format_) {
+      return rotation_file_format_(filename, index);
+    }
+    return calc_filename(filename, index);
 }
 
 // calc filename according to index and file extension if exists.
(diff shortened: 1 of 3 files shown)


jjyao:
@tt4g could you review this PR when you get a chance? Thanks!

tt4g:
I don`t think `set_rotate_filename_format()` is necessary because it causes bugs.
It is not thread-safe and if the filename pattern changes during rotation, the rotation will not work.

`rotation_file_format_` should only be set in the constructor.

include/spdlog/sinks/rotating_file_sink-inl.h
```diff
@@ -41,14 +42,28 @@ SPDLOG_INLINE rotating_file_sink<Mutex>::rotating_file_sink(
...
+    file_helper_.open(get_filename_for_rotation_(base_filename_, 0));
     current_size_ = file_helper_.size();  // expensive. called only once
     if (rotate_on_open && current_size_ > 0) {
         rotate_();
         current_size_ = 0;
     }
 }
 
+template <typename Mutex>
+SPDLOG_INLINE void rotating_file_sink<Mutex>::set_rotate_filename_format(std::function<filename_t(const filename_t &filename, std::size_t index)> rotation_file_format) {
```
tt4g: I don`t think `set_rotate_filename_format()` is necessary because it causes bugs.
It is not thread-safe and if the filename pattern changes during rotation, the rotation will not work.

`rotation_file_format_` should only be set in the constructor.

matteodelseppia:
Are we sure we want to give that much freedom to users that they could break the sink if the lambda is badly written?

include/spdlog/sinks/rotating_file_sink-inl.h
dentiny: Sure, updated.

dentiny:
@tt4g Thanks for the quick review and feedback!

dentiny:
> Are we sure we want to give that much freedom to users that they could break the sink if the lambda is badly written?

The formatter function from application **is supposed to** be read-only and behave similarly as existing `calc_filename`.
If any bug happens, I would think it's application's responsibility to debug and triage themselves.

The most free form for spdlog sink is application completely rewrite its own sink logic.

Would like to know if you have better suggestions :)

gabime:
From my experience with the daily logger, offering such flexibility can lead to unexpected issues (which is why I removed this feature in the v2.x branch).  

A much simpler solution is to create spdlog logs in a separate subfolder to prevent clashes with ray or adopt a similar approach.

dentiny:
> From my experience with the daily logger, offering such flexibility can lead to unexpected issues (which is why I removed this feature in the v2.x branch).

Thanks for the reply! I'm curious what issue did you meet? I would like to see if there's anything we could do to keep the flexibility for filename formatting, meanwhile avoid unintended behaviors.

gabime:
It was hard to clean the old files correctly with custom filenames.

dentiny:
The file deletion is implemented at https://github.com/gabime/spdlog/blob/24dde318fe034f0c7e809daa5a6e81dc85e1155a/include/spdlog/details/os-inl.h#L157-L163

Checking all usage in spdlog:
- https://github.com/gabime/spdlog/blob/24dde318fe034f0c7e809daa5a6e81dc85e1155a/include/spdlog/sinks/rotating_file_sink-inl.h#L121
- https://github.com/gabime/spdlog/blob/24dde318fe034f0c7e809daa5a6e81dc85e1155a/include/spdlog/sinks/hourly_file_sink.h#L81

The file deletions happen at rotation logic, I'm not too worried about it since we're using the same filename decision logic on both file creation and file deletion, would like to know your thoughts. Thank you!

gabime:
Sorry, I prefer not to merge even if this would work. This is a rarely needed feature and can be solved by other means.

PR #3312, opened 2025-01-07
fix: Compatibility with external fmtlib 11.1.1

cblichmann:
Include fmtlib's `xchar` header to include `fmt::basic_format_string`. Otherwise, compilation with an external fmtlib 11.1.1 fails with

```
In file included from include/spdlog/spdlog.h:12:
include/spdlog/common.h:369:49: error: no template named 'basic_format_string' in namespace 'fmt'; did you mean 'std::basic_format_string'?
  369 | inline fmt::basic_string_view<T> to_string_view(fmt::basic_format_string<T, Args...> fmt) {
      |                                                 ^~~~~
```


gabime:
Thanks @cblichmann

PR #3301, opened 2024-12-26
fix: update `to_string_view` function for fmt 11.1

chenrui333:
update `to_string_view` function for fmt 11.1

```
/tmp/spdlog-20241225-25576-12pt7n/spdlog-1.15.0/include/spdlog/common.h:369:54: error: no template named 'basic_format_string' in namespace 'fmt'; did you mean 'basic_format_arg'?
  369 | inline fmt::basic_string_view<Tasic_sfmt::basic_string_view<T> to_string_view(fmt::basic_format_string<T, Args...> fmt) {
      |                                                 ~~~~~^~~~~~~~~~~~~~~~~~~
      |                                                      basic_format_arg
/opt/homebrew/include/fmt/base.h:2428:35: note: 'basic_format_arg' declared here
 2428 | template <typename Context> class basic_format_arg {
      |                                   ^
```

relates to https://github.com/Homebrew/homebrew-core/pull/202442

PR #3314, opened 2025-01-09
fix: remove unused to_string_view overload in fmt >= 11.1

jdrouhard:
This overload was added in #3301 based on a compiler error quickfix suggestion.

fmt's  `fmt::format_string` type [has had an implicit conversion operator](https://github.com/fmtlib/fmt/blob/3f864a4505426ed19e5ba55261af37669a123e4c/include/fmt/base.h#L2738C3-L2738C72) to `fmt::string_view` since fmt 8.0 ([introduced here](https://github.com/fmtlib/fmt/commit/8d70c0edab4b2f68f52ab0644bab8e80dee7a604#diff-f5b9e34d142700358618faa7363cd41bc5662525a82c49f7e6e9dfa2ec8bfc93R2930)).

Since [there is already an overload](https://github.com/gabime/spdlog/blob/7f8060d5b280eac9786f92ac74d263cc8359c5ed/include/spdlog/common.h#L350-353) for `to_string_view()` that no-ops `spdlog::string_view_t` (which is typedef'd to `fmt::string_view` when using fmt), we don't need any other overloads for `format_string`.

### DIFF
--- a/include/spdlog/common.h
+++ b/include/spdlog/common.h
@@ -364,19 +364,7 @@ SPDLOG_CONSTEXPR_FUNC spdlog::wstring_view_t to_string_view(spdlog::wstring_view
 }
 #endif
 
-#ifndef SPDLOG_USE_STD_FORMAT
-#if FMT_VERSION >= 110100
-template <typename T, typename... Args>
-inline fmt::basic_string_view<T> to_string_view(fmt::basic_format_arg<T> fmt) {
-    return fmt;
-}
-#else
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @jdrouhard

Corin-EU:
Thank you so much for this fix as it means that the latest fmt release v11.1.1 (2024-12-27) can be successfully used with latest (at time of compilation) spdlog r.96a7d2a1

PR #3315, opened 2025-01-10
Enabled bin_to_hex utest for stdformat, fixed std::formatter

koniarik:
Turns out it's not hard to fix bin_to_hex to be compatible with std::format - the std::formatter overload just has to be copyable.

Made the fix, and enabled the utest for bin_to_hex in Mac CI for verification that fix works


gabime:
Thanks. Please fix the ci

koniarik:
> Thanks. Please fix the ci

Fixed, had to do another change to test_sink.h as explicit conversion was needed

gabime:
Merged. Thanks @koniarik !

PR #3322, opened 2025-01-17
Compile with `-fvisibility=hidden`.

teo-tsirpanis:
Fixes errors when spdlog is statically linked into two shared libraries, caused by the one copy calling a function of the other. This will not cause any problems for consumers because the exported APIs are already marked with `SPDLOG_API`.


tt4g:
The user's project that is creating a custom sink must be using an API that is not publicly available with `SPDLOG_API`.
I do not think `CMAKE_CXX_VISIBILITY_PRESET=hidden` should be set in the spdlog project.

Instead, you should pass the `-DCMAKE_CXX_VISIBILITY_PRESET=hidden` option to the `cmake` command to achieve your goal.

teo-tsirpanis:
Which API is this, and why not also mark it with `SPDLOG_API`?

tt4g:
Some of the `namespace details` APIs are not publicly available.
However, we assume that some users are using them in custom sinks.

For example `spdlog::details::circular_q`: https://github.com/gabime/spdlog/issues/2774#issuecomment-1611149616

teo-tsirpanis:
`curcular_q` is generic and thus will not be imported from a compiled library, and will not be affected by `-fvisibility=hidden`.

I looked at all headers that have an `-inl.h` counterpart, and the only API declaration that is defined in the `-inl.h` file and has not `SPDLOG_API` is this one: https://github.com/gabime/spdlog/blob/326f8870c2746aea9b295e9b366174116f266b6e/include/spdlog/logger.h#L421

tt4g:
I'm concerned that some users may be using symbols that have not been exported in the constructors of custom sinks and function arguments and return values.
So it is my opinion that we should not `set(CMAKE_CXX_VISIBILITY_PRESET hidden)` in `CMakeLists.txt` of spdlog project.
And users who do not want to expose the spdlog symbols can hide them with `cmake -DCMAKE_CXX_VISIBILITY_PRESET=hidden ` to hide the symbols, so your goal is achieved without any changes.

Finally, this is just my opinion.
@gabime is the one who makes the decision to merge.

gabime:
Agree with @tt4g — who knows what user code this might break. I'll add it to the v2.x branch since it's meant for breaking changes anyway.

gabime:
Merged to v2.x  in PR #3324.  It required some changes to  ansicolor_sink  to make it work, but seems ok now.

PR #3324, opened 2025-01-18
Gabime/visibilty-hidden 2.x

gabime:
Set CMAKE_CXX_VISIBILITY_PRESET and VISIBILITY_INLINES_HIDDEN when build shared lib


CMakeLists.txt
```diff
@@ -242,6 +242,10 @@ if (BUILD_SHARED_LIBS)
     endif ()
     add_library(spdlog SHARED ${VERSION_RC})
     target_compile_definitions(spdlog PUBLIC SPDLOG_SHARED_LIB)
+    set_target_properties(spdlog PROPERTIES
+            CXX_VISIBILITY_PRESET hidden
+            VISIBILITY_INLINES_HIDDEN ON)
```
teo-tsirpanis, line 247: Can this be set always, even when building a static library? See awslabs/aws-c-common#1142 for reasons why.

gabime: Sure, If there is a good reason.

PR #3327, opened 2025-01-23
Support custom environment variables for load_env_levels

ken-matsui:
SPDLOG_LEVEL is currently supported to control log levels via load_env_levels.

This patch adds support for other environment variable names, such as MYAPP_LEVEL, for load_env_levels.

### DIFF
--- a/tests/test_cfg.cpp
+++ b/tests/test_cfg.cpp
@@ -19,6 +19,15 @@ TEST_CASE("env", "[cfg]") {
 #endif
     load_env_levels();
     REQUIRE(l1->level() == spdlog::level::warn);
+
+#ifdef CATCH_PLATFORM_WINDOWS
+    _putenv_s("MYAPP_LEVEL", "l1=trace");
+#else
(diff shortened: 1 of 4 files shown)


tests/test_cfg.cpp
```diff
@@ -19,6 +19,15 @@ TEST_CASE("env", "[cfg]") {
 #endif
     load_env_levels();
     REQUIRE(l1->level() == spdlog::level::warn);
+
+#ifdef CATCH_PLATFORM_WINDOWS
```
gabime, line 23: Please remove this CATCH_PLATFORM_WINDOWS ifdef

gabime (changes requested):

tests/test_cfg.cpp
ken-matsui: Would you mean we can simply use setenv on Windows as well?

gabime: No, its fine. my mistake

gabime:
Thanks @ken-matsui

PR #3356, opened 2025-03-08
Support using a priority inheritance protocol

jdknight:
Provides initial support for utilizing a priority inheritance protocol for mutexes if supported by the threading library. Developers can opt to build with `SPDLOG_PRIORITY_INHERITANCE` enabled to help avoid priority inversion scenarios.

----

The following change set is being proposed to help the spdlog framework to be better equip for (near) real-time applications. I was asked about a suggested logging framework to use for a C++ application and recommended spdlog; however, I was told by others that they could not use it since the risk of priority inversion scenarios with the use of the standard library's mutex implementations. This spawned this change set by adding support for priority inheritance for environments that support such a capability. At this time, this only brings support for Linux gcc/clang environments (as least, that is all that has been tested with).

If such a change is not desired in the logging framework, no worries.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -22,12 +22,14 @@ jobs:
           - { compiler: gcc, version: 9, build_type: Release, cppstd: 17 }
           - { compiler: gcc, version: 11, build_type: Debug, cppstd: 20 }
           - { compiler: gcc, version: 12, build_type: Release, cppstd: 20 }
+          - { compiler: gcc, version: 12, build_type: Release, cppstd: 20, pi: ON }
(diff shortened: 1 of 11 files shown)


gabime:
Thanks. Indeed, I am not sure if I want to support this. In any case the change can be less intrusive. Just define a new kind of mutex (see null_mutex.h for example) and let users pass  it as template param when creating the sinks. This would cover most of use cases, and for the rest (the queue and registry), a `using mutex=pi_mutex` could be used with ifdef. So I think there is no need for a wrapper class with mutex in it.

jdknight:
K k. Thanks for the feedback; will close.

PR #3357, opened 2025-03-09
Add Swift Package Manager support

wwallace-exo:


gabime:
Thanks. However we don't support code for package managers here. only usage details

PR #3359, opened 2025-03-13
Update CMake minimum version required

jan-moravec:
FILE_SET functionality was introduced in CMake version 3.23.

https://cmake.org/cmake/help/latest/command/target_sources.html


gabime:
Thanks @jan-moravec

PR #3378, opened 2025-04-10
Fix links to local reference.

cgringmuth:
Enables local navigation in IDE and removes hardcoded link to branch `tree/v1.x`  which requires being updated every time you switch to other default branch. :grin: 

PS: Thx for merging last PR this fast. :+1:


gabime:
Thanks @cgringmuth

PR #3381, opened 2025-04-14
Fix links for spdlog/wiki

non-descriptive:


gabime:
Thanks @non-descriptive

Issue #3188, opened 2024-09-15
log level of messages of the dup_filter_sink


yalov:
Hello, as of now I have:

```
[2024-09-15_13:58:06.955] [65.178] [TRCE] message
[2024-09-15_13:58:07.040] [65.262] [INFO] Skipped 1 duplicate messages..
```

The message that is duplicated is at the TRACE level, but the "duplication message" is at the INFO level.
Is this expected?
Is there a configuration to make the "duplication message" use the same level as the message being duplicated (which may vary)?

tt4g:
> Is there a configuration to make the "duplication message" use the same level as the message being duplicated (which may vary)?

This feature is not supported.

Skip messages are notified at the log level configured in the constructor.

https://github.com/gabime/spdlog/blob/ffd5aa41d687ec1d9b2d6ff48ca1c816f7c7f0c5/include/spdlog/sinks/dup_filter_sink.h#L43-L46

Also there is one problem if you are achieving the behavior you expect.
dup_filter_sink detects duplicate log messages, but does not check the log level.
If dup_filter_sink skips the same message with different log levels, what should the log level of message “Skipped N duplicate message.” be?

Note, I am not the author of dup_filter_sink.

yalov:
> If dup_filter_sink skips the same message with different log levels, what should the log level of message “Skipped N duplicate message.” be?

The maximum log level of the skipped messages, so the "Skipped N duplicate" message will not be skipped, and will not have the higher log level than the skipped messages themselves.

tt4g:
I know what the desired behavior is.
However, if the log level that generates dup_filter_sink log message “Skipped N duplicate.” is changed dynamically, the behavior will change.
This would be an undesirable change for users who have been discarding “Skipped N duplicate.” messages by setting the trace level of dup_filter_sink.

spdlog allows you to define custom sinks, so create your own dup_filter_sink that behaves the way you want: https://github.com/gabime/spdlog/wiki/4.-Sinks#implementing-your-own-sink

yalov:
ok, I will to that.

PR #3389, opened 2025-04-18
dup_filter_sink: Add option to use log level of duplicated messages for notification message

SleepingSoul:
In current implementation of dup_filter_sink, we set it up with some log level (info by default) and all notification messages use this log level.
There are some problems with that approach:
For example, I have file sink that has log level info (it ignores debug messages). When we get debug duplicates - it will generate [info] notification about it, and this notification will get to my file, even though it is information about debug messages that I would like to discard. I have added option to make it "dynamic" and fix this problem. So, for debug duplicates I will get debug notification, for error duplicates I will get error notification, and so on.

To not break any existing code, I have added third argument to the constructor, and it is false by default. So, behaviour will remain same unless third argument will be explicitly specified. I understand that second argument does not have any sense if we specify "use_message_level_for_notification" as true, however, this is the only backward-compatible approach I could make up.

### DIFF
--- a/include/spdlog/sinks/dup_filter_sink.h
+++ b/include/spdlog/sinks/dup_filter_sink.h
@@ -45,7 +45,7 @@ public:
                              bool use_message_level_for_notification = false)
         : max_skip_duration_{max_skip_duration},
           log_level_{notification_level},
-          use_message_level_for_notification_(use_message_level_for_notification) {}
+          use_message_level_for_notification_{use_message_level_for_notification} {}
 
 protected:
     std::chrono::microseconds max_skip_duration_;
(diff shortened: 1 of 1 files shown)


SleepingSoul:
Requires further testing

gabime:
Related: #3188

SleepingSoul:
Tested locally and fixed issue with wrong log level

gabime:
Thanks @SleepingSoul . 
The original behavior is confusing and likely a bug. Adding yet another parameter would only add to the confusion. 
Better to always use the last message’s level and remove the `notification_level` param  from the constructor (minor API break, but justified since it fixes a bug). 

PR is  welcome.

SleepingSoul:
@gabime got it, will create new PR

PR #3390, opened 2025-04-18
dup_filter_sink: Fix incorrect behaviour when producing notifications about skipped messages

SleepingSoul:
Issue: https://github.com/gabime/spdlog/issues/3188
Previous PR (with discussion): https://github.com/gabime/spdlog/pull/3389
As discussed, current behaviour is considered as bugged.

- dup_filter_sink no longer has "notification_level" argument
- dup_filter_sink always uses log level of skipped messages when producing notification messages

### DIFF
--- a/include/spdlog/sinks/dup_filter_sink.h
+++ b/include/spdlog/sinks/dup_filter_sink.h
@@ -40,22 +40,21 @@ template <typename Mutex>
 class dup_filter_sink : public dist_sink<Mutex> {
 public:
     template <class Rep, class Period>
-    explicit dup_filter_sink(std::chrono::duration<Rep, Period> max_skip_duration,
-                             level::level_enum notification_level = level::info)
-        : max_skip_duration_{max_skip_duration},
-          log_level_{notification_level} {}
+    explicit dup_filter_sink(std::chrono::duration<Rep, Period> max_skip_duration)
+        : max_skip_duration_{max_skip_duration} {}
 
 protected:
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @SleepingSoul

yalov:
Nice to see it was implemented after the https://github.com/gabime/spdlog/issues/3188

Two questions:

1. Why it is the last level, and not the maximum level ?
```
if (msg.level > skipped_msg_log_level_)
    skipped_msg_log_level_ = msg.level;
```

2. Locally on my own version of the same sink, I was having a bug of wrong level of the skipped message, I can't remember specifics, but I had fixed it by

```
        // ....
        // log the "skipped.." message
        if (skip_counter_ > 0) {
            char buf[64];
            auto msg_size = ::snprintf(buf, sizeof(buf), "Skipped %u duplicate messages..",
                                       static_cast<unsigned>(skip_counter_));
            if (msg_size > 0 && static_cast<size_t>(msg_size) < sizeof(buf)) {
                details::log_msg skipped_msg{msg.source, msg.logger_name, skipped_msg_log_level_,
                                             string_view_t{buf, static_cast<size_t>(msg_size)}};
                dist_sink<Mutex>::sink_it_(skipped_msg);
            }

            log_level_ = level::level_enum::off;
            skip_counter_ = 0;
        }
```



**UPD.** 
I got it. changes in 2 is necessary for diff in 1

PR #3393, opened 2025-04-23
Fix warning C4530

roeas:
C4530 is because when using `/EHs-c-`, std::chrono and std::string stuff still attempts to use try-catch. Defining `_HAS_EXCEPTIONS=0` can prevent MSVC STL from using try-catch code, bundled fmt is also affected by this：
fmt/bundled/base.h
image


gabime:
Seems that fmt renamed `FMT_EXCEPTIONS` to `FMT_USE_EXCEPTIONS` in version [11.1.0](url).

Please also change 
``` target_compile_definitions(spdlog PUBLIC FMT_EXCEPTIONS=0)```
to 
``` target_compile_definitions(spdlog PUBLIC FMT_USE_EXCEPTIONS=0)```

roeas:
Oh, right

gabime:
Thanks @roeas

Issue #3414, opened 2025-06-06
Too short environment variables are allowed


kfeng123:
https://github.com/gabime/spdlog/blob/b18a234ed6af38638678a3338c0d7ed90210ae6c/include/spdlog/details/os-inl.h#L566-L580

In the wrapped function `getenv`, for windows, it seems the environment variable length is restricted to within 128 characters, which may be easily exceeded if there are multiple loggers.

Why not just use std::getenv for windows?

kfeng123:
fixed in #3415

PR #3415, opened 2025-06-06
Use std::getenv #3414

kfeng123:


include/spdlog/details/os-inl.h
```diff
@@ -563,21 +563,21 @@ SPDLOG_INLINE filename_t dir_name(const filename_t &path) {
     return pos != filename_t::npos ? path.substr(0, pos) : filename_t{};
 }
 
+#ifdef _MSC_VER
+    #pragma warning(push)
+    #pragma warning(disable : 4996)
+#endif  // _MSC_VER
```
gabime, line 569: why disable warning 4996?

include/spdlog/cfg/helpers-inl.h
```diff
@@ -71,7 +71,7 @@ inline std::unordered_map<std::string, std::string> extract_key_vals_(const std:
 }
 
 SPDLOG_INLINE void load_levels(const std::string &input) {
-    if (input.empty() || input.size() > 512) {
+    if (input.empty()) {
```
gabime: i prefer to have some sane limit of size to be on the safe side. just put some bigger number e.g. 2048

flagarde: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getenvironmentvariable

`
An environment variable has a maximum size limit of 32,767 characters, including the null-terminating character.
`

gabime: ok, lets limit to 32k

kfeng123: done

include/spdlog/details/os-inl.h
kfeng123: CI shows that warning 4996 may be produced by std::getenv for MSVC.

gabime:
Thanks @kfeng123

PR #3428, opened 2025-06-27
compatible for c++11 but GCC < 5

aengusjiang:
a little change to support c++11 but GCC < 5, such as centos7.9 gcc4.8.5


tt4g:
`include/spdlog/fmt/bundled/format.h` is copied from the fmt repository.
If merged in the fmt repository, the changes will be imported into spdlog after the fmt release.

EDIT: This PR is open in fmt repository: fmtlib/fmt#4482

aengusjiang:
Ok， thanks the testcase to improve this change.
I will close.

PR #3430, opened 2025-06-29
Remove unnecessary and inconsistent "final" from color sinks

### LINKED ISSUE #3429: `wincolor_sink::log()` shouldn't be `final`
Please remove `final` from `wincolor_sink::log()` (and possibly other member functions such as `flush()`).

Rationale: consider the code like this

```cpp
class ConsoleSink : public spdlog::sinks::stderr_color_sink_mt
{
public:
  ConsoleSink()
  {
    // Show just the message text in the console output, we don't want any extra
    // details like time, message level etc in it.
    set_pattern("%v");
  }

  void log(const spdlog::details::log_msg &msg) override
  {
    // Do something before logging [...]

    // Call the base class implementation to do the actual logging.
    spdlog::sinks::stderr_color_sink_mt::log(msg);

    // Possibly do something else after logging [...]
  }
};
```

It compiles and works fine under Unix when `stdout_color_sink_mt` is a `ansicolor_sink_mt`, but fails to compile under Windows when it is `wincolor_stderr_sink_mt`. IMO there is no reason for `log()` to be final in the latter class, especially if it's not final in the former.

As it stands, this has to be worked around by defining a delegating logger, but this is just unnecessary extra work which could be avoided and comes as a nasty surprise when porting from Unix to Windows.

I can make the trivial PR doing this if the maintainers agree to do it, of course, please let me know if I should.

vadz:
The use of "final" differed between ansicolor_sink and wincolor_sink, resulting in the code inheriting from std{err,out}_color_sink classes, which are defined as one or the other on different platforms, being able to override most of the functions under non-Windows platforms, but not under Windows.

This seems gratuitously inconsistent, so just remove all "final" keywords from both classes, especially because there doesn't seem any good reason to use it and the other sink classes don't use it (with the exception of base_sink, which is special).

This also incidentally fixes using "final override" in most places but "override final" in wincolor_sink.h.

Fixes #3429.

### DIFF
--- a/include/spdlog/sinks/ansicolor_sink.h
+++ b/include/spdlog/sinks/ansicolor_sink.h
@@ -40,7 +40,7 @@ public:
 
     void log(const details::log_msg &msg) override;
     void flush() override;
-    void set_pattern(const std::string &pattern) final override;
+    void set_pattern(const std::string &pattern) override;
     void set_formatter(std::unique_ptr<spdlog::formatter> sink_formatter) override;
 
     // Formatting codes
--- a/include/spdlog/sinks/wincolor_sink.h
+++ b/include/spdlog/sinks/wincolor_sink.h
@@ -31,10 +31,10 @@ public:
 
     // change the color for the given level
     void set_color(level::level_enum level, std::uint16_t color);
-    void log(const details::log_msg &msg) final override;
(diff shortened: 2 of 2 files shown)


gabime:
Thanks @vadz

PR #3433, opened 2025-06-30
chore(cmake): add option to override CMAKE_DEBUG_POSTFIX

joshuachp:
This will make it possible to use the pkg-config with CMake debug build.


gabime:
Thanks @joshuachp

PR #3437, opened 2025-07-08
Update README.md

yalov:
This pull request updates the `README.md` file and introduces a new example for managing multiple loggers with different formats and levels.

The example showcasing 2 loggers and `spdlog::set_level()`, which set level NOT only to default logger, but to ALL registered loggers

(I won’t say how many hours it took me to figure this out)

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,8 +246,8 @@ void multi_sink_example()
 ```c++
 
 // Creation of loggers. Registration of loggers.
-// Setting default logger. Setting global level to all registered loggers. 
-void multi_sink_example()
+// Setting a default logger. Setting a global level to all registered loggers. 
+void multi_loggers_example()
 {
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. If the intention is to clarify set_level, this could be much shorter. The example is little big for the frontage.  Please try to shorten it.
Another option is to add it to the wiki.

yalov:
I made it a little bit sorter.

I've kept the two loggers in the example to demonstrate that set_level applies globally to all registered loggers.

gabime:
Still too long ,sorry. For example “replace_default_logger_example()” already explains about default loggers. no  need to repeat this here. This pr should take very few lines to convey the set level stuff.

yalov:
ok, I made it more short

Issue #3138, opened 2024-07-17
ANSI color sink on Windows


cschreib-ibex:
Most modern terminals on Windows now support ANSI color codes (this includes, e.g., Windows Terminal, and the good old console if created with the appropriate flags; see `ENABLE_VIRTUAL_TERMINAL_PROCESSING`, starting from Windows 10). ANSI color codes have the advantage of being able to survive when stdout is piped, while Windows console color commands do not (in fact, the output is just lost if the colours are set to `always`).

Unfortunately:
 1. When using spdlog built as a shared or static library, the `ansicolor_sink-inl.h` is never included in the library, therefore simply trying to include `ansicolor_sink.h` and using the class results in linker errors. This can be worked around by including `ansicolor_sink-inl.h` explicitly.
 2. The ANSI color sink uses `::fwrite` unconditionally, which results in double carriage returns being printed on Windows. This is the same issue as https://github.com/gabime/spdlog/issues/1675, which applied to the regular stdout sink, but now applies to the ANSI sink. The solution would be to use the same fix here, namely, use `::WriteFile` on Windows and `::fwrite` on linux.

PR #3442, opened 2025-07-18
Fix extra carriage returns in piped ANSI output on Windows

LowLevelLore:
## Description

This PR fixes incorrect line endings in ANSI-colored output when redirected or piped on Windows terminals. A possible fix for #3138.

### Problem

When using `fwrite()` on Windows, the CRT applies text-mode translation, converting `\n` to `\r\n`. This results in an extra carriage return (`\r`) being injected, especially noticeable when ANSI-colored output is redirected or piped.

### Solution

On Windows, replace `fwrite()` with a direct call to `WriteFile()` using the native file handle obtained from `_get_osfhandle(_fileno(...))`. This bypasses the CRT entirely, ensuring output is written exactly as intended, without automatic newline translation.

### Code Change

```cpp
#ifdef _WIN32
    DWORD bytes_written = 0;
    HANDLE h = reinterpret_cast<HANDLE>(_get_osfhandle(_fileno(target_file_)));
    // WriteFile bypasses the extra \r injection
    WriteFile(h, color_code.data(), static_cast<DWORD>(color_code.size()), &bytes_written, nullptr);
#else
    // existing fwrite() path
#endif
```

### CMake
Also included ansicolor_sink-inl.h explicitly in the build to avoid linking issues (when linking to static library).


LowLevelLore:
Please ignore previous commits.

LowLevelLore:
@gabime any updates / recommendations on this ?

LowLevelLore:
@gabime Any progress here ?

gabime:
The ansicolor sink was never designed for windows. Why not use the wincolor sink

flagarde:
New terminal on windows can be set to understand ANSI escape code. https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences

CMakeLists.txt
```diff
@@ -156,6 +156,7 @@ message(STATUS "Build type: " ${CMAKE_BUILD_TYPE})
 # Static/Shared library
 # ---------------------------------------------------------------------------------------
 set(SPDLOG_SRCS src/spdlog.cpp src/stdout_sinks.cpp src/color_sinks.cpp src/file_sinks.cpp src/async.cpp src/cfg.cpp)
+list(APPEND SPDLOG_SRCS include/spdlog/sinks/ansicolor_sink-inl.h)
```
gabime, line 159: Should be included in https://github.com/gabime/spdlog/blob/v1.x/src/stdout_sinks.cpp instead

include/spdlog/sinks/ansicolor_sink-inl.h
```diff
@@ -113,14 +117,28 @@ SPDLOG_INLINE void ansicolor_sink<ConsoleMutex>::set_color_mode_(color_mode mode
...
 }
 
 template <typename ConsoleMutex>
 SPDLOG_INLINE void ansicolor_sink<ConsoleMutex>::print_range_(const memory_buf_t &formatted,
                                                               size_t start,
                                                               size_t end) const {
+#ifdef _WIN32
+    DWORD bytes_written = 0;
+    HANDLE h = reinterpret_cast<HANDLE>(_get_osfhandle(_fileno(target_file_)));
+    WriteFile(h, formatted.data() + start, static_cast<DWORD>(end - start), &bytes_written,
```
gabime, line 137: _get_osfhandle shoild be called only once at construction time and store the result in a class member

include/spdlog/sinks/ansicolor_sink-inl.h
```diff
@@ -10,6 +10,10 @@
 #include <spdlog/details/os.h>
 #include <spdlog/pattern_formatter.h>
 
+#if defined(_WIN32)
```
gabime, line 13: please use #ifdef _WIN32

gabime (changes requested):

PR #3444, opened 2025-07-20
Remove the fileapi.h include in os-inl.h

MasonAngelio:
As instructed in https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-flushfilebuffers

The reason for this change is that it causes compilation errors for older SDKs (such as 7.1A)

```cpp
fatal error C1083: Cannot open include file: 'fileapi.h': No such file or directory
```

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -23,7 +23,6 @@
 
 #ifdef _WIN32
     #include <spdlog/details/windows_include.h>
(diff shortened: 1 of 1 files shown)


gabime:
windows_include.h already includes windows.h. So i think just remove the fileapi.h include

gabime:
Thanks @MasonAngelio

MasonAngelio:
Appreciate that. Best of luck

PR #3472, opened 2025-10-10
file_helper support mmap

seker:
'file_helper' support support mmap.
in some platform, that will be faster then file I/O.


gabime:
Thanks.  Do you have any numbers on the improvement gains?

PR #3473, opened 2025-10-10
update to fmt 12.0.0

alexander-riss:
this comes with a significant performance uplift for writing floating point numbers of about 60%


gabime:
Thanks. I updated  it already  in commit 4418909af29e3b328084560345807dbda38de3eb

PR #3474, opened 2025-10-14
Emscripten logging sink

DisabledMallis:
I like using spdlog for my projects, and recently I've been building a web app using Emscripten, so I implemented a sink that uses the emscripten_log function. I whipped this up haphazardly for my own purposes, so if I need to add or change anything LMK.


gabime:
Thanks. Not sure the demand for this. So iI think i will close the for now.

Note:
Much more efficient to do formatted.push_back('\0') to convert to c string  instead of calling fmt:to_string().c_str()

PR #3479, opened 2025-10-27
fix sign-compare warning

AleksandrSokolkin:
I'm using spdlog in my project and I have to set option `-Wno-sign-compare` for compilation with `-Werror`. Because of that I can miss an incorrect comparison in my code.


gabime:
Thanks @AleksandrSokolkin

Issue #345, opened 2017-01-05
write error and critical logs to stderr


Tmplt:
In a project I work on I use spdlog to report errors to the user. To match the behavior of other cli-programs, I want that output written to `stderr`. Could `level::{err,critical}` logging be changed to write to `stderr`, or could a define perhaps be exposed in `tweakme.h`?

Cheers.

gabime:
You can add an additional stderr sink with its level to level::error.
This way your logger would have 2 sinks with different levels.

Tmplt:
The wiki seems to lack the level-setting part, unless I missed it completely. Could you show how this would be done?

gabime:
Just call set_level(..) on any sink:
```c++
sink->set_level(spdlog::level::err);
```

Tmplt:
Ah, thanks.

I've come thus far:
```cpp
auto setup_logger()
{
    std::vector<spdlog::sink_ptr> sinks;

    auto stdout_sink = spdlog::sinks::stdout_sink_mt::instance();

    auto stderr_sink = spdlog::sinks::stderr_sink_mt::instance();
    auto color_stderr_sink = std::make_shared<spdlog::sinks::ansicolor_sink>(stderr_sink);
    color_stderr_sink->set_level(spdlog::level::err);

    sinks.push_back(stdout_sink);
    sinks.push_back(color_stderr_sink);

    auto logger = std::make_shared<spdlog::logger>("main", std::begin(sinks), std::end(sinks));
    logger->set_pattern("%l: %v");

    return logger;
}
```

As wanted, error logs are now printed to `stderr`, but they are also printed to `stdout`. Can I assign `stdout_sink` to print everything except for error and critical logs, which `color_stderr_sink` handles?

Tmplt:
To answer my own questions:

`logger->set_level(spdlog::level::info);` does affect both sinks.

And I gather from [`details/logger_impl.h`](https://github.com/gabime/spdlog/blob/master/include/spdlog/details/logger_impl.h#L64) that it is not possible to assign a singular level or range of levels to a sink: `stdout_sink` prints `err`-logs since `level::err`>`level::info`.

gabime:
Yes, you got it right, but note that if for example the logger has "info" level then debug messages won't get forwarded by the logger to the underlying sinks, regardless of their own level. 

[...]

PR #3486, opened 2025-11-09
Change access scope for ANSI target_file_ from private to protected

fab4100:
This change allows for a custom minimal ANSI color sink implementation that supports, for example, splitting between `stdout` and `stderr` file streams depending on the log level.  An example application specific custom sink that realizes this behavior would look like:

```cpp
template <typename ConsoleMutex>
class SplitSink : public sinks::ansicolor_sink<ConsoleMutex>
{
    using Base = sinks::ansicolor_sink<ConsoleMutex>;

public:
    SplitSink(color_mode mode = color_mode::automatic) : Base(stdout, mode) {}

    void log(const details::log_msg &msg) override
    {
        if (msg.level <= SPDLOG_LEVEL_WARN) {
            this->target_file_ = stdout;
        } else {
            this->target_file_ = stderr;
        }
        Base::log(msg);
    }
};
```

Inspired by https://github.com/gabime/spdlog/issues/345 and https://github.com/eic/EICrecon/issues/456.  This commit aims at reusing all of the `ansicolor_sink` code with the exception of dynamic target file selection that can be implemented in application code based on example above.

### DIFF
--- a/include/spdlog/sinks/ansicolor_sink.h
+++ b/include/spdlog/sinks/ansicolor_sink.h
@@ -78,8 +78,10 @@ public:
     const string_view_t red_bold = "\033[31m\033[1m";
     const string_view_t bold_on_red = "\033[1m\033[41m";
 
-private:
+protected:
     FILE *target_file_;
+
+private:
     mutex_t &mutex_;
     bool should_do_colors_;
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. Note that your SplitSink is no thread safe.

PR #3489, opened 2025-11-15
Fix UWP detection.

### LINKED ISSUE #3488: Detection of UWP platform
Hi,

This line:
https://github.com/gabime/spdlog/blob/cdbd64e2305a712ec9a360e5a1e05aa1dcab85e7/include/spdlog/details/os-inl.h#L570
is causing problems for us when compiling spdlog 1.1.6.0 for UWP without enabling /ZW. According to [this documentation](https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/migrate-to-windows-app-sdk/misc-info), it seems that detection of UWP platform should be done via the `WINAPI_FAMILY` macro:
```c++
#if (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP)
    // Win32/Desktop code, including Windows App SDK code
#else
    // UWP code
#endif
```

jdumas:
Fixes #3488

### DIFF
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -567,7 +567,7 @@ SPDLOG_INLINE filename_t dir_name(const filename_t &path) {
     #pragma warning(disable : 4996)
 #endif  // _MSC_VER
 std::string SPDLOG_INLINE getenv(const char *field) {
-#if defined(_MSC_VER) && defined(__cplusplus_winrt)
+#if defined(_MSC_VER) && (WINAPI_FAMILY != WINAPI_FAMILY_DESKTOP_APP)
     return std::string{};  // not supported under uwp
 #else
     char *buf = std::getenv(field);
--- a/include/spdlog/details/os-inl.h
+++ b/include/spdlog/details/os-inl.h
@@ -567,7 +567,8 @@ SPDLOG_INLINE filename_t dir_name(const filename_t &path) {
     #pragma warning(disable : 4996)
 #endif  // _MSC_VER
 std::string SPDLOG_INLINE getenv(const char *field) {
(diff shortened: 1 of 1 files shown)


include/spdlog/details/os-inl.h
```diff
@@ -567,7 +567,7 @@ SPDLOG_INLINE filename_t dir_name(const filename_t &path) {
     #pragma warning(disable : 4996)
 #endif  // _MSC_VER
 std::string SPDLOG_INLINE getenv(const char *field) {
-#if defined(_MSC_VER) && defined(__cplusplus_winrt)
+#if defined(_MSC_VER) && (WINAPI_FAMILY != WINAPI_FAMILY_DESKTOP_APP)
```
gabime: Better is check if WINAPI_FAMILY defined:

#if defined(_MSC_VER) && defined(WINAPI_FAMILY) && (WINAPI_FAMILY != WINAPI_FAMILY_DESKTOP_APP)
// non-desktop MSVC code
#endif

gabime (changes requested):

include/spdlog/details/os-inl.h
jdumas: Hmm should we also check if WINAPI_FAMILY_DESKTOP_APP is defined then?

gabime: Not sure. What the docs say?

jdumas: Unclear. I cannot find any official doc aside from [this page](https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/migrate-to-windows-app-sdk/misc-info). That being said, it's unlikely one macro would be defined but not the other.

gabime: I guess checking them both is the safe thing

jdumas: Condition updated!

gabime:
Thanks @jdumas

PR #3504, opened 2025-12-17
Docs: fix misleading comment in blocking_queue header

zanewong233:
The header comment for enqueue_nowait in blocking_queue.h incorrectly stated that it returns false if the queue is full.

However, the actual implementation uses circular_q which overwrites the oldest item when full, and the function returns void.


gabime:
Thanks @zanewong233

PR #3516, opened 2026-01-10
Rename local variables to avoid shadowing member functions

KaganCanSit:
Hello,

This PR addresses CppCheck `shadowFunction` style warnings by renaming local variables that shadow member functions. These warnings can often be ignored, as modern compilers handle this well, but I believe it's good to strengthen the fundamentals.

CppCheck style warning messages addressed:
```bash
Shadow Function
-------------------------------------
include/spdlog/logger-inl.h:167:style:shadowFunction -> Local variable 'flush_level' shadows outer function
include/spdlog/sinks/daily_file_sink.h:110:style:shadowFunction -> Local variable 'filename' shadows outer function
include/spdlog/sinks/daily_file_sink.h:134:style:shadowFunction -> Local variable 'filename' shadows outer function
```

- `should_flush_()`: The local variable `flush_level` was directly modified by a call to the member function `flush_level()`. This returns the appropriately converted `level::level_enum` return type instead of the raw `int` from the atomic load (prevents implicit type conversion). 

   **However, it would be good to verify this change for atomic. I am not sure this!**

- I updated the naming of local variables with the `new_` prefix for the `filename()` function declaration in the `daily_file_sink.h` file. 

   For consistency, I also adjusted the constructor method for compatibility. This shadowing warning is not given because the `filename()` member function is not yet accessible in the constructor body (the object has not been fully created).

Best regards.

### DIFF
--- a/include/spdlog/logger-inl.h
+++ b/include/spdlog/logger-inl.h
@@ -164,8 +164,7 @@ SPDLOG_INLINE void logger::dump_backtrace_() {
 }
 
 SPDLOG_INLINE bool logger::should_flush_(const details::log_msg &msg) const {
-    auto flush_level = flush_level_.load(std::memory_order_relaxed);
-    return (msg.level >= flush_level) && (msg.level != level::off);
+    return (msg.level >= flush_level()) && (msg.level != level::off);
 }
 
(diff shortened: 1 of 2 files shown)


gabime:
Thanks @KaganCanSit

PR #3509, opened 2026-01-04
Replace C-style cast with reinterpret_cast in udp_client

KaganCanSit:
Hello,

I'm not fully familiar with the `spdlog` project, but I wanted to take a quick look by performing a `cppcheck`. Besides the `fmt` library included in the bundle and the `unused` function warnings, I obtained the list at the end of this PR message. If you have anything you'd like to address, please let me know. I'd be happy to help.

For now, to avoid confusion, I've addressed the following two points:
```bash
include/spdlog/details/udp_client.h:74:style:unreadVariable -> Variable 'toslen' is assigned a value that is never used.
include/spdlog/details/udp_client.h:74:style:cstyleCast -> C-style pointer casting
```

Accordingly, I replaced the C-style pointer conversion with `reinterpret_cast<const sockaddr*>` in the `send()` method for better type safety, and removed the unused `toslen` variable in the `send()` method.

I formatted and tested it with the following calls, **but I think we will need additional testing. I can do it if you guide me.**

### Compilation & Testing
```bash
./scripts/format.sh
mkdir build && cd build
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DSPDLOG_BUILD_TESTS=ON ..
cmake --build .
tests/spdlog-utests
```

### CppCheck Command
```bash
cppcheck --project=compile_commands.json --enable=all --inconclusive --platform=unix64 --check-level=exhaustive --suppress=missingIncludeSystem --template="{file}:{line}:{severity}:{id} -> {message}" --output-file=cppcheck_report_spdlog.txt
```

Regards

---

### Full cppcheck Output:
```bash
spdlog/include/spdlog/details/null_mutex.h:19:warning:uninitMemberVar -> Member variable 'null_atomic_int::value' is not initialized in the constructor.
spdlog/include/spdlog/details/backtracer-inl.h:11:warning:missingMemberCopy -> Member variable 'backtracer::mutex_' is not assigned in the copy constructor. Should it be copied?
spdlog/include/spdlog/details/backtracer-inl.h:17:warning:missingMemberCopy -> Member variable 'backtracer::mutex_' is not assigned in the move constructor. Should it be moved?
spdlog/include/spdlog/details/registry-inl.h:44:performance:useInitializationList -> Variable 'default_logger_' is assigned in constructor body. Consider performing initialization in initialization list.
[...]

PR #3521, opened 2026-01-15
Fix shadow member warning in example file

KaganCanSit:
Hello,

This PR resolves the following compiler warning occurring in the `example.cpp` file.

```bash
spdlog/example/example.cpp: In constructor 'my_type::my_type(int)':
spdlog/example/example.cpp:271:26: error: declaration of 'i' shadows a member of 'my_type' [-Werror=shadow]
271 | explicit my_type(int i)
| ~~~~^
```

I thought using a different name instead of `i` would be more understandable. Additionally, as you previously mentioned in another PR request, I added an underscore `_` to the end of the member variable, in line with the project approach.

If any other changes are needed, please leave a comment.

---

**Additional Note:** What are your thoughts on additional flags compatible with older compiler versions (GCC 4.8+, Clang 3.3+)? The `-Wold-style-cast` flag came to mind during development #3509. Enabling such flags can help contributors detect these types of errors early on.

I cannot run MSVC tests because I don't have a Windows environment. I can add additional flags to the `utils.cmake` file for GCC/Clang. (`-Wold-style-cast`, `-Wshadow` maybe)

Best Regards.

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -267,16 +267,16 @@ void multi_sink_example() {
 
 // User defined types logging
 struct my_type {
-    int i = 0;
-    explicit my_type(int i)
-        : i(i) {}
+    int value_ = 0;
+    explicit my_type(int value)
+        : value_(value) {}
 };
 
 #ifndef SPDLOG_USE_STD_FORMAT  // when using fmtlib
 template <>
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @KaganCanSit . Not sure about additional flags. There are so many we could add.

PR #3530, opened 2026-01-20
Allow empty DEBUG_POSTFIX property in CMakeLists

t0mmy742:
CMake cannot call set_target_properties() with an empty property.

Adding double quotes allows an empty DEBUG_POSTFIX property.

[For example, protobuf is doing it](https://github.com/protocolbuffers/protobuf/blob/v33.4/cmake/libprotobuf.cmake#L40)


gabime:
Thanks @t0mmy742

Issue #3525, opened 2026-01-16
Loading log levels from env does not work with uppercase level names


neundorf:
Hi,

when I set the log levels names to be uppercase via SPDLOG_LEVEL_NAMES in tweakme.h, 
SPDLOG_INLINE void load_levels() does not work.
It calls to_lower_(lvl) on the level name from the environment, and then checks in from_str() whether this is equal to one of the string in SPDLOG_LEVEL_NAMES. This is always false, if the level names are not lowercase.

tt4g:
You're right. The current behavior appears to be a bug.
Can you send PR?

neundorf:
What is the right way to fix it in your opinion ?
Would it be Ok if spdlog::levels::from_str() (which is public API) would become case-insensitive, or should this stay case-sensitive ?
Then only load_levels() would be case-insensitive.

tt4g:
I believe making it case-sensitive is the correct approach.
Otherwise, it will also fail to work when defining log levels in `SPDLOG_LEVEL_NAMES` that differ by only a single uppercase or lowercase letter.

BTW.
It appears only `“warn”` and `“err”` are treated specially. For compatibility reasons, these should be left as-is.

https://github.com/gabime/spdlog/blob/472945ba489e3f5684761affc431ae532ab5ed8c/tests/test_misc.cpp#L66-L76

https://github.com/gabime/spdlog/blob/472945ba489e3f5684761affc431ae532ab5ed8c/include/spdlog/common-inl.h#L36-L42

neundorf:
Sorry, I'm a bit confused by your second sentence.
Do you mean it should be possible to define log level names like "Debug" and "debug", and they should be treated and recognized as different log levels ?
Or do you mean that if I set the level to "Debug" in the environment variable, it should be recognized as "debug" from_str() ?

tt4g:
> Do you mean it should be possible to define log level names like "Debug" and "debug"

I apologize for the confusion. That's the first meaning.
I don't know if anyone actually defines level names like the following, but if it's possible, it should be supported.
```cpp
#define SPDLOG_LEVEL_NAMES { "debug", "Debug", ... }
```

neundorf:
Ok, I can do that.
But that might "break" the setup of some users. Currently they can do 
export SPDLOG_LEVELS="DEBUG", and it will work, even if it is named "debug".

tt4g:
It might be best to check with @gabime.

neundorf:
[...]

PR #3531, opened 2026-01-24
make level_enum from_str() case insensitive #3525

neundorf:
This patch changes the behaviour of from_str() so that the level names are compared case-insensitive.
This has the side effect of making load_levels() work if the names were set to UPPER oder CamelCase in tweakme.h.

Beside that, now also the special cases for "off", "err" and "warn" now check that the names are actually "off", "error" and "warning" and have not been set to something else.
It would be strange behaviour if I would set the name of the error level e.g. to "Fehler" and spdlog would recognize "fehler" and also "err", but not "error" as that level.

The test also checks a mixed case string now, and also "err".

### DIFF
--- a/include/spdlog/common-inl.h
+++ b/include/spdlog/common-inl.h
@@ -29,15 +29,18 @@ SPDLOG_INLINE const char *to_short_c_str(spdlog::level::level_enum l) SPDLOG_NOE
 }
 
 SPDLOG_INLINE spdlog::level::level_enum from_str(const std::string &name) SPDLOG_NOEXCEPT {
-    auto it = std::find(std::begin(level_string_views), std::end(level_string_views), name);
+    auto it = std::find_if(std::begin(level_string_views), std::end(level_string_views),
+                           [name](string_view_t sv) { return strings_equal_ci(name, sv); } );
+
     if (it != std::end(level_string_views))
         return static_cast<level::level_enum>(std::distance(std::begin(level_string_views), it));
 
     // check also for "warn" and "err" before giving up..
-    if (name == "warn") {
+    if (strings_equal_ci(name, "warn") && strings_equal_ci(to_string_view(level::warn), "warning")) {
         return level::warn;
     }
-    if (name == "err") {
+    else if (strings_equal_ci(name, "err") && strings_equal_ci(to_string_view(level::err), "error")) {
+    // if (name == "err") {
         return level::err;
     }
     return level::off;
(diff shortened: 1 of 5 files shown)


neundorf:
@gabime This patch makes from_str() case-insensitive.
I'm not sure why the builds here in CI failed initially, it was building for me.
As far as I can see it was related to #ifndef SPDLOG_HEADER_ONLY. I looked around, and I didn't find the cmake option to turn this on and off. How do I ?

tt4g:
IMO.

Common symbols used in spdlog are defined in `common.h`. You should not use `cfg/helpers.h` in this header file. It may be better to move the lowercase API to `common.h`.

Another way: Since `to_lower_()` is only used in `cfg/helpers-inl.h`, it may be possible to define it without external linkage in `common-inl.h` (move function).

neundorf:
@tt4g what do you think about my current version ?

tt4g:
@neundorf I'm concerned that the `to_lower()` API declaration is missing in `common.h`.

EDIT: This API might be reused in other spdlog code, so it might be better to have a declaration.

neundorf:
Hmm, to_lower() is currently only used locally in strings_equal_ci(),which is in common.h.
I don't see a strong reason to make to_lower() "public" right now. It also wasn't public before.

tt4g:
Yes. Whether to export `to_lower()` in `common.h` is a matter of personal preference.
Therefore, whether we should change it depends on what @gabime thinks.

include/spdlog/common-inl.h
```diff
@@ -65,4 +68,24 @@ SPDLOG_INLINE void throw_spdlog_ex(const std::string &msg, int last_errno) {
 
 SPDLOG_INLINE void throw_spdlog_ex(std::string msg) { SPDLOG_THROW(spdlog_ex(std::move(msg))); }
 
+#if __cplusplus >= 201703L
+constexpr
+#endif
+SPDLOG_API char to_lower(char ch) {
```
gabime, line 74: why SPDLOG_API. I would expect SPDLOG_INLINE

include/spdlog/common.h
```diff
@@ -296,6 +299,11 @@ enum class pattern_time_type {
     utc     // log utc
 };
 
+//
+// Returns true if two strings are equal, case-insensitive
+//
+SPDLOG_API bool strings_equal_ci(string_view_t s1, string_view_t s2);
```
gabime, line 305: Why SPDLOG_API ?  i would expect SPDLOG_INLINE

neundorf: I can change that, no problem.

But if I didn't mix things up, there is a second merge request for the same issue:
https://github.com/gabime/spdlog/pull/3535

If you prefer the other one, I wouldn't put more work into my merge request.

gabime: Not sure yet. The other merge seems bit leaner indeed

gabime:
I merged #3535 instead. Thanks anyway @neundorf !

PR #3534, opened 2026-02-01
Fix should_log comment in logger.h

eli-b:
"return true logging is enabled for the given level." -> "return true **if** logging is enabled for the given level."


gabime:
Thanks @eli-b

PR #3535, opened 2026-02-04
Fix #3525: Make level name matching case-insensitive

### LINKED ISSUE #3525: Loading log levels from env does not work with uppercase level names

ssam18:
This fixes the issue where uppercase level names in environment variables weren't recognized. The from_str() function now does case-insensitive matching, so SPDLOG_LEVEL="DEBUG" works just like "debug". All existing tests pass plus new tests verify uppercase and mixed-case level names work correctly.

### DIFF
--- a/include/spdlog/common-inl.h
+++ b/include/spdlog/common-inl.h
@@ -29,15 +29,31 @@ SPDLOG_INLINE const char *to_short_c_str(spdlog::level::level_enum l) SPDLOG_NOE
 }
 
 SPDLOG_INLINE spdlog::level::level_enum from_str(const std::string &name) SPDLOG_NOEXCEPT {
-    auto it = std::find(std::begin(level_string_views), std::end(level_string_views), name);
+    auto it = std::find_if(std::begin(level_string_views), std::end(level_string_views),
+                           [&name](const string_view_t &level_name) {
+                               return level_name.size() == name.size() &&
+                                      std::equal(name.begin(), name.end(), level_name.begin(),
+                                                 [](char a, char b) {
+                                                     return std::tolower(static_cast<unsigned char>(a)) ==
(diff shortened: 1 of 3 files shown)


tt4g:
Thank you for contribution but duplicate this PR: #3531

SamareshSingh:
Hmm. I didn't realized it. Thanks for checking it

include/spdlog/common-inl.h
```diff
@@ -29,15 +29,31 @@ SPDLOG_INLINE const char *to_short_c_str(spdlog::level::level_enum l) SPDLOG_NOE
...
 
 SPDLOG_INLINE spdlog::level::level_enum from_str(const std::string &name) SPDLOG_NOEXCEPT {
-    auto it = std::find(std::begin(level_string_views), std::end(level_string_views), name);
+    auto it = std::find_if(std::begin(level_string_views), std::end(level_string_views),
+                           [&name](const string_view_t &level_name) {
+                               return level_name.size() == name.size() &&
+                                      std::equal(name.begin(), name.end(), level_name.begin(),
+                                                 [](char a, char b) {
+                                                     return std::tolower(static_cast<unsigned char>(a)) ==
+                                                            std::tolower(static_cast<unsigned char>(b));
```
gabime, line 38: std::tolower requires include<cctype>

gabime (changes requested):

PR #3549, opened 2026-03-01
Add constructor for dup_filter_sink with sinks parameter

Niram7777:


Niram7777:
for some strange reasons,  I am still not able to add any sink at all to dup_filter_sink (my other sinks work)
I have tried pretty much everything...
image
image

Niram7777:
the issue also happens with just
```cpp
        std::initializer_list<spdlog::sink_ptr> sinks{
            console_sink, file_sink,
            std::make_shared<
                spdlog::sinks::
                    null_sink_mt>() /*for POSITON_GAME_LOG id + ".log" */};
        spdlog::sinks::dist_sink<std::mutex>().set_sinks(sinks);
```

tt4g:
The cause is the usage of `std::initializer_list`.
Copying `std::initializer_list` does not copy its internal elements: https://stackoverflow.com/questions/27157202/can-it-be-safe-to-keep-a-copy-of-an-stdinitializer-list-what-is-the-rationale
Change `sinks` to a different container type (such as `std::vector`).

Niram7777:
> The cause is the usage of `std::initializer_list`. Copying `std::initializer_list` does not copy its internal elements: 

it was actually an issue with lldb 23, the vector size is correct with 19 - I am surprized with the rootcause but will report the issue to LLVM

my PR is ready

tt4g:
No, your code is incorrect.
As noted in the StackOverflow comment, the C++ standard explicitly states that "Copying an initializer list does not copy the underlying elements".

As you can see on the cppreference.com page for [the `std::initializer_list` constructor](https://en.cppreference.com/w/cpp/utility/initializer_list/initializer_list.html), `std::initializer_list` does not have a copy constructor.
This means that copying the `std:initializer_list` is not defined by the C++ standard.

Applying the following changes to your code will make it work:

```diff
+        std::vector<spdlog::sink_ptr> sinks({
-        std::initializer_list<spdlog::sink_ptr> sinks{
             console_sink, file_sink,
             std::make_shared<
                 spdlog::sinks::
+                    null_sink_mt>() /*for POSITON_GAME_LOG id + ".log" */});
-                    null_sink_mt>() /*for POSITON_GAME_LOG id + ".log" */};
         spdlog::sinks::dist_sink<std::mutex>().set_sinks(sinks);
```

include/spdlog/sinks/dup_filter_sink.h
```diff
@@ -43,6 +43,11 @@ class dup_filter_sink : public dist_sink<Mutex> {
     explicit dup_filter_sink(std::chrono::duration<Rep, Period> max_skip_duration)
         : max_skip_duration_{max_skip_duration} {}
 
+    template <class Rep, class Period>
+    explicit dup_filter_sink(std::chrono::duration<Rep, Period> max_skip_duration, std::vector<std::shared_ptr<sink>> sinks)
+        : dist_sink<Mutex>(std::move(sinks))
+        , max_skip_duration_{max_skip_duration} {}
+
```
gabime, line 50: max_skip_duration  should be initialized before the vector , to follow the  declare order

gabime (changes requested):

tt4g:
@Niram7777 I think this change was created for shadps4-emu/shadPS4#4069, but since [the previous suggestion](https://github.com/gabime/spdlog/pull/3549#issuecomment-3981789479) resolved your issue, is this PR no longer needed?

Niram7777:
> @Niram7777 I think this change was created for [shadps4-emu/shadPS4#4069](https://github.com/shadps4-emu/shadPS4/pull/4069), but since [the previous suggestion](https://github.com/gabime/spdlog/pull/3549#issuecomment-3981789479) resolved your issue, is this PR no longer needed?

sorry I have almost no free time during the week, and I am super busy on the weekend.
I am not forgetting this PR although I though it was trivial, I will look at it this weekend

include/spdlog/sinks/dup_filter_sink.h
Niram7777: https://en.cppreference.com/w/cpp/language/initializer_list.html

> 2) Then, direct bases are initialized in left-to-right order as they appear in this class's base-specifier list.
> 3) Then, non-static data member are initialized in order of declaration in the class definition.

gabime: Even so, the code should be clean as possible

Niram7777:
> No, your code is incorrect. As noted in the StackOverflow comment, the C++ standard explicitly states that "Copying an initializer list does not copy the underlying elements".
> 
> As you can see on the cppreference.com page for [the `std::initializer_list` constructor](https://en.cppreference.com/w/cpp/utility/initializer_list/initializer_list.html), `std::initializer_list` does not have a copy constructor. This means that copying the `std:initializer_list` is not defined by the C++ standard.
> 
> Applying the following changes to your code will make it work:
> 
> ```diff
> +        std::vector<spdlog::sink_ptr> sinks({
> -        std::initializer_list<spdlog::sink_ptr> sinks{
>              console_sink, file_sink,
>              std::make_shared<
>                  spdlog::sinks::
> +                    null_sink_mt>() /*for POSITON_GAME_LOG id + ".log" */});
> -                    null_sink_mt>() /*for POSITON_GAME_LOG id + ".log" */};
>          spdlog::sinks::dist_sink<std::mutex>().set_sinks(sinks);
> ```


*"Copying an initializer list does not copy the underlying elements".*

yes and what the standard says is that it will be a *shallow* copy -> so the representation will be copied, so the size should not be 0

> std::initializer_list does not copy the [backing array](https://en.cppreference.com/w/cpp/language/list_initialization.html#List-initializing_std::initializer_list) of the corresponding initializer list.
```cpp
std::initializer_list<int> il{-3, -2, -1};
assert(il.begin()[2] == -1); // note the replacement for absent operator[]
il = al; // shallow-copy
assert(il.begin() == al.begin()); // guaranteed
```

also
> A std::initializer_list object is automatically constructed when:
a brace-enclosed initializer list is used as the right operand of [assignment](https://en.cppreference.com/w/cpp/language/operator_assignment.html#Builtin_direct_assignment) or as a [function call argument](https://en.cppreference.com/w/cpp/language/overload_resolution.html#Implicit_conversion_sequence_in_list-initialization), and the corresponding assignment operator/function accepts an std::initializer_list parameter,

https://en.cppreference.com/w/cpp/container/vector/vector.html
> vector( [std::initializer_list](https://en.cppreference.com/w/cpp/utility/initializer_list.html)<T> init,
        const Allocator& alloc = Allocator() );
...
> 11) Equivalent to vector(il.begin(), il.end(), alloc).

gabime:
Thanks @Niram7777

PR #3561, opened 2026-03-14
docs: fix SPDLOG_LEVEL env example

omerrbbulut:
## Summary

Fix the `SPDLOG_LEVEL` environment variable example in `include/spdlog/cfg/env.h`.

The previous example used:

`*=off,logger1=debug`

However, the current `helpers::load_levels()` implementation treats global
levels as entries without a logger name, so the correct example is:

`off,logger1=debug`

This update aligns the documentation example with the current behavior.

### DIFF
--- a/include/spdlog/cfg/env.h
+++ b/include/spdlog/cfg/env.h
@@ -17,7 +17,7 @@
 // export SPDLOG_LEVEL=debug
 //
 // turn off all logging except for logger1:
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @omerrbbulut

omerrbbulut:
Thanks! Glad to help.

PR #3571, opened 2026-03-25
add new class serial_port_sink allow sending msg over serial comm port

cppforever1:
add  serial_port_sink for both Windows and Linux OS


gabime:
Thanks. There are many ways one can want to use serial port (overllaped/async/etc). It is not really the scope of this lib

PR #3575, opened 2026-04-01
[ci] Update actions/checkout to latest major relese

gruenich:
Fix warning that Node.js 20 will stop working in June


gabime:
Thanks @gruenich

PR #3579, opened 2026-04-03
[bench] Update Google benchmark to version 1.8.4

gruenich:
Fix CMake warning from more recent versions.
Latest benchmark release support C++11.
Does not increase CMake requirements.


gabime:
Thanks @gruenich

PR #3583, opened 2026-04-08
Set C++ standard to 20 if not defined

loic-lopez:
Why? to be able to use a greater c++ when using conan.

Because conan reports: 

> Warning: Standard CMAKE_CXX_STANDARD value defined in conan_toolchain.cmake to 23 has been modified to 20 by .conan2/p/b/spdlo815dcec129526/b/src/CMakeLists.txt


gabime:
Thanks @loic-lopez

PR #3594, opened 2026-04-13
Fix socket closure in tcp_client to prevent errors on invalid socket

### LINKED ISSUE #3587: [Windows] tcp_client::close() calls closesocket() unconditionally, generating spurious WSAENOTSOCK errors on disconnected sockets
### Summary

In include/spdlog/details/tcp_client-windows.h, the close() method calls closesocket() unconditionally without first checking whether socket_ holds a valid handle. This generates a spurious WSAENOTSOCK Winsock error every time close() is invoked on an already-disconnected client — including on every normal destructor call.

### Affected file

include/spdlog/details/tcp_client-windows.h

### Current code

void close() {
```c++
 ::closesocket(socket_); // called even when socket_ == INVALID_SOCKET
    socket_ = INVALID_SOCKET;
}
```

### Expected behaviour

close() should be idempotent. Calling it on an already-closed or never-connected client should be a safe no-op, consistent with standard RAII socket hygiene and the behaviour of the POSIX equivalent.

### Impact

Every tcp_client destruction after a failed connection or after the server has dropped the connection triggers a silent WSAENOTSOCK error internally.
Any error handler or diagnostic tooling (e.g. Application Verifier, custom Winsock error hooks) that monitors for socket errors will flag this as a defect on every normal teardown path.
In reconnecting scenarios — where close() may be called repeatedly between reconnect attempts — the spurious error is generated on every attempt.
Proposed fix

Guard the closesocket() call with an INVALID_SOCKET check:
``` c++
void close() {
+    if (socket_ != INVALID_SOCKET) {
        ::closesocket(socket_);
        socket_ = INVALID_SOCKET;
+    }
}
```

nbakovljev:
Fixes #3587

### DIFF
--- a/include/spdlog/details/tcp_client-windows.h
+++ b/include/spdlog/details/tcp_client-windows.h
@@ -52,8 +52,10 @@ public:
     bool is_connected() const { return socket_ != INVALID_SOCKET; }
 
     void close() {
-        ::closesocket(socket_);
-        socket_ = INVALID_SOCKET;
+        if (socket_ != INVALID_SOCKET) {
+            ::closesocket(socket_);
+            socket_ = INVALID_SOCKET;
+        }
     }
 
     SOCKET fd() const { return socket_; }


gabime:
Thanks @nbakovljev

PR #3595, opened 2026-04-13
Tcp client auto reconnect

### LINKED ISSUE #3586: [Windows] tcp_sink fails to reconnect after server restart due to premature WSACleanup() calls
### Summary

On Windows, the tcp_sink permanently loses its ability to reconnect after the remote TCP server crashes or restarts. All subsequent reconnect attempts throw an exception, and the sink never recovers for the lifetime of the process.

### Root cause

WSAStartup() is called exactly once, in the tcp_client constructor. However, WSACleanup() is called in three additional error paths inside connect():

When getaddrinfo() fails
When socket() fails for a candidate address (inside the address loop)
When all addresses are exhausted without a successful connection
Winsock's reference count is decremented on each of these calls. After the first failed reconnect attempt the count reaches zero, Winsock is torn down, and every subsequent socket call returns WSANOTINITIALISED.

### Steps to reproduce

Create a tcp_sink_mt pointed at a local TCP server on Windows.
Log several messages successfully.
Kill the TCP server.
Restart the TCP server.
Attempt to log further messages — the sink throws and never reconnects.
Expected behaviour

The sink should reconnect transparently on the next log call, consistent with the behaviour described in the tcp_sink documentation ("Will attempt to reconnect if connection drops").

### Affected file

include/spdlog/details/tcp_client-windows.h

### Proposed fix

Remove the three erroneous WSACleanup() calls from connect(). Winsock lifetime should mirror the object lifetime — WSAStartup in the constructor, WSACleanup in the destructor only.

``` c++
- WSACleanup(); // in getaddrinfo error path
- WSACleanup(); // in socket() error path (inside address loop)
- WSACleanup(); // in final INVALID_SOCKET check
```

nbakovljev:
Fixes #3586

### DIFF
--- a/include/spdlog/details/tcp_client-windows.h
+++ b/include/spdlog/details/tcp_client-windows.h
@@ -52,9 +52,9 @@ public:
     bool is_connected() const { return socket_ != INVALID_SOCKET; }
 
     void close() {
-        ::closesocket(socket_);
-        socket_ = INVALID_SOCKET;
-    }
+            ::closesocket(socket_);
+            socket_ = INVALID_SOCKET;
+        }
 
     SOCKET fd() const { return socket_; }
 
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @nbakovljev

PR #3596, opened 2026-04-13
docs: fix misleading unit wording in comment

Daschi1:
Fixes a misleading comment in `stopwatch.h` where the wording referred to "double" instead of the actual unit ("seconds").

If I've misunderstood and the original wording was intentional, feel free to disregard this PR.


gabime:
It is actually a double representing seconds(e.g. 2.34 seconds), so we are both right I guess

PR #3604, opened 2026-04-30
converted fractional_part_rounding_thresholds to use an array of u32s

SystemPanic:
MSVC fails to compile format.h with "character value is out of range" errors.

PR is the same already merged on [https://github.com/fmtlib/fmt/pull/4719](https://github.com/fmtlib/fmt/pull/4719)


tt4g:
Thank you for your contribution.
However, spdlog uses a method where the entire source code is replaced when a new version of fmt is released.
Please don’t be disappointed if this PR isn’t merged.

If you can't wait for the next fmt release, you can use an external fmt instead of the fmt bundled in spdlog.

SystemPanic:
@tt4g Sure, no problem. I already have a temporary patch for FlashInfer that fix it automatically.

Feel free to close this.

PR #3606, opened 2026-05-05
Add systemd_namespace_sink for logging to specific systemd namespace

oko256:
Adds systemd_namespace_sink that can open a stream to a specific systemd journal namespace for logging. This was created as a separate sink from systemd_sink because it has significant difference due to systemd's journal API: `sd_journal_send()` does not have namespace-specific counterpart and thus this cannot utilize other extra journal fields than `PRIORITY` (via kernel-style log level prefixes).

Another difference is that this namespace stream API is only available from systemd version 256, thus a version check was added to CMake file for the unit tests so that they are not automatically included if `libsystemd` is not new enough.

Unit tests are split to two parts: first one runs always and it just uses the stream to log to default namespace (should always work). Second one requires manual setting up of "custom" journal namespace and thus is ignored from the default set of tests (can be run by specifying `[systemd_namespace]` tag explicitly when running tests). On the other hand, the second part also automatically checks the output correctness from journal.

### DIFF
--- a/include/spdlog/sinks/systemd_namespace_sink.h
+++ b/include/spdlog/sinks/systemd_namespace_sink.h
@@ -8,9 +8,6 @@
 #include <spdlog/sinks/base_sink.h>
 
 #include <array>
-#ifndef SD_JOURNAL_SUPPRESS_LOCATION
-#define SD_JOURNAL_SUPPRESS_LOCATION
-#endif
 #include <systemd/sd-journal.h>
 #include <systemd/sd-daemon.h>
 
@@ -29,14 +26,7 @@ public:
     explicit systemd_namespace_sink(std::string ident,
                                     std::string name_space,
                                     bool enable_formatting)
-        : enable_formatting_{enable_formatting},
-          level_prefixes_{{/* spdlog::level::trace      */ SD_DEBUG,
-                           /* spdlog::level::debug      */ SD_DEBUG,
-                           /* spdlog::level::info       */ SD_INFO,
-                           /* spdlog::level::warn       */ SD_WARNING,
-                           /* spdlog::level::err        */ SD_ERR,
-                           /* spdlog::level::critical   */ SD_CRIT,
-                           /* spdlog::level::off        */ SD_INFO}} {
+        : enable_formatting_{enable_formatting} {
         int stream_fd = -1;
         if (name_space.empty()) {
             stream_fd = ::sd_journal_stream_fd(ident.c_str(), LOG_INFO, 1);
@@ -75,7 +65,15 @@ protected:
     FILE *journal_ = nullptr;
     bool enable_formatting_ = false;
     using level_prefix_array = std::array<const char *, 7>;
-    level_prefix_array level_prefixes_;
+    static constexpr level_prefix_array level_prefixes_ = {
+        {/* spdlog::level::trace      */ SD_DEBUG,
+         /* spdlog::level::debug      */ SD_DEBUG,
+         /* spdlog::level::info       */ SD_INFO,
+         /* spdlog::level::warn       */ SD_WARNING,
+         /* spdlog::level::err        */ SD_ERR,
+         /* spdlog::level::critical   */ SD_CRIT,
+         /* spdlog::level::off        */ SD_INFO}};
+    static constexpr size_t level_prefix_length_ = 3;
 
     void sink_it_(const details::log_msg &msg) override {
         string_view_t payload;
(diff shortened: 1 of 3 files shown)


include/spdlog/sinks/systemd_namespace_sink.h
```diff
@@ -0,0 +1,145 @@
...
+}
+
+template <typename Factory = spdlog::synchronous_factory>
+inline std::shared_ptr<logger> systemd_namespace_logger_st(const std::string &logger_name,
+                                                           const std::string &ident,
+                                                           const std::string &name_space,
+                                                           bool enable_formatting) {
+    return Factory::template create<sinks::systemd_namespace_sink_st>(
+        logger_name, ident, name_space, enable_formatting);
+}
```
gabime, line 157: what about flush_() impl?
`_IOLBF` handles line-based flushing, but  I think  flush()_ should explicitly call `fflush(journal_)` to ensure any partially buffered data or pending stream state is synchronized.

include/spdlog/sinks/systemd_namespace_sink.h
```diff
@@ -0,0 +1,145 @@
...
+            }
+
+            // Write log level prefix
+            details::os::fwrite_bytes(
+                level_prefixes_.at(static_cast<level_prefix_array::size_type>(msg.level)), 3,
+                journal_);
+            // Write the message
+            details::os::fwrite_bytes(one_message.data(), one_message.size(), journal_);
+            // Append newline if the message didn't have one
+            if (one_message.empty() || one_message.back() != '\n') {
```
gabime: why check again back for newline? you already know it if (nl_pos != std::string_view::npos);

include/spdlog/sinks/systemd_namespace_sink.h
```diff
@@ -0,0 +1,145 @@
...
+            if ((length + 3) > static_cast<size_t>(std::numeric_limits<int>::max())) {
+                length = static_cast<size_t>(std::numeric_limits<int>::max()) - 3;
+            }
+
+            // Write log level prefix
+            details::os::fwrite_bytes(
+                level_prefixes_.at(static_cast<level_prefix_array::size_type>(msg.level)), 3,
+                journal_);
+            // Write the message
+            details::os::fwrite_bytes(one_message.data(), one_message.size(), journal_);
```
gabime: Shouldn't `length ` that was calculated above be used here?

gabime (changes requested):

include/spdlog/sinks/systemd_namespace_sink.h
oko256: Actually I thought about this a bit more and this whole truncation logic is not really correct nor needed. In systemd_sink, this cap exists just to prevent the formatting function from overflowing (cast safety basically). Here the actual message length limit is enforced by journal itself (based on its own configuration), and if it is exceeded, journald truncates at its end. Whatever cap we would choose here would be arbitrary and wrong. So I'll just drop the truncation completely.

include/spdlog/sinks/systemd_namespace_sink.h
oko256: You are correct, can just use nl_pos to check this, thanks for spotting. Actually a follow-up question: What is the spdlog's intended behavior when trying to log an empty message? This implementation of sink now skips empty messages completely, but should I change it so that it actually does log an empty message to journal?

include/spdlog/sinks/systemd_namespace_sink.h
oko256: True, I will add flush_() implementation.

include/spdlog/sinks/systemd_namespace_sink.h
oko256: > You are correct, can just use nl_pos to check this, thanks for spotting. Actually a follow-up question: What is the spdlog's intended behavior when trying to log an empty message? This implementation of sink now skips empty messages completely, but should I change it so that it actually does log an empty message to journal?

Answering to myself: journald itself ignores zero-length messages, so no need to change the behavior of this sink regarding that.

oko256:
Did a bunch of fixes, cleanup, and some optimizations. See the commit message for details.

include/spdlog/sinks/systemd_namespace_sink.h
```diff
@@ -87,44 +85,48 @@ class systemd_namespace_sink : public base_sink<Mutex> {
             payload = msg.payload;
         }
 
+        auto write_or_throw = [this](const void *data, size_t n) {
```
gabime: please avoid needless lambdas.

include/spdlog/sinks/systemd_namespace_sink.h
```diff
@@ -0,0 +1,147 @@
...
+        if (name_space.empty()) {
+            stream_fd = ::sd_journal_stream_fd(ident.c_str(), LOG_INFO, 1);
+        } else {
+            stream_fd = ::sd_journal_stream_fd_with_namespace(name_space.c_str(), ident.c_str(),
+                                                              LOG_INFO, 1);
+        }
+        if (stream_fd < 0) {
+            throw_spdlog_ex("Failed opening systemd journal stream to namespace '" + name_space +
+                            "': " + ::strerror(-stream_fd));
+        }
```
gabime, line 41: Please use throw_spdlog_ex(..) like the rest of code. Also, why negate the stream_fd? it is already negative as strerror expects.

include/spdlog/sinks/systemd_namespace_sink.h
```diff
@@ -0,0 +1,147 @@
...
+        if (journal_ == nullptr) {
+            // Capture errno from the failed fdopen() instead of close() (in case that fails too)
+            const int saved_errno = errno;
+            ::close(stream_fd);
+            throw_spdlog_ex(
+                "Failed opening systemd journal stream to namespace '" + name_space + "'",
+                saved_errno);
+        }
+        // Use line buffering which matches with journald's line-oriented protocol
+        ::setvbuf(journal_, nullptr, _IOLBF, 0);
```
gabime: should check retval. might fail

gabime (changes requested):

include/spdlog/sinks/systemd_namespace_sink.h
oko256: AFAIK strerror() takes errno like any other function and valid errno values are always positive. This is also according to example in [sd_journal_stream_fd() documentation](https://www.freedesktop.org/software/systemd/man/latest/sd_journal_stream_fd.html) where they also negate the fd to pass it as a valid errno.

I will change the code to use throw_spdlog_ex() without strerror() and just pass the errno as a parameter to the function (which I assume is what you meant), but to my understanding the fd still needs to be negated to be a valid positive errno value.

include/spdlog/sinks/systemd_namespace_sink.h
oko256: My rationale for this was that since this is only used inside sink_it_(), it would be cleanest like this, but I will change it to a class method instead.

include/spdlog/sinks/systemd_namespace_sink.h
gabime: Yes, you are right regarding the negative retvals.
Better use spdlog_ex anyway

include/spdlog/sinks/systemd_namespace_sink.h
```diff
@@ -0,0 +1,159 @@
...
+    using level_prefix_array = std::array<const char *, 7>;
+    static constexpr level_prefix_array level_prefixes_ = {
+        {/* spdlog::level::trace      */ SD_DEBUG,
+         /* spdlog::level::debug      */ SD_DEBUG,
+         /* spdlog::level::info       */ SD_INFO,
+         /* spdlog::level::warn       */ SD_WARNING,
+         /* spdlog::level::err        */ SD_ERR,
+         /* spdlog::level::critical   */ SD_CRIT,
+         /* spdlog::level::off        */ SD_INFO}};
+    static constexpr size_t level_prefix_length_ = 3;
```
gabime: We cannot rely it will be exactly 3 in all platforms. what if it changes?

gabime (changes requested):

include/spdlog/sinks/systemd_namespace_sink.h
oko256: Systemd is Linux-only, and I highly doubt that they would decide to change that since it's aligned with kernel logging prefix, but I agree that it's a magic number and actually string_view is much nicer way to do this anyway so I changed it. Thanks for pointing it out :+1:

include/spdlog/sinks/systemd_namespace_sink.h
```diff
@@ -104,8 +103,8 @@ class systemd_namespace_sink : public base_sink<Mutex> {
 
             // Write log level prefix
             write_or_throw_(
-                level_prefixes_.at(static_cast<level_prefix_array::size_type>(msg.level)),
-                level_prefix_length_);
+                level_prefixes_.at(static_cast<level_prefix_array::size_type>(msg.level)).data(),
+                level_prefixes_.at(static_cast<level_prefix_array::size_type>(msg.level)).size());
             // Write the message
```
gabime, line 108: this is cleaner and more efficient:
```c++
const auto &prefix = static_cast<level_prefix_array::size_type>(msg.level);
write_or_throw_(prefix.data(), prefix.size());
```

gabime (changes requested):

gabime:
Thanks @oko256 !

PR #3618, opened 2026-06-05
Fix invalid spdlog::std::make_unique in mongo_sink

hel-ableton:
mongo_sink_config constructed its client with spdlog::std::make_unique, a stray spdlog:: prefix left over from a bulk find-and-replace in commit 19d4e605 ("Replaced details::make_unique with std::make_unique"), which rewrote details::make_unique to std::make_unique but left the leading spdlog:: on this line that originally read spdlog::details::make_unique.

It only resolved by accident via enclosing-namespace lookup of ::std, and would break outright if the namespace were renamed. Call std::make_unique directly.

### DIFF
--- a/include/spdlog/sinks/mongo_sink.h
+++ b/include/spdlog/sinks/mongo_sink.h
@@ -44,7 +44,7 @@ public:
           db_name_(db_name),
           coll_name_(collection_name) {
         try {
-            client_ = spdlog::std::make_unique<mongocxx::client>(mongocxx::uri{uri});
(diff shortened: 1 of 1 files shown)


gabime:
Thanks @hel-ableton . What's the status of PRs #3610 and #3611 ? I see lots of activity. Not sure if they can already merged or if there is still work..

hel-ableton:
@gabime You're welcome! As far as #3610 and #3611 go, I kept pushing fixups (and a few new commits) as reactions to the CoPilot reviews. Both branches also have an added test now. Do you want to give them a final review?

PR #3620, opened 2026-06-06
Grafana Loki sink

Alvov1:
Adds a sink that pushes log entries to Grafana Loki via its HTTP push API. Consistent with `kafka_sink` and `mongo_sink` — header-only, requires two external dependencies. Default pattern is `%v` (message text only): timestamp and log level go into Loki structured fields rather than the log line

https://grafana.com/docs/loki/latest/reference/loki-http-api/#ingest-logs

**Usage**
```cpp
#include "spdlog/sinks/loki_sink.h"

spdlog::sinks::loki_sink_config config("localhost", 3100);
config.labels          = {{"job", "myapp"}, {"env", "production"}};
config.add_level_label = true;   // adds level="info" etc. as a stream label
config.timeout_seconds = 5;

auto logger = spdlog::loki_logger_async_mt("loki", config);
logger->info("Hello from spdlog!");
```

**Dependencies**
- https://github.com/nlohmann/json
- https://github.com/yhirose/cpp-httplib

### DIFF
--- a/include/spdlog/sinks/loki_sink.h
+++ b/include/spdlog/sinks/loki_sink.h
@@ -61,6 +61,7 @@ public:
         // Users can call set_pattern() to include additional context (e.g. thread id).
         base_sink<Mutex>::set_formatter_(
             details::make_unique<pattern_formatter>("%v", pattern_time_type::local, ""));
+        client_.set_keep_alive(true);
         if (config_.timeout_seconds > 0) {
             client_.set_connection_timeout(config_.timeout_seconds);
             client_.set_read_timeout(config_.timeout_seconds);
--- a/include/spdlog/sinks/loki_sink.h
+++ b/include/spdlog/sinks/loki_sink.h
@@ -15,8 +15,9 @@
 // For production use, prefer loki_logger_async_mt: sink_it_() makes a blocking
 // HTTP call and should not run on the caller's thread.
 //
-// TODO: batching multiple log entries per HTTP request for high-throughput use cases.
-// TODO: TLS support and Authorization header for Grafana Cloud / auth-protected setups.
+// Limitations:
+// - Sends one log entry per HTTP request (no batching).
+// - No TLS/auth header support (e.g. Grafana Cloud / auth-protected setups).
 
 #include <spdlog/async.h>
 #include <spdlog/common.h>
--- a/include/spdlog/sinks/loki_sink.h
+++ b/include/spdlog/sinks/loki_sink.h
@@ -93,7 +93,7 @@ private:
         if (ns < 0) ns = 0;
 
         nlohmann::json labels = labels_json_;
-        if (config_.add_level_label) {
+        if (config_.add_level_label && !labels.contains("level")) {
             auto sv = level::to_string_view(msg.level);
             labels["level"] = std::string(sv.data(), sv.size());
         }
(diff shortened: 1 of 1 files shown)


include/spdlog/sinks/loki_sink.h
```diff
@@ -77,7 +76,7 @@ class loki_sink : public base_sink<Mutex> {
     void sink_it_(const details::log_msg &msg) override {
         memory_buf_t formatted;
         base_sink<Mutex>::formatter_->format(msg, formatted);
-        auto body = build_json_(msg, std::string_view(formatted.data(), formatted.size()));
+        auto body = build_json_(msg, std::string(formatted.data(), formatted.size()));
```
gabime, line 81: just use spdlog’s  “string_view_t”.

include/spdlog/sinks/loki_sink.h
```diff
@@ -86,7 +85,7 @@ class loki_sink : public base_sink<Mutex> {
     void flush_() override {}
 
 private:
-    std::string build_json_(const details::log_msg &msg, std::string_view line) const {
+    std::string build_json_(const details::log_msg &msg, const std::string &line) const {
```
gabime, line 97: ditto

include/spdlog/sinks/loki_sink.h
```diff
@@ -95,7 +94,7 @@ class loki_sink : public base_sink<Mutex> {
         nlohmann::json labels = labels_json_;
         if (config_.add_level_label) {
             auto sv = level::to_string_view(msg.level);
-            labels["level"] = std::string_view(sv.data(), sv.size());
+            labels["level"] = std::string(sv.data(), sv.size());
```
gabime, line 106: ditto

gabime (changes requested):

include/spdlog/sinks/loki_sink.h
Alvov1: I tried using **string_view_t** in build_json_'s signature, but nlohmann::json doesn't know how to serialize fmt::basic_string_view<char> (the fmtlib backend case), so the request body is malformed and Loki returns 400
  
How I see it, for both formatted and msg.level at least one std::string conversion is required to work properly with the fmtlib backend

include/spdlog/sinks/loki_sink.h
Alvov1: Answered above

include/spdlog/sinks/loki_sink.h
Alvov1: Answered above

gabime:
@copilot suggest a way to avoid allocations of std::string without using nlohmann json if possible

include/spdlog/sinks/loki_sink.h
```diff
@@ -0,0 +1,140 @@
...
+          labels_json_(config_.labels) {
+        // Timestamp and level go into Loki structured fields, so default to message text only.
+        // Users can call set_pattern() to include additional context (e.g. thread id).
+        base_sink<Mutex>::set_formatter_(
+            details::make_unique<pattern_formatter>("%v", pattern_time_type::local, ""));
+        if (config_.timeout_seconds > 0) {
+            client_.set_connection_timeout(config_.timeout_seconds);
+            client_.set_read_timeout(config_.timeout_seconds);
+            client_.set_write_timeout(config_.timeout_seconds);
+        }
```
gabime, line 70: how about
 ```cpp
client_.set_keep_alive(true);
```

to avoid reopening the connection for each log entry?

gabime (changes requested):

gabime:
Thanks @Alvov1

PR #3624, opened 2026-06-23
Make null_mutex compliant with "Mutex" named requirement

KabelitzJ:
I know this does not add that much value but I just stumbled over the fact that `null_mutex` is not fully compliant with the [Mutex named requirement](https://en.cppreference.com/cpp/named_req/Mutex).

I build locally and ran `ctest` with no errors.

Feel free to just delete this if this is just noise but I thought "why not shoot your shot" :)

### DIFF
--- a/include/spdlog/details/null_mutex.h
+++ b/include/spdlog/details/null_mutex.h
@@ -12,8 +12,18 @@
 SPDLOG_NAMESPACE_BEGIN
 namespace details {
 struct null_mutex {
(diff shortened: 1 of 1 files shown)


gabime:
Thanks. While probably correct, Not sure if this won’t break existing code the rely in null mutex to copy etc.

KabelitzJ:
Thx for your insight. I will close this PR

PR #3625, opened 2026-06-26
Handle EEXIST when creating directories

sage-mode-hunter:
This change treats EEXIST from mkdir() as a successful case in create_dir(). When multiple threads try to create the same directory at the same time, one thread may see EEXIST even though the directory already exists. Instead of treating that as a failure, the code now continues creating the remaining subdirectories.

I also ran the existing test suite after this change, and all tests passed.


gabime:
Thanks @sage-mode-hunter

PR #3626, opened 2026-06-26
Treat ENOENT as success in remove_if_exists

sage-mode-hunter:
remove_if_exists() now treats ENOENT from remove() as success.

This avoids a race condition where the file is deleted between the existence check and remove(), causing a false failure even though the file is already gone.

All other error cases are unchanged.


gabime:
Thanks @sage-mode-hunter

PR #3627, opened 2026-06-26
Improve robustness of daily_file_sink filename initialization

sage-mode-hunter:
This change makes the filename history initialization in daily_file_sink more robust when older rotated log files are missing.

In some environments, log files may be removed externally while newer ones still exist. In such cases, the previous logic could stop scanning early.

This update ensures the scan behaves more consistently when files are missing due to external cleanup.

No change in normal logging behavior.


gabime:
this doesnt make sense

PR #3628, opened 2026-06-26
Improve size tracking consistency in rotating_file_sink

sage-mode-hunter:
rotating_file_sink uses an internal cached size counter to track file growth and determine when log rotation should occur.

In environments where files may be externally modified (log rotation tools, truncation, or unexpected writes), this cached value can become temporarily inconsistent with the actual file size.

This change adds a lightweight resynchronization after rotation to ensure internal state remains consistent with the underlying file.

No change to rotation behavior or public API.

### DIFF
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -132,9 +132,9 @@ private:
         auto now = log_clock::now();
         while (filenames.size() < max_files_) {
             const auto new_filename = FileNameCalc::calc_filename(base_filename_, now_tm(now));
-            if (!path_exists(new_filename)) {
-                break;
-            }
+            if (filenames.empty() && !path_exists(new_filename)) {
+    break;
+}
             filenames.emplace_back(new_filename);
             now -= std::chrono::hours(24);
         }
--- a/include/spdlog/sinks/daily_file_sink.h
+++ b/include/spdlog/sinks/daily_file_sink.h
@@ -132,9 +132,9 @@ private:
         auto now = log_clock::now();
         while (filenames.size() < max_files_) {
             const auto new_filename = FileNameCalc::calc_filename(base_filename_, now_tm(now));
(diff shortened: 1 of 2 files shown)


include/spdlog/sinks/daily_file_sink.h
```diff
@@ -132,9 +132,9 @@ class daily_file_sink final : public base_sink<Mutex> {
         auto now = log_clock::now();
         while (filenames.size() < max_files_) {
             const auto new_filename = FileNameCalc::calc_filename(base_filename_, now_tm(now));
-            if (!path_exists(new_filename)) {
-                break;
-            }
+            if (filenames.empty() && !path_exists(new_filename)) {
+    break;
```
gabime: this logic is wrong. and not related to the pr. pls revert it

include/spdlog/sinks/rotating_file_sink-inl.h
```diff
@@ -117,6 +117,7 @@ SPDLOG_INLINE void rotating_file_sink<Mutex>::sink_it_(const details::log_msg &m
         file_helper_.flush();
         if (file_helper_.size() > 0) {
             rotate_();
+            current_size_ = file_helper_.size();
```
gabime: no need to call twice file_helper_.size(). just store into var the first result and use it. logic is wrong amyway, sinice lime 125 sets again the current size

gabime (changes requested):

include/spdlog/sinks/rotating_file_sink-inl.h
sage-mode-hunter: Thanks for the review , I’ve removed the unrelated daily_file_sink change and updated rotating_file_sink to avoid duplicate size calls as suggested.

gabime: Sorry this doesnt make any sense to me. Now size() is called in each call. Closing.

PR #3629, opened 2026-06-27
Use WSAGetLastError for UDP send errors on Windows

sage-mode-hunter:
### Summary

This change updates the Windows UDP client to use `WSAGetLastError()` when `sendto()` fails.

Winsock APIs report socket errors through `WSAGetLastError()`, while the current implementation uses `errno`, which may produce an incorrect or misleading error message. This change uses the existing `throw_winsock_error_()` helper so that the reported exception contains the actual Winsock error.


gabime:
Thanks @sage-mode-hunter

PR #3630, opened 2026-06-29
Use WSAStartup return value for error reporting

sage-mode-hunter:
WSAStartup() returns the error code directly when initialization fails. This change uses the returned value instead of WSAGetLastError() so the reported error matches the Winsock API behavior.


gabime:
Thanks @sage-mode-hunter

PR #3633, opened 2026-07-10
Buffer source_loc filename/funcname in log_msg_buffer

### LINKED ISSUE #3195: log_msg_buffer should also buffer source_loc char*
The `source_loc` struct holds `const char *filename` and `const char *funcname` which need buffering for the same reason the logger name and payload `string_view` do.

I encountered this issue when using `ringbuffer_sink`.

vsaraikin:
Fixes #3195.

`log_msg_buffer` copies `logger_name` and `payload` into its own buffer and re-points the string_views at that owned storage, precisely so a queued message does not depend on the caller's stack/temporary memory. The `source` field was left out: `source_loc` holds `const char* filename` and `const char* funcname`, and those pointers were copied as-is, so a buffered message keeps pointing at the caller's memory.

For the `SPDLOG_*` macros this is harmless because `__FILE__` / `SPDLOG_FUNCTION` are static string literals. But when the `source_loc` is built from non-static strings (a runtime `source_loc`, as with `ringbuffer_sink` in the issue, or async logging), those pointers dangle once the caller's strings go out of scope while the message is still buffered.

This appends `filename` and `funcname` (null-terminated) to the same buffer next to `logger_name`/`payload`, and re-points `source.filename`/`source.funcname` into it in `update_string_views()`, mirroring how the existing string_views are handled. `nullptr` filename/funcname are preserved as-is.

## Reproduction (AddressSanitizer)

Logging a message whose `source_loc` strings live on the stack, then retrieving it after they go out of scope:

- **before:** `AddressSanitizer: stack-use-after-scope READ of size 17 ... in strrchr` (the pattern formatter reading `source.filename`)
- **after:** clean, and the filename/funcname format correctly.

## Testing

- Added `ringbuffer buffers source_loc strings` in `tests/test_ringbuffer.cpp`: logs with a scoped runtime `source_loc`, lets the strings go out of scope, overwrites the freed stack, then asserts the buffered message still formats `%g:%# %!` with the original filename/funcname. Fails under ASan (`SPDLOG_SANITIZE_ADDRESS`) without the fix, passes with it.
- Full `spdlog-utests` suite passes with ASan enabled.

### DIFF
--- a/include/spdlog/details/log_msg_buffer-inl.h
+++ b/include/spdlog/details/log_msg_buffer-inl.h
@@ -7,6 +7,9 @@
 #include <spdlog/details/log_msg_buffer.h>
 #endif
 
+#include <cstring>
+#include <cstddef>
+
 SPDLOG_NAMESPACE_BEGIN
 namespace details {
 
@@ -14,6 +17,7 @@ SPDLOG_INLINE log_msg_buffer::log_msg_buffer(const log_msg &orig_msg)
     : log_msg{orig_msg} {
     buffer.append(logger_name.begin(), logger_name.end());
     buffer.append(payload.begin(), payload.end());
+    append_source();
     update_string_views();
 }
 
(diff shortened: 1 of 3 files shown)


gabime:
Thanks @vsaraikin

PR #3634, opened 2026-07-11
Add a pkg-config file for header-only usage

### LINKED ISSUE #3106: pkg-config file does not support spdlog_header_only
`spdlog.pc` has `CFlags: "-I${includedir}" -DSPDLOG_COMPILED_LIB -DSPDLOG_FMT_EXTERNAL`.

Perhaps there should be a separate `spdlog_header_only.pc`?

c-tonneslan:
Closes #3106.

`spdlog.pc` describes the compiled library: its `CFlags` include `-DSPDLOG_COMPILED_LIB` and its `Libs` link `-lspdlog`. A header-only consumer that pulls flags from `pkg-config --cflags --libs spdlog` therefore gets `-DSPDLOG_COMPILED_LIB` (which makes the headers expect a library to link) and `-lspdlog` (which may not be built), breaking the build.

This installs a companion `spdlog_header_only.pc`, generated from the `spdlog_header_only` INTERFACE target, so its defines and libs match header-only use: no `-DSPDLOG_COMPILED_LIB`, no `-lspdlog` (just the include dir and `-pthread`, plus `SPDLOG_FMT_EXTERNAL` / `Requires: fmt` when external fmt is selected, mirroring the existing `.pc`).

Verified the generated files with `cmake -DSPDLOG_INSTALL=ON`:
- `spdlog.pc` — unchanged.
- `spdlog_header_only.pc` — `CFlags: -I${includedir}`, `Libs: -L${libdir} -pthread`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -395,6 +395,19 @@ if(SPDLOG_INSTALL)
     configure_file("cmake/${PROJECT_NAME}.pc.in" "${pkg_config}" @ONLY)
     install(FILES "${pkg_config}" DESTINATION "${pkgconfig_install_dir}")
 
+    # Header-only pkg-config file: no libspdlog to link and no SPDLOG_COMPILED_LIB,
+    # so header-only consumers get the correct flags.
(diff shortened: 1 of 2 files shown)


gabime:
Thanks @c-tonneslan

PR #3637, opened 2026-07-24
docs: correct daily/hourly sink filename format in comments

dvirarad:
The `daily_filename_calculator` and `hourly_filename_calculator` build filenames with an underscore separator (`basename_YYYY-MM-DD.ext` and `basename_YYYY-MM-DD_HH.ext`), but the accompanying comments described the format with a dot separator — and, for the hourly sink, a dash before the hour. This updates the comments to match the output actually produced by `calc_filename` (see the regex in `tests/test_daily_logger.cpp`). Comments only; no behavior change.


gabime:
Thanks @dvirarad

Issue #3223, opened 2024-10-18
v1.14.1, GCC 14.2.1, With Precompiled Header -- Noisy Compilation From Spdlog


limitedAtonement:
I have a small example that reproduces the problem:

1. `git clone --recurse-submodules git@github.com:limitedAtonement/spdlog-problem`
   - Clones the example repository and spdlog submodule at v1.14.1
   - To check the spdlog version, `cd spdlog-problem/src/spdlog && git log --oneline --graph --decorate --all` or whatever
1. `cd spdlog-problem`
4. `git checkout origin/illustration`
5. `make clean && make`

I get output like

```
In file included from src/spdlog/include/spdlog/fmt/bundled/format.h:4530,                                                               
                 from src/spdlog/include/spdlog/fmt/bundled/core.h:2967,                                                                 
                 from src/spdlog/include/spdlog/fmt/fmt.h:24,                                                                                                                                                                                                                     
                 from src/spdlog/include/spdlog/common.h:50,
                 from src/spdlog/include/spdlog/spdlog.h:12,
                 from src/stdafx.hpp:1:                                                                                                                                                                                                                                           src/spdlog/include/spdlog/fmt/bundled/format-inl.h: In instantiation of ‘fmt::v10::detail::thousands_sep_result<Char> fmt::v10::detail::thousands_sep_impl(locale_ref) [with Char = wchar_t]’:
src/spdlog/include/spdlog/fmt/bundled/format.h:1268:37:   required from here                                                             
 1268 |   return thousands_sep_impl<wchar_t>(loc);      
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~                                                                                                                                                                                                                                 src/spdlog/include/spdlog/fmt/bundled/format-inl.h:95:9: warning: possibly dangling reference to a temporary [-Wdangling-reference]      
   95 |   auto& facet = std::use_facet<std::numpunct<Char>>(loc.get<std::locale>());                                                     
      |         ^~~~~                                               
src/spdlog/include/spdlog/fmt/bundled/format-inl.h:95:52: note: the temporary was destroyed at the end of the full expression ‘std::use_facet<__cxx11::numpunct<wchar_t> >(fmt::v10::detail::locale_ref::get() const [with Locale = std::locale]())’             
   95 |   auto& facet = std::use_facet<std::numpunct<Char>>(loc.get<std::locale>());                                                     
[...]

PR #3638, opened 2026-08-02
fix: avoid -Wdangling-reference false positives in bundled fmt locale helpers

viordash:
GCC 13/14 warns in the bundled fmt when a precompiled header is used (see #3223):

    include/spdlog/fmt/bundled/format-inl.h:114: warning: possibly dangling reference
      to a temporary [-Wdangling-reference]
    include/spdlog/fmt/bundled/xchar.h:49: same

It is a false positive: `locale_ref::get<std::locale>()` returns a copy, and a facet
reference stays valid as long as any copy of that locale exists. GCC normally suppresses
it because `std::use_facet` is declared in a system header, but that suppression is lost
when the declaration comes from a PCH, reproducible with plain `<locale>` and no fmt
involved.

The fix binds the locale copy to a named variable at the two call sites, so no reference
is bound to a temporary. Same single copy, no behaviour change, and the lifetime becomes
explicit instead of relying on the "valid while any copy exists" guarantee.

Checked with GCC 13.4: warnings gone with a PCH, formatting output unchanged
(`{:L}` with a custom `numpunct`, narrow and wide).

These files are a vendored copy of fmt, so fixing it here is what actually removes the
warning for spdlog users.

### DIFF
--- a/include/spdlog/fmt/bundled/format-inl.h
+++ b/include/spdlog/fmt/bundled/format-inl.h
@@ -111,7 +111,8 @@ inline void fwrite_all(const void* ptr, size_t count, FILE* stream) {
 
 template <typename Char>
 FMT_FUNC auto thousands_sep_impl(locale_ref loc) -> thousands_sep_result<Char> {
-  auto&& facet = use_facet<numpunct<Char>>(loc.get<locale>());
+  auto loc_copy = loc.get<locale>();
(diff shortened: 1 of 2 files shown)


gabime:
Thanks. This belongs to the [fmt](https://github.com/fmtlib/fmt) codebase. Please reopen this there

PR #3640, opened 2026-08-03
Reuse backtrace ring-buffer slots when recording messages

fyrsta7:
## Summary

When backtracing is enabled, each logged message is copied into a fixed-size
ring buffer. The current `backtracer::push_back()` path builds a temporary
`log_msg_buffer` and then move-assigns it into the next queue slot.

This change lets the backtracer write directly into the next existing
ring-buffer slot:

- `circular_q` gets an internal overwrite helper that advances the tail with
  the same overrun behavior as `push_back()`.
- `log_msg_buffer` gets assignment from `log_msg`, preserving the owned logger
  name, payload, and source-location strings.
- `backtracer::push_back()` uses that path to reuse the destination buffer
  capacity across overwrites.

Normal `log_msg_buffer` copy/move construction and assignment are left
unchanged.

## Why this helps

Backtrace buffers are fixed-size circular queues, so after the queue has warmed
up, each push usually overwrites an existing slot. Reusing that slot avoids the
temporary buffer and lets repeated backtrace messages reuse already allocated
payload storage.

## Validation

Correctness:

```sh
cmake -S . -B build -DSPDLOG_BUILD_TESTS=ON -DSPDLOG_BUILD_EXAMPLE=OFF -DSPDLOG_BUILD_BENCH=ON -DCMAKE_BUILD_TYPE=Release
cmake --build build -j8
ctest --test-dir build --output-on-failure
```

Focused benchmark, measured locally with repeated `details::backtracer::push_back`
calls into a 32-entry backtrace buffer:

| Payload | Before | After | Change |
| ---: | ---: | ---: | ---: |
| 64 bytes | 12.532 ns/push | 8.205 ns/push | 34.52% faster |
| 256 bytes | 22.024 ns/push | 11.637 ns/push | 47.16% faster |
| 1024 bytes | 35.303 ns/push | 22.625 ns/push | 35.91% faster |

The benchmark focuses on the path changed here and avoids file sink I/O noise.

### DIFF
--- a/include/spdlog/details/backtracer-inl.h
+++ b/include/spdlog/details/backtracer-inl.h
@@ -42,7 +42,7 @@ SPDLOG_INLINE bool backtracer::enabled() const { return enabled_.load(std::memor
 
 SPDLOG_INLINE void backtracer::push_back(const log_msg &msg) {
     std::lock_guard<std::mutex> lock{mutex_};
-    messages_.push_back(log_msg_buffer{msg});
+    messages_.push_back_overwrite([&msg](log_msg_buffer &dest) { dest = msg; });
 }
 
 SPDLOG_INLINE bool backtracer::empty() const {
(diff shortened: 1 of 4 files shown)


fyrsta7:
Closing this draft in favor of an updated branch with the same scope, added validation coverage, and refreshed benchmark results.

PR #3644, opened 2026-08-11
feat(pattern): truncate @ source location from the start (#2329)

### LINKED ISSUE #2329: format tags: option to crop from beggining of the string
When using "%30@" for example, it would be more useful to show the end of the path, instead of cropping the end and showing only the beginning.

rachelwelchio:
Fixes #2329

When `%!@` truncates a source location, keep the tail of the path (filename and line) instead of the prefix. Other formatters keep the existing truncate-from-end behavior.

## Changes
- Add `truncate_from_start_` to `padding_info` and teach `scoped_padder` to erase from the beginning when set.
- Enable tail truncation automatically for the `@` formatter when `!` is present.
- Add a regression test in `tests/test_pattern_formatter.cpp`.

## Test plan
- [x] `tests/test_pattern_formatter.cpp` — `source location truncate from start`


gabime:
Thanks. Please fix the CI compile failures . Also tests wont pass in windows because pf the har coded “\n”

rachelwelchio:
Thanks for the review!

Both issues should be fixed in the latest commit (84833207):

1. **CI compile failure** — memory_buf_t is mt::basic_memory_buffer in the default build, which has no erase(). Truncate-from-start now uses memmove + 
esize, consistent with the existing truncate-from-end path.
2. **Windows test** — removed the hardcoded \n in the assertion; 	est_sink.lines() already strips default_eol (see 	est_sink.h).

Could you please approve the workflows when you have a moment? The follow-up commit is currently waiting on maintainer approval before CI runs on the fork PR.

Happy to adjust if anything else looks off.

rachelwelchio:
Thanks for the review — both items should be fixed in this push: memmove+resize for truncate-from-start (basic_memory_buffer has no erase), and the custom eol test now uses the same eol variable instead of a hardcoded newline.

PR #3647, opened 2026-08-24
cmake: optionally generate tweakme header

### LINKED ISSUE #3077: Auto generate tweakme.h
It looks like the `include/spdlog/tweakme.h` file should be automatically generated by CMake depending on the build configuration.

Various Linux distributions need to manually patch it in downstream to resolve issues with packages not using pkg-config or CMake.

fyrsta7:
Fixes #3077.

## Motivation

Several spdlog build options are represented both as CMake target definitions
and as settings in `tweakme.h`. Consumers that do not use the exported CMake or
pkg-config metadata cannot recover the configuration used to build an
installed spdlog library and currently need downstream patches to the header.

## Solution

Add an opt-in `SPDLOG_GENERATE_TWEAKME_H` CMake option, disabled by default.
When enabled, CMake generates `include/spdlog/tweakme.h` from the existing
spdlog configuration options. Source-tree compiled and header-only targets use
the generated header, and installation selects it in place of the editable
source header.

When generation is disabled, the existing source header and installation
behavior are unchanged. The option does not change the existing exported
target definitions, C++11 support, or pkg-config metadata.

## Validation

The CMake integration test covers source-tree and installed consumers for both
compiled and header-only targets, static and shared builds, and a raw installed
header-only consumer that does not use a spdlog CMake target or pkg-config
metadata. It also checks that generated-header include precedence is effective.

The default and generated configurations both pass the official compiled and
header-only test suites. Static and shared C++11 source-tree consumers,
installed package consumers, and the raw installed-header consumer pass
locally. I also verified separately that a default installation produces a
`tweakme.h` byte-for-byte identical to the source header. The nested host-run
integration test is omitted while cross-compiling.


tests/cmake/tweakme_test.cmake
gabime: Please add some short explanation on what these tests do. It is not trivial to understand

PR #3648, opened 2026-09-02
Add flush_all() to spslog api

r2p2:
When my application was quitting not so gracefully, I realized I was missing log output which could help finding the issue.

While it was easy to issue flush on the default logger, getting access to the other loggers seemed to be a bit complicated. Making details::registry::flush_all() publicly available did solve my issue pretty easily and I don't see why it should not be accessible.


tt4g:
If you want to flush the logs when the application exits, `spdlog::shutdown()` will flush all loggers. Have you tried that?

r2p2:
> If you want to flush the logs when the application exits, `spdlog::shutdown()` will flush all loggers. Have you tried that?

I just tested it and it seems the logs are not getting flushed when calling `shutdown()`. I also don't see which part of the code under shutdown would cause the loggers to be flushed.

It would be possible to add a call to `flush_all()` to `shutdown()` before `drop_all()` to solve my stated issue I guess. But I would not be opposed to having the public `flush_all()` nevertheless. After each event loop cycle, I call flush on the default logger, not realizing there might be more loggers registered. So that would be useful there too but I would be happy for now if it would work when calling `shutdown()`.

tt4g:
I suggested using `spdlog::shutdown()` as a workaround until this change is released.

> I just tested it and it seems the logs are not getting flushed when calling `shutdown()`. I also don't see which part of the code under shutdown would cause the loggers to be flushed.

A flush operation occurs when the reference to both the logger and the sink held by the logger is removed.
If a reference remains somewhere, the flush may not occur.

r2p2:
That makes sense, I keep logger references in various components which I cannot simply destroy when the crash handler is called. Thats why `shutdown()` does not flush in my case.

I think I will keep my patched spdlog version running and wait for confirmation if this PR will proceed in any way or if it gets rejected, I'll check if I find a way to refer differently to my loggers so they get destroyed on shutdown.

tt4g:
You can use `spdlog::apply_all()` instead of `spdlog::shutdown()` as a workaround.

```cpp
spdlog::apply_all([](auto logger)
{
    logger->flush();
});
```

gabime:
Merged. Thanks @r2p2

PR #3649, opened 2026-09-03
Don't show spdlog message and build type when rerunning CMake

vadz:
These messages are useful, but only once, showing them again and again and again as CMake reruns in the same build tree is more annoying than anything else.

Skip doing this by defining a helper spdlog_message_once() function and use it instead of message().

----

I'm submitting this because I use spdlog as as a submodule in my main project and I got tired of seeing

```
[0/2] Re-checking globbed directories...
[1/2] Re-running CMake...
-- Build spdlog: 1.12.0
-- Build type: Debug
-- Configuring done (0.8s)
-- Generating done (0.3s)
-- Build files have been written to: $builddir
```

when running `ninja` all the time. With this PR this reduces to

```
[0/2] Re-checking globbed directories...
ninja: no work to do.
```


gabime:
Thanks @vadz

PR #3651, opened 2026-09-14
Fix CI failures caused by aging of Debian Bullseye and Windows 2022

dlex:
* Remove `debian-security` from `sources.list` on gcc≤11 to fix broken Bullseye mirror. The `bullseye-security` mirror has expired timestamps and missing .deb files. That causes apt-get install to fail in the `gcc:11` and `gcc:9` containers.

* Pin Win runner to 2022; the latest can resolve to 2025 which does not have VS17 anymore


gabime:
whats the windows related change?

dlex:
It was same as in 4800ae0 apparently, see the 2nd commit. I will squash it on merge and it will disappear. PR description updated.

PR #3658, opened 2026-09-19
Prevent deadlock with zero-sized async queues

DosX-dev:
Creating a `thread_pool` with a queue size of zero succeeds, but the pool cannot enqueue any messages. During destruction, it tries to enqueue a `terminate` message using the blocking policy and waits forever because the queue has no capacity.

For example:

```cpp
{
    spdlog::details::thread_pool pool(0, 1);
}  //  <-  hangs while destroying the pool
```

Reject a zero queue size in the constructor before starting worker threads. This is consistent with the existing validation of the thread count and avoids turning an invalid configuration into a shutdown deadlock.

A regression test verifies that the constructor throws spdlog::spdlog_ex for a zero-sized queue.


gabime:
Thanks @DosX-dev

PR #3659, opened 2026-09-21
fix: format null C string arguments

### LINKED ISSUE #3653: crash when nullptr
```
const char *test1 = nullptr
spdlog::info("value: {}", test1);
```

then the program crash.
Can the code automatically detect and handle this? 
Thanks

sb123sb123:
Fixes #3653

## Problem

Formatting a null C string crashes instead of producing a useful log value. This affects both direct bundled fmt formatting and log messages that pass a null `const char*` argument.

## Cause

The bundled fmt string-pointer overloads call `report_error` when the pointer is null. The logger therefore aborts while formatting a valid nullable C string argument.

## Fix

Render null C string pointers as `(null)` in the bundled fmt overloads. Non-null strings and pointer presentation keep their existing behavior. The regression covers both direct formatting and the logger output.

## Tests

- Focused null C-string regression: 2 assertions passed.
- Bundled-fmt unit suite: 705 assertions in 190 test cases passed.
- `std::format` unit suite: 697 assertions in 186 test cases passed; the bundled-fmt-specific regression is excluded for that backend.
- `git diff --check` and clang-format check passed.

## Limitations

Validation ran on the Windows host with the repository's Visual Studio toolchain. The broader CTest run also reports the existing raw-header consumer failure that requires compiling with `/utf-8`; that environment/toolchain limitation is unrelated to this patch. Unix and other compiler matrices were not available on the host.

## AI assistance disclosure

This pull request was prepared with AI assistance. OpenAI Codex assisted with issue triage, implementation, regression-test design, validation, and drafting this description. The submitter reviewed the change and takes responsibility for every submitted line; no AI co-author was added.


tt4g:
The files in the `include/spdlog/fmt/bundled/` directory were copied from [fmt](https://github.com/fmtlib/fmt).
Requests must be submitted to the [fmt repository](https://github.com/fmtlib/fmt). When a new fmt version is released, the files in the `include/spdlog/fmt/bundled/` directory will be updated.

